PR #2, opened 2021-06-13
build: Add --with-qml configure option

hebasto:
This PR gets the build system ready for QML-based GUI builds.

Additionally, BDB wallet support by default is disabled -- it seems no reason to take BDB wallets in account for the next-gen GUI.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1801,6 +1801,7 @@ AM_CONDITIONAL([USE_ASM],[test x$use_asm = xyes])
 AM_CONDITIONAL([WORDS_BIGENDIAN],[test x$ac_cv_c_bigendian = xyes])
 AM_CONDITIONAL([USE_NATPMP],[test x$use_natpmp = xyes])
 AM_CONDITIONAL([USE_UPNP],[test x$use_upnp = xyes])
+AM_CONDITIONAL([BUILD_WITH_QML], [test x$use_qml = xyes])
 
 AC_DEFINE(CLIENT_VERSION_MAJOR, _CLIENT_VERSION_MAJOR, [Major version])
 AC_DEFINE(CLIENT_VERSION_MINOR, _CLIENT_VERSION_MINOR, [Minor version])
@@ -1940,6 +1941,7 @@ if test "x$enable_wallet" != "xno"; then
 fi
 echo "  with gui / qt   = $bitcoin_enable_qt"
 if test x$bitcoin_enable_qt != xno; then
+    echo "    with qml      = $use_qml"
     echo "    with qr       = $use_qr"
 fi
 echo "  with zmq        = $use_zmq"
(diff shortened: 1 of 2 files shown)


fanquake:
Isn't diverging from upstream just going to make rebasing / merging here needlessly harder? 

> --disable-fuzz-binary and --enable-suppress-external-warnings by default -- it is expected that designers will also make their builds, and these defaults will make their lives a bit more comfortable :)

Why are designers capable of doing `./configure` but not `./configure --disable-fuzz-binary --enable-suppress-external-warnings`? I'd assume they would be rarely re-configuring.

hebasto:
@fanquake 
> Isn't diverging from upstream just going to make rebasing / merging here needlessly harder?
> 
> > --disable-fuzz-binary and --enable-suppress-external-warnings by default -- it is expected that designers will also make their builds, and these defaults will make their lives a bit more comfortable :)
> 
> Why are designers capable of doing `./configure` but not `./configure --disable-fuzz-binary --enable-suppress-external-warnings`? I'd assume they would be rarely re-configuring.

Ok, the last two commits are dropped, the OP updated.

configure.ac
```diff
@@ -127,10 +127,10 @@ AC_ARG_WITH([sqlite],
   [use_sqlite=auto])
 
 AC_ARG_WITH([bdb],
-  [AS_HELP_STRING([--without-bdb],
-  [disable bdb wallet support (default is enabled if wallet is enabled)])],
+  [AS_HELP_STRING([--with-bdb],
```
jarolrod, line 130: I don't think we should include such a change in this repo, we should only focus on adding support for building with qml. 

What we can do is just not program support for generating a `bdb` wallet with the new QML gui, but leave this as is. I think, as fanquake states, we should try not to diverge too much from the core repo.

hebasto: Ok, the commit dropped.

doc/qml-gui.md
```diff
@@ -0,0 +1,15 @@
+# QML-BASED GUI BUILD NOTES
```
jarolrod, line 1: can you drop this change, i will include it in my doc pr

hebasto: Done.

jarolrod (commented):
ACK d21a6a3bc1cc1ec6f9408dc4dbed75fd84a6ed1d

Prior to installing the required dependencies (`qtdeclarative5-dev` and `qtquickcontrols2-5-dev`):
``` 
checking for QT_QML... no
configure: WARNING: Qt5Qml >= 5.9.5 not found; bitcoin-qt frontend will not be built
```

After installing dependencies:
```
checking for QT_QML... yes
checking for QT_QUICK... yes
checking for QT_QUICKCONTROLS... yes
```

PR #5, opened 2021-07-04
Add basic QML support to Qt resource system

hebasto:
Added basic support for QML to Qt resource system.

Suggested the following file directory layout:
- `src/qml/`
- `src/qml/*.{h|cpp}`
- `src/qml/pages/*.qml`
- `src/qml/<other resource directories>`

### DIFF
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -105,6 +105,7 @@ QT_QRC_LOCALE_CPP = qt/qrc_bitcoin_locale.cpp
 QT_QRC_LOCALE = qt/bitcoin_locale.qrc
 
 BITCOIN_QT_H = \
+  qml/bitcoin.h \
   qt/addressbookpage.h \
   qt/addresstablemodel.h \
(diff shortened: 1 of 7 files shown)


jarolrod (commented):
Concept ACK

Tested on macOS 11.3, can produce the stub window 🎉 (very aesthetic). Will review more in depth after you rebase on top of the changes to #2 and merge #4.

Screen Shot 2021-07-05 at 8 14 29 PM

hebasto:
> Will review more in depth after you rebase on top of the changes to #2 and merge #4.

Rebased and ready :)

promag (commented):
Approack ACK. Initially I though having this under src/qt but having separate seems preferable to avoid conflicts. Will test.

promag (commented):
ACK e2104f8909. Comparing to https://github.com/bitcoin/bitcoin/pull/16883 I prefer this approach especially because we can change the startup without affecting the existing bitcoin-qt.

PR #19, opened 2021-08-12
build: Add required packages for static builds

hebasto:
This PR makes possible to cross-compile QML-based `bitcoin-qt` for all supported platforms.


hebasto:
@Bosch-0 

... and, of course, for Windows :)

PR #17, opened 2021-08-07
Handle initialization errors

hebasto:
This PR add functionality to handle initialization errors, and show error messages to users, e.g.:

Screenshot from 2021-08-09 22-31-21

---

Note for designers. The easiest way to trigger an error is to provide an invalid command line option:

```
$ ./src/qt/bitcoin-qt -qq
```

PR #21, opened 2021-08-23
doc: document runtime dependencies

jarolrod:
This is a follow-up to #17, which introduced new runtime dependencies. 
This PR documents the runtime dependencies required for proper execution. 📝


hebasto:
Concept ACK.

To successfully run qml app (particulary, to process `stub.qml`) the following packages also are required:
- `qml-module-qtquick2`
- `qml-module-qtquick-controls2`
- `qml-module-qtquick-window2`

PR #11, opened 2021-07-14
Add start/shutdown functionality

hebasto:
This PR:
- is based on #10 
- introduces the `NodeModel` class as a "backend" for the QML-based GUI
- adds basic start/shutdown functionality

Init errors handling could be a natural followup.

PR #28, opened 2021-08-28
Sync with the main repo

hebasto:
Sync with the main repo up to https://github.com/bitcoin-core/gui/commit/83daf47898f8a79cb20d20316c64becd564cf54c.

PR #25, opened 2021-08-27
Share qInstallMessageHandler between QWidget and QML code

hebasto:
This PR effectively makes `qInstallMessageHandler` call available in the QML code.

On master (2aa44e608ac70ed7e56994634d8c15cf192f8390):
```
$ src/qt/bitcoin-qt -signet -printtoconsole
...
shutdown : Running Shutdown in thread
2021-08-27T23:12:48Z [qt-init] tor: Thread interrupt
2021-08-27T23:12:48Z [dnsseed] dnsseed thread exit
2021-08-27T23:12:48Z [qt-init] Shutdown: In progress...
2021-08-27T23:12:48Z [addcon] addcon thread exit
2021-08-27T23:12:48Z [torcontrol] RemoveLocal(ruw7mr3wdl3uq3hshqvq372zmzxo7sstndz4le2auscyajgi53tsp3qd.onion:38333)
2021-08-27T23:12:48Z [torcontrol] torcontrol thread exit
2021-08-27T23:12:48Z [msghand] msghand thread exit
2021-08-27T23:12:48Z [net] net thread exit
2021-08-27T23:12:48Z [opencon] ERROR: Error while reading proxy response
2021-08-27T23:12:48Z [opencon] opencon thread exit
2021-08-27T23:12:48Z [shutoff] DumpAnchors: Flush 2 outbound block-relay-only peer addresses to anchors.dat started
2021-08-27T23:12:48Z [shutoff] DumpAnchors: Flush 2 outbound block-relay-only peer addresses to anchors.dat completed (0.00s)
2021-08-27T23:12:48Z [scheduler] scheduler thread exit
2021-08-27T23:12:48Z [shutoff] Writing 0 unbroadcast transactions to disk.
2021-08-27T23:12:48Z [shutoff] Dumped mempool: 6e-06s to copy, 0.002382s to dump
2021-08-27T23:12:49Z [shutoff] FlushStateToDisk: write coins cache to disk (0 coins, 0kB) started
2021-08-27T23:12:49Z [shutoff] FlushStateToDisk: write coins cache to disk (0 coins, 0kB) completed (0.00s)
2021-08-27T23:12:49Z [shutoff] FlushStateToDisk: write coins cache to disk (0 coins, 0kB) started
2021-08-27T23:12:49Z [shutoff] FlushStateToDisk: write coins cache to disk (0 coins, 0kB) completed (0.00s)
2021-08-27T23:12:49Z [shutoff] [210528-bdb] Releasing wallet
[...]

PR #31, opened 2021-08-29
Add BlockCounter QML component which handles NotifyBlockTip signal

hebasto:
This PR:
1) introduces additional `components` sub-directory for QML objects for basic GUI elements
2) adds a prototype of QML object which displays block tip height
3) handles the `NotifyBlockTip` signal from the core

---

From Qt [docs](https://doc.qt.io/qt-5.15/qmlapplications.html):
> QML is a declarative language that allows user interfaces to be described in terms of their visual components and how they interact and relate with one another. It is a highly readable language that was designed to enable components to be interconnected in a dynamic manner, and it allows components to be easily reused and customized within a user interface.
> ...
> It allows **developers _and_ designers** alike to create highly performant, fluidly animated and visually appealing applications.

#### Review notes for developers

This PR continues to exploit the technique of embedding an instance of the `NodeModel` class into QML code as a context property (see #11). This approach differs from alternative ones, in a way that there are no signal handlers for signals from C++ code in the QML files -- only properties bindings are used.

The benefits of such an approach are:
- QML files are easy to read, reason about, and modify. Designers could easy work on them
- JavaScript code assignments in signal handlers break properties bindings

Exposing properties of the `NodeModel` class to QML code justifies decision to not re-use the `ClientModel` class because these classes have different designs.

#### Review notes for designers

This PR demonstrates an idea of small QML files that describe GUI elements and their interaction in a _declarative_ way.

You could easy and safely tinker with object properties. Just run `make clean && make` again to see the result of your changes.

Names of QML files in the `components` sub-directory are actually the names of defined QML objects. Therefore, to make your life easier, you could suggest names you are comfortable with.

Don't hesitate to request new QML files that describe your design ideas :)

### DIFF
--- a/src/qml/components/BlockCounter.qml
+++ b/src/qml/components/BlockCounter.qml
@@ -1,0 +1,21 @@
+// Copyright (c) 2021 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+// The BlockCounter component.
+
+import QtQuick 2.12
+import QtQuick.Controls 2.12
+
+Label {
+    property int blockHeight: 0
+    background: Rectangle {
+        color: "black"
+    }
+    color: "orange"
+    padding: 16
+    horizontalAlignment: Text.AlignHCenter
+    verticalAlignment: Text.AlignVCenter
+    font.pixelSize: height / 3
+    text: blockHeight
+}
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -85,7 +85,6 @@ static void RegisterMetaTypes()
 
     qRegisterMetaType<std::function<void()>>("std::function<void()>");
     qRegisterMetaType<QMessageBox::Icon>("QMessageBox::Icon");
-    qRegisterMetaType<interfaces::BlockAndHeaderTipInfo>("interfaces::BlockAndHeaderTipInfo");
 }
 
 static QString GetLangTerritory()
--- a/src/qml/bitcoin.cpp
+++ b/src/qml/bitcoin.cpp
@@ -131,7 +131,7 @@ int QmlGuiMain(int argc, char* argv[])
     InitExecutor init_executor{*node};
     QObject::connect(&node_model, &NodeModel::requestedInitialize, &init_executor, &InitExecutor::initialize);
     QObject::connect(&node_model, &NodeModel::requestedShutdown, &init_executor, &InitExecutor::shutdown);
-    // QObject::connect(&init_executor, &InitExecutor::initializeResult, &node_model, &NodeModel::initializeResult);
+    QObject::connect(&init_executor, &InitExecutor::initializeResult, &node_model, &NodeModel::initializeResult);
     QObject::connect(&init_executor, &InitExecutor::shutdownResult, qGuiApp, &QGuiApplication::quit, Qt::QueuedConnection);
     // QObject::connect(&init_executor, &InitExecutor::runawayException, &node_model, &NodeModel::handleRunawayException);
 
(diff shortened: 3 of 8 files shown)


hebasto:
@jonasschnelli @Talkless Do you mind reviewing this PR at a concept level?

hebasto:
Updated 7cc9c80ac80b245d75e16e87e5926716ee6a7ac0 -> 5467df4cd0e29794d38d1951fc9cb87e718a7809 ([pr31.01](https://github.com/hebasto/gui-qml/commits/pr31.01) -> [pr31.02](https://github.com/hebasto/gui-qml/commits/pr31.02)):

- rebased due to the conflict with #25
- the `blockTipHeight` is read-only now

hebasto:
Rebased 5467df4cd0e29794d38d1951fc9cb87e718a7809 -> e7b9a09fbdd9fa18ebd8bfd981421ecffde044e0 ([pr31.02](https://github.com/hebasto/gui-qml/commits/pr31.02) -> [pr31.03](https://github.com/hebasto/gui-qml/commits/pr31.03)) on top of the #28.

src/qml/components/BlockCounter.qml
```diff
@@ -0,0 +1,17 @@
+import QtQuick 2.12
+import QtQuick.Controls 2.12
+
+Rectangle {
+    color: "black"
+    width: label.width * 1.5
+
+    property int blockHeight
+
+    Label {
```
promag, line 10: These controls have a `background` property:

```qml
Label {
    property int blockHeight: 0
    background: Rectangle {
        color: "black"
    }
    color: "orange"
    padding: 16
    text: blockHeight
}
```

src/qt/bitcoin.cpp
```diff
@@ -85,7 +85,6 @@ static void RegisterMetaTypes()
 
     qRegisterMetaType<std::function<void()>>("std::function<void()>");
     qRegisterMetaType<QMessageBox::Icon>("QMessageBox::Icon");
-    qRegisterMetaType<interfaces::BlockAndHeaderTipInfo>("interfaces::BlockAndHeaderTipInfo");
```
promag, line 88: Why is this removed?

hebasto:
Updated e7b9a09fbdd9fa18ebd8bfd981421ecffde044e0 -> d473a47fe3e46bc68f946bb222b4e1b280016190 ([pr31.03](https://github.com/hebasto/gui-qml/commits/pr31.03) -> [pr31.04](https://github.com/hebasto/gui-qml/commits/pr31.04)):

- rebased due to the conflict with #34
- addressed @promag's [comment](https://github.com/bitcoin-core/gui-qml/pull/31#discussion_r698833036)

src/qml/components/BlockCounter.qml
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui-qml/pull/31#issuecomment-908944132).

jarolrod (commented):
ACK d473a47fe3e46bc68f946bb222b4e1b280016190

This feature works nicely and code is clean. The main thing to talk about here is the approach taken in terms of handling the Core node instance. 

Here, we have a certain layer of separation between the reference to the core node and what we do at the QML layer. This should be the approach we take going forward. It is scary to be passing around this reference around raw.

Additionally, here we are showcasing why using QML is a good approach for our project. There is great separation between our business logic and the UI logic. We can have our fellow designers live in one part of the codebase where programming is not required, while us developers focus on making things functional. To me, this showcases that we made the right decision.✨

hebasto:
Trivially rebased d473a47fe3e46bc68f946bb222b4e1b280016190 -> 3ead757f60434933a915b0c726e58b419ea8b8c4 ([pr31.04](https://github.com/hebasto/gui-qml/commits/pr31.04) -> [pr31.05](https://github.com/hebasto/gui-qml/commits/pr31.05)) on top of the recent [sync](https://github.com/bitcoin-core/gui-qml/pull/36) with the main repo.

promag (commented):
Code review ACK 3ead757f60434933a915b0c726e58b419ea8b8c4.

jarolrod (commented):
ACK 3ead757 

Changes since my last review is only rebase over the weekly upstream sync

PR #37, opened 2021-09-26
Expose init executor instance

promag:
This PR exposes `InitExecutor` instance to QML so that we can:
 - show the window as early as possible
 - display loading feedback on that window
 - connect/bind in QML to the relevant object's properties/signals to update the UI

The `InitExecutor` is then kept in the main GUI thread (otherwise QML engine complains) and the instance is exposed in the root engine context.

This change also removes the "proxy role" of `NodeModel`. Further simplifications/improvements will follow, but I'd like to present this approach first before going forward.

### DIFF
--- a/src/qml/bitcoin.cpp
+++ b/src/qml/bitcoin.cpp
@@ -163,16 +163,11 @@ int QmlGuiMain(int argc, char* argv[])
 
     NodeModel node_model{*node};
     InitExecutor init_executor{*node};
-    QObject::connect(&node_model, &NodeModel::requestedInitialize, &init_executor, &InitExecutor::initialize);
-    QObject::connect(&node_model, &NodeModel::requestedShutdown, &init_executor, &InitExecutor::shutdown);
-    QObject::connect(&init_executor, &InitExecutor::initializeResult, &node_model, &NodeModel::initializeResult);
-    QObject::connect(&init_executor, &InitExecutor::shutdownResult, qGuiApp, &QGuiApplication::quit, Qt::QueuedConnection);
-    // QObject::connect(&init_executor, &InitExecutor::runawayException, &node_model, &NodeModel::handleRunawayException);
 
     qGuiApp->setQuitOnLastWindowClosed(false);
     QObject::connect(qGuiApp, &QGuiApplication::lastWindowClosed, [&] {
         node->startShutdown();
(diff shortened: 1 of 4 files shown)


promag:
Related https://github.com/bitcoin-core/gui/pull/434

hebasto:
Concept ACK.

hebasto:
> Concept ACK.

To elaborate, I'm "Concept ACK" on: 

> * show the window as early as possible
> 
> * display loading feedback on that window

But I'm not sure it is required to expose an `InitExecutor` instance to achieve these goals.

A relevant discussion in https://github.com/bitcoin-core/gui/pull/434:
- https://github.com/bitcoin-core/gui/pull/434#issuecomment-929788149
- https://github.com/bitcoin-core/gui/pull/434#issuecomment-930028976

On of the main goals of starting this project is involving Bitcoin Design community in the development process from the day 0. And to achieve it, keeping QML files as simple as possible is important for the project.

As were mentioned in #31, with the current [approach](https://doc.qt.io/qt-5.12/qtqml-cppintegration-contextproperties.html#setting-an-object-as-a-context-property):
>    * QML files are easy to read, reason about, and modify. Designers could easy work on them

---

Anyway, as https://github.com/bitcoin-core/gui/pull/434 is merged, this PR requires rebasing on top of the updated sync with the main repo that is currently blocked by #41.

promag (commented):
Currently the designed is already exposed to `nodeModel` and `nodeModel.startNodeInitializionThread()`. And eventually it will have to be exposed to more stuff since that's the way QML works.

This change at least removes the wiring from `InitExecutor` and `NodeModel`. 

Follow-ups can improve how QML deals with the application state.

hebasto:
> Currently the designer is already exposed to `nodeModel` and `nodeModel.startNodeInitializionThread()`.

It is all about the same object :)

> And eventually it will have to be exposed to more stuff since that's the way QML works.

"more stuff" could remain the same single `nodeModel` object.

> This change at least removes the wiring from `InitExecutor` and `NodeModel`.

But it exposes _another_ object into QML code.

Having _the only_ object, which is exposed to the QML, has a simple benefit of a guarantee that its internal state, i.e. internal state of the C++ backend, is consistent at anytime.

PR #38, opened 2021-09-28
Instantiate NodeModel in qml

promag:
Based on #37, this change moves `NodeModel` instantiation to QML. It is instanced only when `InitExecutor` is ready.

The QML engine is specialized in order to provide access to `interfaces::Node` instance and possible others.

A nice follow-up is to improve how `InitExecutor` is exposed in QML.

### DIFF
--- a/src/qml/pages/stub.qml
+++ b/src/qml/pages/stub.qml
@@ -2,6 +2,7 @@
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
+import BitcoinCore 1.0
 import QtQuick 2.12
 import QtQuick.Controls 2.12
 import QtQuick.Layouts 1.11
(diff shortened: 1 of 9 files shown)


hebasto (commented):
I think the _"qml: Update block tip in the gui thread"_ commit (54924fc0c1e0b5db349ad51ba742abd26929d618) is valuable by its own as `NodeModel::setBlockTipHeight` emits the `blockTipHeightChanged` signal. Maybe put it in its own PR?

Also is it worth to `assert(QThread::currentThread() == thread());` in each function which emits signals?

src/qml/pages/stub.qml
```diff
@@ -16,16 +17,21 @@ ApplicationWindow {
...
-    }
-
+    Loader {
+        id: loader
+        active: initExecutor.ready
+        anchors.fill: parent
+        sourceComponent: Item {
+            NodeModel {
+                id: node_model
+            }
```
hebasto, line 27: Does instantiating an object in QML have performance penalties comparing to instantiating it in C++ code?

PR #40, opened 2021-09-28
Add ChainModel

promag:
Based on #38. This PR adds and uses the `ChainModel`. This model exposes chain data, like block heights and hashes. Can be used as a source for list views and repeaters. The model uses a lazy load approach, ie, data is fetched from `interfaces::Chain` as needed.

### DIFF
--- a/src/qml/chainmodel.cpp
+++ b/src/qml/chainmodel.cpp
@@ -1,0 +1,94 @@
+// Copyright (c) 2021 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include <qml/chainmodel.h>
+
+#include <interfaces/chain.h>
+#include <interfaces/handler.h>
+#include <interfaces/node.h>
+#include <qml/engine.h>
+#include <qt/guiutil.h>
+
+#include <vector>
+
+struct Block
+{
+    const int height;
+    const std::string hash;
+};
+
+struct ChainModelPrivate
+{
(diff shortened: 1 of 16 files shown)


src/qml/chainmodel.cpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+bool ChainModel::canFetchMore(const QModelIndex&) const
+{
+    return d->blocks.size() == 0 || d->blocks[0].height > 0;
+}
+
+void ChainModel::fetchMore(const QModelIndex& parent)
+{
+    auto& chain = Engine::chain(this);
```
promag, line 70: Add `qDebug() << Q_FUNC_INFO;` while testing (scrolling the list) to see how lazy loading works.

src/qml/pages/stub.qml
```diff
@@ -14,12 +15,38 @@ ApplicationWindow {
...
+            BitcoinCoreComponents.BlockCounter {
+                id: blockCounter
+                anchors.centerIn: parent
+                height: parent.height / 3
+                blockHeight: node_model.blockTipHeight
+            }
+
+            ListView {
+                anchors.fill: parent
+                model: ChainModel {
```
promag, line 38: @ryanofsky this is a dumb example, but the idea is to access block details.

src/qml/chainmodel.cpp
promag: @ryanofsky using `interfaces::Chain` here.

jarolrod:
Concept ACK, we're going to need to expose more than just the nodemodel to qml. One big backend object would be quite a mess. All that's needed is good documentation of what is exposed to qml and why.

PR #41, opened 2021-09-28
Revert a commit from #5 to avoid merge conflict with bitcoin/bitcoin#22219

hebasto:
This PR reverts the _"refactor: Move qwidget and qml common code into the main() function"_ commit (8efd330d54e75e4a5de36278299b5227e4e5bfdf) from #5, and it allows to sync with the main repo without a merge conflict with bitcoin/bitcoin#22219.

### DIFF
--- a/src/qml/bitcoin.cpp
+++ b/src/qml/bitcoin.cpp
@@ -15,10 +15,12 @@
 #include <qt/guiutil.h>
 #include <qt/initexecutor.h>
 #include <util/system.h>
+#include <util/threadnames.h>
 #include <util/translation.h>
 
 #include <boost/signals2/connection.hpp>
 #include <memory>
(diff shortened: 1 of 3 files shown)


promag:
Tested ACK 498a318f86b05af91b1665ceae9dbca330224850, a clean merge is possible after this.

jarolrod (commented):
Post-merge ACK

I would like it if we can get some other opinions on the best way to deal with sync conflicts going forward.

hebasto:
> I would like it if we can get some other opinions on the best way to deal with sync conflicts going forward.

Yeah. To workaround that merge conflict this PR and #48 are required.

From my limited experience, the best way to deal with sync conflicts is to be a part of the same repo.

PR #42, opened 2021-10-02
Introduce ImageProvider class

hebasto:
The new `ImageProvider` class allows images in QML to be loaded using `QPixmap`s.

The last commit is a demo for our fellow designers.

For testing this PR one can run `bitcoin-qt` with different values of the `-chain=...` command-line option.

### DIFF
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -108,6 +108,7 @@ QT_QRC_LOCALE = qt/bitcoin_locale.qrc
 
 BITCOIN_QT_H = \
   qml/bitcoin.h \
+  qml/imageprovider.h \
   qml/nodemodel.h \
   qml/util.h \
   qt/addressbookpage.h \
@@ -285,6 +286,7 @@ BITCOIN_QT_WALLET_CPP = \
 
 BITCOIN_QML_BASE_CPP = \
   qml/bitcoin.cpp \
+  qml/imageprovider.cpp \
   qml/nodemodel.cpp \
   qml/util.cpp
 
--- a/src/qml/bitcoin.cpp
+++ b/src/qml/bitcoin.cpp
@@ -4,6 +4,7 @@
 
 #include <qml/bitcoin.h>
 
+#include <chainparams.h>
 #include <init.h>
 #include <interfaces/init.h>
 #include <interfaces/node.h>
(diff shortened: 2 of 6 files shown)


promag (commented):
Concept ACK. This is the correct way of exposing the dynamically generated icon based on the chain.

However, I wonder if we want to instead ship with an SVG for each chain?

hebasto:
> However, I wonder if we want to instead ship with an SVG for each chain?

Does using SVG make `ImageProvider` unneeded? How about caching?

promag:
Something like `Image { source: "qrc:/mainnet.svg" }` should work without custom image provider. What cache?

promag:
Let's discuss whether we want to use SVG. Qt support is detailed at https://doc.qt.io/qt-5/svgrendering.html.

hebasto:
> Something like `Image { source: "qrc:/mainnet.svg" }` should work without custom image provider.

It should work for any image, even non-svg.

> What cache?

https://doc.qt.io/qt-5.12/qquickimageprovider.html#image-caching

jarolrod (commented):
@Bosch-0 has an opinion on the usage of SVG, per [Icon Policy Documentation](https://github.com/bitcoin-core/gui/pull/310):

> "Both an icon source file, in Scalable Vector Graphics (SVG) format,
 and an optimized production file, in Portable Network Graphics (PNG) format"

> "SVGs are used as source files because they can scale while retaining image quality."

hebasto:
> @Bosch-0 has an opinion on the usage of SVG, per [Icon Policy Documentation](https://github.com/bitcoin-core/gui/pull/310):
> 
> > "Both an icon source file, in Scalable Vector Graphics (SVG) format,
> > and an optimized production file, in Portable Network Graphics (PNG) format"
> 
> > "SVGs are used as source files because they can scale while retaining image quality."

Being base open OpenGL and hardware acceleration, Qt Quick could have a native support for SVG images without any performance penalties. Unfortunately, I have no enough expertise in this area for now.

The only drawback is introducing a new Qt SVG module.

hebasto:
Rebased after the recent [sync](https://github.com/bitcoin-core/gui-qml/pull/47).

promag (commented):
Code review ACK b7171c8301c46b6e90988526cf04ac120a5522d0, tested but in a different branch.

PR #44, opened 2021-10-02
Do not swallow QML error messages

hebasto:
After #25 error messages in `QQmlApplicationEngine::load` call are effectively swallowed because they are routed into our logging facility which has no chance to print them by `return EXIT_FAILURE;`.

The duplication of the `DebugMessageHandler` function could be addressed later, after integrating this repo into the main one.

### DIFF
--- a/src/qml/bitcoin.cpp
+++ b/src/qml/bitcoin.cpp
@@ -6,6 +6,7 @@
 
 #include <init.h>
 #include <interfaces/node.h>
+#include <logging.h>
 #include <node/context.h>
 #include <node/ui_interface.h>
 #include <noui.h>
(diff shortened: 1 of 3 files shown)


promag (commented):
Tested ACK 6521cb8f118535bacc4d1d291a97c6d2109855d1.

Locally I comment out `qInstallMessageHandler` in order to see QML warnings.

I wouldn't mind moving `DebugMessageHandler` to src/qt/guiutil.h and then each app would call `GUIUtil::installMessageHandler()`.

jarolrod (commented):
post-merge tested ACK 6521cb8

Tested on macOS 12 and Ubuntu 20.04. Also made sure each commit works independently.

PR #46, opened 2021-10-03
Fixed width for BlockCounter

katesalazar:
Without this, BlockCounter is continuously changing width
unnecessarily. With a fixed width BlockCounter, visual output doesn't
change more than strictly necessary. At current window min size, this
width fits up to 9 digits, which is enough to fit all of main, test
and signature networks.

You can spot the difference these changes add by creating two dummy
data directories and bootstrapping one with this change and
bootstrapping the other one without this change.

### DIFF
--- a/src/qml/components/BlockCounter.qml
+++ b/src/qml/components/BlockCounter.qml
@@ -16,6 +16,6 @@ Label {
     padding: 16
     horizontalAlignment: Text.AlignHCenter
     verticalAlignment: Text.AlignVCenter
-    font.pixelSize: height / 3
+    font.pixelSize: (parent.height / 15 + parent.width / 15) / 2
     text: blockHeight
 }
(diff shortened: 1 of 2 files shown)


katesalazar:
Blocks counter label is anchored to window height,
not to blocks counter height. Set draft.

promag (commented):
This is a dummy component just to show how to bind to the current block height. Not actually something that was designed and eventually it will be removed. I'd say there is no point in this change.

hebasto:
> This is a dummy component just to show how to bind to the current block height. Not actually something that was designed and eventually it will be removed. I'd say there is no point in this change.

Yes. From https://github.com/bitcoin-core/gui-qml/pull/31#issue-982034917:
> adds a _prototype_ of QML object which displays block tip height

hebasto:
@katesalazar 

If you are interested in QML app, you could also look into https://github.com/BitcoinDesign/Bitcoin-Core-App

katesalazar:
> Hope to see more of your contributions into this repo :)

I would love to, not much time to tinker around,
maybe in the future, who knows.

From what I've read these weeks
I think all of these are in good hands
and that's the point that matters to me.

PR #49, opened 2021-10-06
Add regular and semibold typefaces of Inter font

promag:
As requested by @GBKS.


hebasto:
Concept ACK.

To reduce risk of conflicts during syncing with the main repo I'd refrain from the first commit.

PR #58, opened 2021-10-11
ci: Disable binaries for Android task explicitly

hebasto:
In this repo GUI test are disabled implicitly as they make no sense for now.

Therefore, the `test_bitcoin_qt` target in the `src/qt/Makefile` is unavailable, that breaks CI Android task.

This PR fix this issue by explicitly defining a set of binaries via `configure` options, that allows to avoid relying on the `test_bitcoin_qt` target from the `src/qt/Makefile`.


hebasto:
@MarcoFalke 

Is it reasonable to suggest this PR in the main repo?

PR #61, opened 2021-10-20
ci: temporarily disable failing ci task

jarolrod:
Temporarily disable the `32-bit + dash [gui] [CentOS 8]` task. The task currently fails on master and needs a fix: https://cirrus-ci.com/task/4928128318439424

For now, and to not have this distract, let's disable the task. The PR that introduces a fix should also re-enable the task.


hebasto:
Effectively dropped in #121.

PR #65, opened 2021-10-21
Set tab focus behavior

promag:
Sets tab focus behavior to iterate all controls and widgets.

Can be tested after adding some controls to the scene.


hebasto (approved):
ACK 500a6d20be4a5993c7c33bc613f841eca2233642

PR #66, opened 2021-10-22
Fix qt_libbitcoinqt_a_SOURCES

promag:
The correct dependency is QML_QRC_CPP which already depends on QML_QRC 
and QML_RES_QML and probably others later on.


hebasto:
> The correct dependency is QML_QRC_CPP...

Why do you think so?

Considering CI failuers, maybe rebase this PR or convert it into a draft?

hebasto:
Is this still relevant?

PR #80, opened 2021-11-03
build: Use qrc file content as source

promag:
Use `rcc --list` instead of also adding files in the Makefile.

### DIFF
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -296,13 +296,7 @@ QML_RES_FONTS = \
 
 QML_QRC_CPP = qml/qrc_bitcoin.cpp
 QML_QRC = qml/bitcoin_qml.qrc
-QML_RES_QML = \
-  qml/components/BlockCounter.qml \
-  qml/components/ConnectionOptions.qml \
-  qml/controls/OptionButton.qml \
-  qml/controls/ProgressIndicator.qml \
-  qml/pages/initerrormessage.qml \
-  qml/pages/stub.qml
+QML_RES_QML := $(shell $(RCC) --list $(QML_QRC))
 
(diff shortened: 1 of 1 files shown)


promag (commented):
@hebasto @fanquake I have no idea if this is a good approach, but the goal is to avoid adding new files to Makefile, just add to the .qrc file.

hebasto:
Rebase?

jarolrod (commented):
Concept ACK

This approach/idea is better than manually keeping a list of our components and pages. Rebase

promag:
I'll rebase and try to sharpen my lame makefile skillz to make it work.

src/Makefile.qt.include
```diff
@@ -296,12 +296,7 @@ QML_RES_FONTS = \
 
 QML_QRC_CPP = qml/qrc_bitcoin.cpp
 QML_QRC = qml/bitcoin_qml.qrc
-QML_RES_QML = \
-  qml/components/BlockCounter.qml \
-  qml/components/ConnectionOptions.qml \
-  qml/controls/OptionButton.qml \
-  qml/pages/initerrormessage.qml \
-  qml/pages/stub.qml
+$(eval QML_RES_QML = $(shell $(RCC) --list $(QML_QRC)))
```
hebasto, line 299: Maybe
```suggestion
QML_RES_QML :=  $(shell $(RCC) --list $(QML_QRC))
```
?

hebasto: My suggestion does not work due to Automake [limitations](https://www.gnu.org/software/automake/manual/html_node/Conditional-Sources.html).

promag:
Closing since workarounds/hacks are needed to address limitation `configure substitutions are not allowed in _SOURCES variables`.

PR #90, opened 2021-12-17
doc: Add run information to README instructions

secondl1ght:
The README was missing the run step after the build information. When I was setting up my local environment I found this missing step a bit confusing as to how to actually get the QML GUI running, so I think it would be good to include here.

image

### DIFF
--- a/src/qml/README.md
+++ b/src/qml/README.md
@@ -71,3 +71,9 @@ you must configure with the following option:
 ```
 ./configure --with-qml
 ```
+### Run
+
+To run the qml GUI:
+```
+./src/qt/bitcoin-qt
+```


shaavan (commented):
Concept ACK

I tested that the added documentation is correct and works as intended.

Just one recommendation. It would be best to squash the commits together before they can be merged.
Reference: https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#squashing-commits

secondl1ght:
Ok thanks for the review @shaavan, I have now squashed my commits into 1.

src/qml/README.md
```diff
@@ -71,3 +71,9 @@ you must configure with the following option:
 ```
 ./configure --with-qml
 ```
+### Run
+
+To run the GUI QML application execute the file:
+```
+/src/qt/bitcoin-qt
```
hebasto, line 78: Now it refers to an absolute path which is pretty questionable.

Usually, it's preferred to use paths that are relative to the repo's directory:
```
./src/qt/bitcoin-qt
```
or
```
src/qt/bitcoin-qt
```

secondl1ght: The reason I chose not to use `./src/qt/bitcoin-qt` is because technically you could run the file from the src folder or also the qt folder directly by using `./bitcoin-qt`. However I can update it to whichever you prefer, for now I'll remove the initial forward slash to match your second option.

hebasto: Technically, I doubt if someone will place a repository into the system root `/` directory. That is why an absolute path like `/src/qt/bitcoin-qt` is meaningless in sane scenarios.

Btw, `./src/qt/bitcoin-qt` and `src/qt/bitcoin-qt` are equivalent, no?

> ... technically you could run the file from the src folder or also the qt folder directly by using `./bitcoin-qt`.

Yes. If your goal is to document the easiest way that is one I use everyday dozens of times:
```
./autogen.sh
./configure
make
./src/qt/bitcoin-qt
```

In Bash when one starts a command with `./` auto-completion works nice for executables :)

I believe that readers of this file will understand that this way to run Bitcoin Core is on testing purpose only.

secondl1ght: Ok sounds good, I think you have a good point which is that we want to document the most common and easiest way to run the file. I'll make the final change and then squash my commits again. Thanks!

src/qml/README.md
```diff
@@ -71,3 +71,9 @@ you must configure with the following option:
 ```
 ./configure --with-qml
 ```
+### Run
+
+To run the GUI QML application execute the file:
```
shaavan, line 76: 2. To maintain consistency with line no. 68
`To add support for building the qml GUI,`
It would be best if you made the following change to the added message:
```suggestion
To run the qml GUI, execute the file:
```
It is better to keep terminology in our docs consistent, to begin with.

3. This one is a non-essential suggestion. How about mentioning, in the comment message itself, that the command should be run from root directory:
```suggestion
To run the qml GUI, run the following command from the root directory.
```

shaavan (commented):
Changes since my last review:

- Squashed the commits together.
- `src/qt/bitcoin-qt` -> `./src/qt/bitcoin-qt`


Some minor nitpicks: 

1. The line `Remove forward slash from file path`, in the commit message does not seem to refer to any specific change done to this PR. I reckon it could be removed.

secondl1ght:
Ok I have made the adjustments and I think that should be everything now, thanks for the reviews.

shaavan (approved):
ACK 7934def27e9b5746da266af9ec5ff09d73949ce0

Changes since my last review:
- Addressed [this](https://github.com/bitcoin-core/gui-qml/pull/90#pullrequestreview-836035548) comment.

I think this PR is complete and can be merged.

Btw, no need to mark me as a co-author :) Cheers :clinking_glasses:

src/qml/README.md
```diff
@@ -71,3 +71,9 @@ you must configure with the following option:
 ```
 ./configure --with-qml
 ```
+### Run
+
+To run the qml GUI, run the following command from the root directory.
```
jarolrod, line 76: I think the current text isn't optimal because:
1) Generally throughout our documentation, commands are written assuming that you are executing them from the root of the cloned repo. Including the text "... run the following command from the root directory" is redundant information, and actually a bit confusing (system root or repo root?)
2) The text has an awkward flow to it. We are saying "To run A, run B"

this really could just be the following, no more information is needed:
```suggestion
To run the qml GUI:
```

jarolrod (commented):
Concept ACK

secondl1ght:
Sounds good, I am not too attached to any specific wording, whatever the maintainers would like is fine with me. I just want there to be some kind of run command instructions in the README, I have merged the suggestion. Should I squash again?

jarolrod:
@secondl1ght yes, you would need to squash again. We require a clean commit history, see: https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#squashing-commits

secondl1ght:
Sounds good thanks, just squashed again.

hebasto:
> Sounds good, I am not too attached to any specific wording, whatever the _maintainers would like_ is fine with me.

Not maintainers, but reviewers.

Maintainers follow reviewers' opinion.

hebasto (approved):
ACK 2e3f2a45fad6ecfda62bb4c182141f42184532bf

jarolrod:
ACK 2e3f2a4

PR #97, opened 2021-12-23
Allow user to quit GUI from terminal

jarolrod:
Based on #95

This allows a user to quit the GUI from the terminal. It introduces a QTimer in the NodeModel QObject to periodically call the object's `detectShutdown` function.

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/97)
[macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/97)
[Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/97)

PR #96, opened 2021-12-23
[1/2] Allow user to quit GUI from terminal

jarolrod:
Based on #95

This allows a user to quit the GUI from the terminal. It introduces a QTimer in `bitcoin.cpp` to periodically call the NodeModel detectShutdown function. 

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/96)
[macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/96)
[Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/96)

PR #95, opened 2021-12-23
Add node shutdownRequested detection

jarolrod:
Adds a function to our NodeModel QObject that asks the underlying `m_node` if `shutdownRequested` has been set. If it has been set then it calls `startNodeShutdown`, which emits `requestedShutdown`. The emitted signal can then be used to quit the GUI when a user has attempted to close the application through a terminal.

This is part of enabling the GUI to shutdown when a user has attempted to quit through the terminal. This part has been opened separately because I am opening two PR's with different visions on how to fully accomplish the task. Opening this separately helps keep the discussion focused on the location of the `QTimer` in the following PR's.

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/95)
[macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/95)
[Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/95)

PR #108, opened 2022-01-06
Allow user to quit GUI from terminal

hebasto:
This PR is an alternative to bitcoin-core/gui-qml#95, bitcoin-core/gui-qml#96, bitcoin-core/gui-qml#97.

Here are differences:
- used `QObject`'s timer support instead of `QTimer`, it is quite enough
- it works :) -- `node_model.startNodeShutdown()` leads to calling `Shutdown()` twice which is wrong

`NodeModel::startShutdownPolling()` and `NodeModel::stopShutdownPolling()` are parts of the `NodeModel` public interface to cope with `AbortShutdown()` in the future.

### DIFF
--- a/src/qml/bitcoin.cpp
+++ b/src/qml/bitcoin.cpp
@@ -170,7 +170,6 @@ int QmlGuiMain(int argc, char* argv[])
     qGuiApp->setQuitOnLastWindowClosed(false);
     QObject::connect(qGuiApp, &QGuiApplication::lastWindowClosed, [&] {
         node->startShutdown();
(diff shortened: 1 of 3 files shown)


jarolrod (commented):
tACK ea7f523bd62d409d1f9e5c3d65c6a39a0f8f79a4

This is a nice alternative!

It was unintuitive to me why the name is `*_id`, per [Qt Docs](https://doc.qt.io/qt-5/qobject.html#startTimer):
> Starts a timer and returns a timer identifier

PR #112, opened 2022-01-13
qml: Follow-up to PR#101

shaavan:
- This PR is a follow-up to #101 and addresses suggestions of [this](https://github.com/bitcoin-core/gui-qml/pull/101#discussion_r781290059) comment.
- The changes are split into separate commits so that they can be easier to review, and reason with.


promag (commented):
Code review ACK f5458f2c63ef201eddf112729da66d7c1f2f86f1.

PR #118, opened 2022-05-25
Introduce Caret-Left Icon

jarolrod:
Introduces the caret-left icon from the Bitcoin Icons set. Based on #117 to avoid conflict

The current production use-case of this icon is for button such as the `Back Button` which requires this icon. 

The [caret-left svg](https://github.com/BitcoinDesign/Bitcoin-Icons/blob/main/svg/filled/caret-left.svg) was taken from the [Bitcoin-Icons Repo](https://github.com/BitcoinDesign/Bitcoin-Icons). This was used to generate a PNG of 128x128 size using inkscape. The resulting PNG was then run through the [optimize-pngs.py](https://github.com/bitcoin-core/bitcoin-maintainer-tools/blob/master/optimize-pngs.py) script. The output of the script is as follows: 

```
optimizing caret-left.png... done
summary:
+++++++++++++++++
caret-left.png
  size diff from: 664 to: 385
  old sha256: de2632dfb4c3ecbeb1407c896722c052348031fc58f4d876b8b3fe19e946329a
  new sha256: 1b1139cf7e940571eddacb1d63d8a35a91ba011c3a67f4c02f2227fff8ca8fc9

completed. Checksum stable: False. Total reduction: 279 bytes

```

A demo of the icon being displayed can be found here: [caret-left-demo](https://github.com/jarolrod/gui-qml/tree/caret-left-demo)

PR #117, opened 2022-05-25
Introduce info icon

jarolrod:
Introduces the info icon from the Bitcoin Icons set. Since this is the first icon being added, also adds inclusion of icons to builds.

The current production use-case of this icon will be to, when pressed, open up the `About` view within the onboarding process. 

The [info icon svg](https://github.com/BitcoinDesign/Bitcoin-Icons/blob/main/svg/filled/info.svg) was taken from the [Bitcoin-Icons Repo](https://github.com/BitcoinDesign/Bitcoin-Icons). This was used to generate a PNG of 128x128 size using inkscape. The resulting PNG was then run through the [optimize-pngs.py](https://github.com/bitcoin-core/bitcoin-maintainer-tools/blob/master/optimize-pngs.py) script. The output of the script is as follows: 

```
optimizing info.png... done
summary:
+++++++++++++++++
info.png
  size diff from: 528 to: 301
  old sha256: e7f1e22b7ad4447aca0bcdbc0cfe532643b679901b57ef7897f8ee4dee484a5b
  new sha256: 097dcdcd795f88cd5f9d61c131078d73806f5e024994ed09930e43cdd9e4ed80

completed. Checksum stable: False. Total reduction: 227 bytes
```

A demo of the icon being displayed can be found here: [info-icon-demo](https://github.com/jarolrod/gui-qml/tree/info-icon-demo)

PR #119, opened 2022-05-25
Introduce Caret-Right Icon

jarolrod:
Introduces the caret-right icon from the Bitcoin Icons set. Based on #117 and #118 to avoid conflict

The current production use-case of this icon is for settings which open to a new view within the onboarding process.

The [caret-right svg](https://github.com/BitcoinDesign/Bitcoin-Icons/blob/main/svg/filled/caret-right.svg) was taken from the [Bitcoin-Icons Repo](https://github.com/BitcoinDesign/Bitcoin-Icons). This was used to generate a PNG of 128x128 size using inkscape. The resulting PNG was then run through the [optimize-pngs.py](https://github.com/bitcoin-core/bitcoin-maintainer-tools/blob/master/optimize-pngs.py) script. The output of the script is as follows: 

```
optimizing caret-right.png... done
summary:
+++++++++++++++++
caret-right.png
  size diff from: 667 to: 392
  old sha256: e43ba08e9b0b3155c634fe3e7a1521ea1ca56827ca7a08bde505947321ccc06a
  new sha256: 3a54985fb8987cc0c77e1d15f66c9e04b98ef25f11de67d95e76143fe722d170

completed. Checksum stable: False. Total reduction: 275 bytes
```

A demo of the icon being displayed can be found here: [caret-right-demo](https://github.com/jarolrod/gui-qml/tree/caret-right-demo)


hebasto (approved):
ACK 303bb4d9a8aa5ad9d1d4bb72c83fc92f44247137

PR #120, opened 2022-05-25
Introduce Export Icon

jarolrod:
Introduces the export icon from the Bitcoin Icons set. Based on #117, #118, #119 to avoid conflict

The current production use-case of this icon is for buttons that direct user to an outside piece of information.

The [export svg](https://github.com/BitcoinDesign/Bitcoin-Icons/blob/main/svg/filled/export.svg) was taken from the [Bitcoin-Icons Repo](https://github.com/BitcoinDesign/Bitcoin-Icons). This was used to generate a PNG of 128x128 size using inkscape. The resulting PNG was then run through the [optimize-pngs.py](https://github.com/bitcoin-core/bitcoin-maintainer-tools/blob/master/optimize-pngs.py) script. The output of the script is as follows: 

```
optimizing export.png... done
summary:
+++++++++++++++++
export.png
  size diff from: 1275 to: 695
  old sha256: a11ded9350986f287efb794e6d83d37a1d744a16e4bcbb0612b4a6263f00f7b7
  new sha256: ccf02071b9dd73ca45cd367a49cca99d8c5b81168a7f1d0389c1decce30215cf

completed. Checksum stable: False. Total reduction: 580 bytes
```

A demo of the icon being displayed can be found here: [export-icon-demo](https://github.com/jarolrod/gui-qml/tree/export-icon-demo)


hebasto (approved):
ACK 50f979e418416448cd111470d76ac640a04cf81f

PR #125, opened 2022-06-01
Handle undefined active state in Header control

jarolrod:
Currently, when checking if the subtext or description labels should be active, the Header control doesn't  handle the case where, for example, root.description is undefined. Here, we handle the case by initializing description and subtext as empty strings.

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/125)
[macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/125)
[Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/125)

### DIFF
--- a/src/qml/controls/Header.qml
+++ b/src/qml/controls/Header.qml
@@ -13,10 +13,10 @@ Control {
     required property string header
     property int headerMargin
     property int headerSize: 28
-    property string description
+    property string description: ""
     property int descriptionMargin: 10
     property int descriptionSize: 18
-    property string subtext
+    property string subtext: ""
     property int subtextMargin
     property int subtextSize: 15
     property bool wrap: true


src/qml/controls/Header.qml
```diff
@@ -36,7 +36,7 @@ Control {
         Loader {
             Layout.fillWidth: true
             Layout.preferredWidth: 0
-            active: root.description.length > 0
+            active: root.description.length > 0 ?? false
```
hebasto, line 39: Is syntax correct?

> The nullish coalescing operator (`??`) is a logical operator that returns its right-hand side operand when its left-hand side operand is null or undefined, and otherwise returns its left-hand side operand.

I thought that `root.description.length` should be a left-hand side operand, no?

hebasto: According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence it is equivalent to `root.description.length > (0 ?? false)`.

hebasto, line 39:
```suggestion
            active: root.description.length ?? 0 > 0
```

promag: This is not right, you will get an error like "can't read length of undefined".

You want the optional chaining operator (https://github.com/tc39/proposal-optional-chaining) which is not available in Qt 5.15.2, see https://bugreports.qt.io/browse/QTBUG-77926.

jarolrod: you are right, these was some sleepy programming :), hebasto's suggestion here is correct from my understanding: https://github.com/bitcoin-core/gui-qml/pull/125#discussion_r886584266

jarolrod: incorporated in dfbbfb1, thanks!

jarolrod:
updated from[ 982ca36 ](https://github.com/bitcoin-core/gui-qml/commit/982ca36616559843fa0af72c33819807178abc99)to dfbbfb1 ([pr125.01](https://github.com/jarolrod/gui-qml/tree/pr125.01) -> [pr125.02](https://github.com/jarolrod/gui-qml/tree/pr125.02), [diff](https://github.com/jarolrod/gui-qml/compare/pr125.01..pr125.02))

Changes:
- addressed @promag [concern](https://github.com/bitcoin-core/gui-qml/pull/125#discussion_r886626277), by using @hebasto [suggestion](https://github.com/bitcoin-core/gui-qml/pull/125#discussion_r886584266)

hebasto:
Could we achieve the same goal with
```diff
--- a/src/qml/controls/Header.qml
+++ b/src/qml/controls/Header.qml
@@ -13,10 +13,10 @@ Control {
     required property string header
     property int headerMargin
     property int headerSize: 28
-    property string description
+    required property string description
     property int descriptionMargin: 10
     property int descriptionSize: 18
-    property string subtext
+    required property string subtext
     property int subtextMargin
     property int subtextSize: 15
     contentItem: ColumnLayout {
```
?

shaavan (commented):
Concept ACK

- The nullish coalescing is applied correctly in this PR for the `Header.qml` file.
- I was able to compile this PR using the static builds successfully.

I tested if the case of description and subtext not provided is appropriately handled by this PR using the following patch.

```diff
--- a/src/qml/pages/stub.qml
+++ b/src/qml/pages/stub.qml
@@ -34,10 +34,10 @@ ApplicationWindow {
             header: qsTr("Bitcoin Core App")
             headerSize: 36
             headerMargin: 30
-            description: qsTr("Be part of the Bitcoin network.")
+            // description: qsTr("Be part of the Bitcoin network.")
             descriptionSize: 24
             descriptionMargin: 0
-            subtext: qsTr("100% open-source & open-design")
+            // subtext: qsTr("100% open-source & open-design")
             subtextMargin: 25

```
- I got no error or warning messages while running the GUI after applying this patch, which shows that changes are working as per expectation

I want to mention two things that I noticed:
1. You could still provide margin and size for the description/subtext, even when no text is provided. This should technically not be allowed.
2. Applying the patch shown above in the main branch leaded to no warning/error. There was no difference between this PR's compiled GUI and the main's.

shaavan:
@hebasto 

- I don't think making description and subtext required would be an optimal solution.
- There are many instances where we need to use Header but don't have a description or subtext associated with the same.

For example:

- The "Developer Options" heading:
image

- The "Storage settings" heading:
image

hebasto:
@shaavan 
>     * I don't think making description and subtext required would be an optimal solution.
> 
>     * There are many instances where we need to use Header but don't have a description or subtext associated with the same.

Then explicitly define them as empty strings?

shaavan:
@hebasto

> Then explicitly define them as empty strings?

I think that would make using **Header** control a little cumbersome.

However, instead of making _description_ and _subtext_ `required`, what if we define them by default as `""`.

```diff
--- a/src/qml/controls/Header.qml
+++ b/src/qml/controls/Header.qml
@@ -13,10 +13,10 @@ Control {
     required property string header
     property int headerMargin
     property int headerSize: 28
-    property string description
+    property string description: ""
     property int descriptionMargin: 10
     property int descriptionSize: 18
-    property string subtext
+    property string subtext: ""
     property int subtextMargin
```


- This solves the problem of redundantly defining these properties every time, even when they are not used.
- This solves the problem of `root.description` being undefined. Hence no need to check if it is defined.

jarolrod:
updated from dfbbfb1 to cc10229 ([pr125.02](https://github.com/jarolrod/gui-qml/tree/pr125.02) -> [pr125.03](https://github.com/jarolrod/gui-qml/tree/pr125.03))

Changes:
- addressed review feedback on changing the approach to just initialize the fields as empty strings
- rebased over master for ease of testing

shaavan (approved):
ACK cc10229741d09ac2b93daef9d154bd9bb71046df

- A simple and effective solution to handle the undefined states.
- I agree with this approach as per my reasoning in [this](https://github.com/bitcoin-core/gui-qml/pull/125#issuecomment-1153257924) comment.

> - This solves the problem of redundantly defining these properties every time, even when they are not used.
> - This solves the problem of the root.description being undefined. Hence no need to check if it is defined.

PR #127, opened 2022-06-03
Introduce `PeersIndicator` component

hebasto:
The reference design: https://bitcoindesign.github.io/Bitcoin-Core-App/assets/images/block-clock-connection-states-big.png

Design features implementation:
- [x] a single blinking dot
- [x] dots are lit up proportional to the count of the outbound peers

Looking for designers' Concept (N)ACK, and developers' notes regarding QML code.

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/127)
[macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/127)
[Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/127)

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -1344,15 +1344,47 @@ void CConnman::DisconnectNodes()
 
 void CConnman::NotifyNumConnectionsChanged()
 {
-    size_t nodes_size;
+    decltype(m_nodes) nodes_copy;
     {
         LOCK(m_nodes_mutex);
-        nodes_size = m_nodes.size();
+        nodes_copy = m_nodes;
     }
-    if(nodes_size != nPrevNodeCount) {
-        nPrevNodeCount = nodes_size;
+
+    int num_outbound_full_relay{0};
+    int num_block_relay{0};
+    int num_manual{0};
+    int num_inbound{0};
+    for (const auto* node : nodes_copy) {
+        switch (node->m_conn_type) {
+        case ConnectionType::OUTBOUND_FULL_RELAY:
+            ++num_outbound_full_relay;
+            break;
+        case ConnectionType::BLOCK_RELAY:
+            ++num_block_relay;
+            break;
+        case ConnectionType::MANUAL:
+            ++num_manual;
+            break;
+        case ConnectionType::INBOUND:
+            ++num_inbound;
+            break;
+        case ConnectionType::FEELER:
+        case ConnectionType::ADDR_FETCH:
+            break;
+        }
+    }
+
+    if (num_outbound_full_relay != m_num_outbound_full_relay ||
+        num_block_relay != m_num_block_relay ||
+        num_manual != m_num_manual ||
+        num_inbound != m_num_inbound) {
+        m_num_outbound_full_relay = num_outbound_full_relay;
+        m_num_block_relay = num_block_relay;
+        m_num_manual = num_manual;
+        m_num_inbound = num_inbound;
         if (m_client_interface) {
-            m_client_interface->NotifyNumConnectionsChanged(nodes_size);
+            m_client_interface->NotifyNumConnectionsChanged(
+                {num_outbound_full_relay, num_block_relay, num_manual, num_inbound, static_cast<int>(nodes_copy.size())});
         }
     }
 }
--- a/src/qml/components/PeersIndicator.qml
+++ b/src/qml/components/PeersIndicator.qml
@@ -1,0 +1,41 @@
+// Copyright (c) 2022 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+// The PeersIndicator component.
+// See reference design: https://bitcoindesign.github.io/Bitcoin-Core-App/assets/images/block-clock-connection-states-big.png
+
+import QtQuick 2.15
+import "../controls"
+
+Item {
+    id: root
+    required property int numOutboundPeers
+    required property int maxNumOutboundPeers
+
+    implicitWidth: dots.childrenRect.width
+    implicitHeight: dots.childrenRect.height
+
+    Row {
+        id: dots
+        property int size: 5
+        spacing: 5
+        Repeater {
+            model: 5
+            Rectangle {
+                width: 3
+                height: 3
+                radius: width / 2
+                color: Theme.color.neutral9
+                opacity: (index === 0 && root.numOutboundPeers > 0) || (index + 1 <= dots.size * root.numOutboundPeers / root.maxNumOutboundPeers) ? 0.95 : 0.45
+                Behavior on opacity { OpacityAnimator { duration: 100 } }
+                SequentialAnimation on opacity {
+                    loops: Animation.Infinite
+                    running: numOutboundPeers === 0 && index === 0
+                    SmoothedAnimation { to: 0; velocity: 2.2 }
+                    SmoothedAnimation { to: 1; velocity: 2.2 }
+                }
+            }
+        }
+    }
+}
--- a/src/qml/nodemodel.cpp
+++ b/src/qml/nodemodel.cpp
@@ -5,6 +5,8 @@
 #include <qml/nodemodel.h>
 
 #include <interfaces/node.h>
+#include <net.h>
+#include <node/ui_interface.h>
 #include <validation.h>
 
 #include <cassert>
@@ -16,6 +18,7 @@ NodeModel::NodeModel(interfaces::Node& node)
     : m_node{node}
 {
     ConnectToBlockTipSignal();
+    ConnectToNumConnectionsChangedSignal();
 }
 
 void NodeModel::setBlockTipHeight(int new_height)
@@ -23,6 +26,14 @@ void NodeModel::setBlockTipHeight(int new_height)
     if (new_height != m_block_tip_height) {
         m_block_tip_height = new_height;
         Q_EMIT blockTipHeightChanged();
+    }
+}
+
+void NodeModel::setNumOutboundPeers(int new_num)
+{
+    if (new_num != m_num_outbound_peers) {
+        m_num_outbound_peers = new_num;
+        Q_EMIT numOutboundPeersChanged();
     }
 }
 
@@ -73,5 +84,15 @@ void NodeModel::ConnectToBlockTipSignal()
         [this](SynchronizationState state, interfaces::BlockTip tip, double verification_progress) {
             setBlockTipHeight(tip.block_height);
             setVerificationProgress(verification_progress);
+        });
+}
+
+void NodeModel::ConnectToNumConnectionsChangedSignal()
+{
+    assert(!m_handler_notify_num_peers_changed);
+
+    m_handler_notify_num_peers_changed = m_node.handleNotifyNumConnectionsChanged(
+        [this](PeersNumByType new_num_peers) {
+            setNumOutboundPeers(new_num_peers.outbound_full_relay + new_num_peers.block_relay);
         });
 }
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -306,6 +306,7 @@ QML_RES_QML = \
   qml/components/BlockCounter.qml \
   qml/components/ConnectionOptions.qml \
   qml/components/ConnectionSettings.qml \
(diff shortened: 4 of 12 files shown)


src/qml/components/PeersIndicator.qml
```diff
@@ -0,0 +1,34 @@
...
+    id: root
+    property int numOutboundPeers: 0
+    property int maxNumOutboundPeers: 10
+    property int size: 5
+
+    implicitWidth: 64
+    implicitHeight: 16
+
+    Row {
+        spacing: 6
```
jarolrod, line 21:
```suggestion
        spacing: 5
```

src/qml/components/PeersIndicator.qml
```diff
@@ -0,0 +1,34 @@
...
+        Repeater {
+            model: root.size
+            delegate: Rectangle {
+                width: 8
```
jarolrod, line 25:
```suggestion
                width: 3
```

src/qml/components/PeersIndicator.qml
```diff
@@ -0,0 +1,34 @@
...
+                height: 8
```
jarolrod, line 26:
```suggestion
                height: 3
```

jarolrod (changes requested):
the design specifies spacing of 5 between each connection dot, and that each connection dot be 3x3, see: https://www.figma.com/file/GaCoOSNHB2yMB9ThiDtred/Bitcoin-Core-App?node-id=3288%3A70945

With these changes, it'll look a bit small in its current demo. But it should look fine when it's a part of the block clock.

hebasto:
Fixed circular dependency.

> the design specifies spacing of 5 between each connection dot, and that each connection dot be 3x3, see: https://www.figma.com/file/GaCoOSNHB2yMB9ThiDtred/Bitcoin-Core-App?node-id=3288%3A70945

Thanks! Updated.

But not dots look like square pixels :)

cc @GBKS

jarolrod:
@hebasto yes that is true, if we need to scale it, we can scale it from the current proportions. The orginal implementation looked a lot like the PageIndicator we had before.

hebasto:
> we can scale it from the current proportions

Talking about scaling, we should scale down, rather up.

src/qml/components/PeersIndicator.qml
```diff
@@ -0,0 +1,35 @@
...
+// The PeersIndicator component.
+// See reference design: https://bitcoindesign.github.io/Bitcoin-Core-App/assets/images/block-clock-connection-states-big.png
+
+import QtQuick 2.15
+import "../controls"
+
+Item {
+    id: root
+    property int numOutboundPeers: 0
+    property int maxNumOutboundPeers: 10
```
shaavan, line 14: How about making it a readonly property? Should we permit other files to edit this property outside of this file?

```suggestion
    readonly property int maxNumOutboundPeers: 10
```

shaavan (commented):
Concept ACK

- The added code changes effectively introduce the progress indicator.
- I was able to successfully compile and run the PR on Ubuntu 22.04 with Qt version 5.15.3
- As @jarolrod mentioned, following the design guide precisely regarding sizing and spacing leads to dots being relatively smaller than the ones displayed in PR.

Demo:

|According to this PR | According design guide|
|----------------------------|--------------------------------|
|Screencast from 04-06-22 01 35 41 AM IST|Screencast from 04-06-22 01 37 59 AM IST|   

I have attached my notes for this PR.

**Notes:**

1. The first commit replaces the “notify only about the total number of nodes change” to “notify about change in any type of nodes”. This change is necessary for the progress indicator component.
    1. This is done by introducing a new struct PeersNumByType, which tracks all the types of nodes along with the total number of nodes.
2. The second commit introduces the variable to track the number of outbound peers connected and the total number of outbound peers that can connect.
3. The third commit introduces the PageIndicator.qml file, which utilizes the above two variables to display a real-time indicator for the number of peers connected to our node.
4. Commit four, including the PageIndicator.qml to the stub.qml file, allowing it to be displayed on the screen.

src/qml/components/PeersIndicator.qml
jarolrod: this can change when there's configuration and options handling, it's set to this value by default

src/net.cpp
```diff
@@ -1344,15 +1344,47 @@ void CConnman::DisconnectNodes()
...
+        nodes_copy = m_nodes;
     }
-    if(nodes_size != nPrevNodeCount) {
-        nPrevNodeCount = nodes_size;
+
+    int num_outbound_full_relay{0};
+    int num_block_relay{0};
+    int num_manual{0};
+    int num_inbound{0};
+    for (const auto* node : nodes_copy) {
```
promag, line 1357: 36e20da8867a73bdae51e211b101ffc0cbf4b55f

Correct me if I'm wrong but I think it's safe to use `node` without `AddRef`/`Release` because `DeleteNode` isn't called concurrently with `CConnman::NotifyNumConnectionsChanged`.

src/qml/components/PeersIndicator.qml
```diff
@@ -0,0 +1,35 @@
...
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+// The PeersIndicator component.
+// See reference design: https://bitcoindesign.github.io/Bitcoin-Core-App/assets/images/block-clock-connection-states-big.png
+
+import QtQuick 2.15
+import "../controls"
+
+Item {
```
promag, line 11: b0b36d6668bcf2d14438b19d849de650951ac02f

This can be `Row`? Then can drop `width` and `height` bindings below.

Otherwise below should bind `implicitWidth` and `implicitHeight`.

hebasto: > This can be `Row`?

Yes, it can. But using `Item` provides better encapsulation and hiding implementation details. For example, `PeersIndicator { padding: 42 }` will fail in the current branch.

hebasto:
Updated 88b99c9a543bc0527a30c1a5d1051fddd3511513 -> aad0ff4d01c3a6fc1100d9717a21c4bce866c159 ([pr127.02](https://github.com/hebasto/gui-qml/commits/pr127.02) -> [pr127.03](https://github.com/hebasto/gui-qml/commits/pr127.03), [diff](https://github.com/hebasto/gui-qml/compare/pr127.02..pr127.03)):

- addressed @promag's [comment](https://github.com/bitcoin-core/gui-qml/pull/127#discussion_r889514484):

> Otherwise below should bind `implicitWidth` and `implicitHeight`.

src/qml/components/PeersIndicator.qml
hebasto: > Otherwise below should bind `implicitWidth` and `implicitHeight`.

Thanks! [Updated](https://github.com/bitcoin-core/gui-qml/pull/127#issuecomment-1146592288).

src/net.cpp
hebasto: Do you mind elaborating? What change do you suggest?

promag: That's my point, I think the correct code is fine. Initially, I wondered if some `AddRef`/`Release` was needed.

src/qml/nodemodel.cpp
```diff
@@ -68,10 +78,16 @@ void NodeModel::timerEvent(QTimerEvent* event)
...
     assert(!m_handler_notify_block_tip);
+    assert(!m_handler_notify_num_peers_changed);
 
     m_handler_notify_block_tip = m_node.handleNotifyBlockTip(
         [this](SynchronizationState state, interfaces::BlockTip tip, double verification_progress) {
             setBlockTipHeight(tip.block_height);
             setVerificationProgress(verification_progress);
         });
+
+    m_handler_notify_num_peers_changed = m_node.handleNotifyNumConnectionsChanged(
```
jarolrod, line 89: why lump this in here, isn't the following cleaner?

``` diff
diff --git a/src/qml/nodemodel.cpp b/src/qml/nodemodel.cpp
index 533baf5ef..3e2232c8a 100644
--- a/src/qml/nodemodel.cpp
+++ b/src/qml/nodemodel.cpp
@@ -18,6 +18,7 @@ NodeModel::NodeModel(interfaces::Node& node)
     : m_node{node}
 {
     ConnectToBlockTipSignal();
+    ConnectToNumConnectionsChangedSignal();
 }
 
 void NodeModel::setBlockTipHeight(int new_height)
@@ -78,16 +79,20 @@ void NodeModel::timerEvent(QTimerEvent* event)
 void NodeModel::ConnectToBlockTipSignal()
 {
     assert(!m_handler_notify_block_tip);
-    assert(!m_handler_notify_num_peers_changed);
 
     m_handler_notify_block_tip = m_node.handleNotifyBlockTip(
         [this](SynchronizationState state, interfaces::BlockTip tip, double verification_progress) {
             setBlockTipHeight(tip.block_height);
             setVerificationProgress(verification_progress);
         });
+}
 
-    m_handler_notify_num_peers_changed = m_node.handleNotifyNumConnectionsChanged(
-        [this](PeersNumByType new_num_peers) {
-            setNumOutboundPeers(new_num_peers.outbound_full_relay + new_num_peers.block_relay);
-        });
+void NodeModel::ConnectToNumConnectionsChangedSignal()
+{
+      assert(!m_handler_notify_num_peers_changed);
+
+      m_handler_notify_num_peers_changed = m_node.handleNotifyNumConnectionsChanged(
+          [this](PeersNumByType new_num_peers) {
+              setNumOutboundPeers(new_num_peers.outbound_full_relay + new_num_peers.block_relay);
+          });
 }
diff --git a/src/qml/nodemodel.h b/src/qml/nodemodel.h
index f81fce7a4..2cd3661f2 100644
--- a/src/qml/nodemodel.h
+++ b/src/qml/nodemodel.h
@@ -72,6 +72,7 @@ private:
     std::unique_ptr<interfaces::Handler> m_handler_notify_num_peers_changed;
 
     void ConnectToBlockTipSignal();
+    void ConnectToNumConnectionsChangedSignal();
 };
 
 #endif // BITCOIN_QML_NODEMODEL_H
```

hebasto:
Updated aad0ff4d01c3a6fc1100d9717a21c4bce866c159 -> 179ca2370afc154944fafc8b651a3cf4e7ec7772 ([pr127.03](https://github.com/hebasto/gui-qml/commits/pr127.03) -> [pr127.04](https://github.com/hebasto/gui-qml/commits/pr127.04), [diff](https://github.com/hebasto/gui-qml/compare/pr127.03..pr127.04)):

- addressed @jarolrod's [comment](https://github.com/bitcoin-core/gui-qml/pull/127#discussion_r889737467)
- added blinking of the first dot according to the design
- other small improvements (I hope)

src/qml/nodemodel.cpp
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/gui-qml/pull/127#issuecomment-1146895205).

jarolrod:
concept ACK on node changes, and ACK on QML changes. Node changes seem ready to open upstream

src/qml/components/PeersIndicator.qml
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2022 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+// The PeersIndicator component.
```
jarolrod, line 5: I guess there's no need for this line, just look at the path and name of the file

src/qml/components/PeersIndicator.qml
```diff
@@ -0,0 +1,41 @@
...
+// See reference design: https://bitcoindesign.github.io/Bitcoin-Core-App/assets/images/block-clock-connection-states-big.png
```
jarolrod, line 6: Perhaps every component with a one to one link to a design file should have a link for easy reference to the design spec?

jarolrod:
replaced by https://github.com/bitcoin-core/gui-qml/pull/236

PR #111, opened 2022-01-13
qml: Add DropShadow to the OptionSwitch indicatorButton

shaavan:
- This PR is a follow-up to #101 and addresses the suggestions given there after it was merged.
- This PR also adds the **DropShadow** for the **OptionSwitch**'s `indicatorButton`, as per the [main design file](https://www.figma.com/file/GaCoOSNHB2yMB9ThiDtred/Bitcoin-Core-App-Main?node-id=1035%3A1883) of the project.

Some points to note:

- Used `===` instead of `==` to avoid type coercion.
- Chose sample value based on official recommended guidelines (sample = 2*radius + 1) https://doc.qt.io/qt-5/qml-qtgraphicaleffects-dropshadow.html#samples-prop
- Used rgba to hex convertor for shadow color: https://rgbacolorpicker.com/rgba-to-hex

PR #45, opened 2021-10-03
qml, build, doc: Allow import QtQuick.Layouts

hebasto:
See #39:
> ... we could at least add quick layouts.

PR #132, opened 2022-06-20
qml: Add DropShadow to the OptionSwitch indicatorButton

shaavan:
Ported #111 to this PR.

This PR rebase #111 on the current master.

Notes:
- I successfully compiled the updated PR on Ubuntu 22.04 using `Qt version 5.15.3`.
- To successfully compile and run the PR, I had to install an additional library.
```
sudo apt install qml-module-qtgraphicaleffects
```
- I have created a convenient demo for checking that the drop shadow works correctly. [Drop Shadow onboarding test](https://github.com/shaavan/gui-qml/tree/Drop-Shadow-Onboarding-Test) branch cherry-picks this PR over the #124. 
- If, in the demo, the color of the drop shadow is too subtle to perceive, try experimenting with changing the color of the drop shadow to get high contrast. For example:
```diff
             spread: 0.0
             samples: 21
-            color: "#00000040"
+            color: "#7900fffc"
             source: indicatorButton
         }
     }
```

### DIFF
--- a/src/qml/bitcoin.cpp
+++ b/src/qml/bitcoin.cpp
@@ -42,6 +42,8 @@ QT_END_NAMESPACE
 
 #if defined(QT_STATICPLUGIN)
 #include <QtPlugin>
+Q_IMPORT_PLUGIN(QtGraphicalEffectsPlugin)
+Q_IMPORT_PLUGIN(QtGraphicalEffectsPrivatePlugin)
 Q_IMPORT_PLUGIN(QtQmlPlugin)
 Q_IMPORT_PLUGIN(QtQmlModelsPlugin)
 Q_IMPORT_PLUGIN(QtQuick2Plugin)
--- a/src/qml/controls/OptionSwitch.qml
+++ b/src/qml/controls/OptionSwitch.qml
@@ -4,6 +4,7 @@
 
 import QtQuick 2.15
 import QtQuick.Controls 2.15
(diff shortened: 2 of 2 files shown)


jarolrod (commented):
CI is failing because you have to allow static builds to pull in these plugins, here's an example on how to accomplish this: https://github.com/bitcoin-core/gui-qml/pull/45/files

hebasto:
@shaavan 

Could you post screenshots please?

Also I'd like to have some Concept ACKs from designers before introducing a new plugin.

src/qml/bitcoin.cpp
```diff
@@ -42,6 +42,8 @@ QT_END_NAMESPACE
 
 #if defined(QT_STATICPLUGIN)
 #include <QtPlugin>
+Q_IMPORT_PLUGIN(QtGraphicalEffectsPlugin)
+Q_IMPORT_PLUGIN(QtGraphicalEffectsPrivatePlugin)
```
hebasto, line 46: Why importing of private plugin is required?

shaavan:
In the recent **Bitcoin Core App Design Call** on 06-29-2022, it was discussed the need to introduce a dependency `qml-module-qtgraphicaleffects` for the **DropShadow** is not worthwhile. And hence, adding DropShadow should be postponed until the same dependency is used for some other important task.

So I shall be closing this PR until later.

PR #144, opened 2022-08-03
Use caret-right icon instead of greater-than text symbol

jarolrod:
Mainly only relevant in the `AboutOptions` component for now.

| master | pr |
| ------- | -- |
| Screen Shot 2022-08-03 at 6 45 51 AM | Screen Shot 2022-08-06 at 7 40 39 PM |

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/144)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/144)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/144)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/144)

### DIFF
--- a/src/qml/components/AboutOptions.qml
+++ b/src/qml/components/AboutOptions.qml
@@ -13,42 +13,56 @@ ColumnLayout {
         Layout.fillWidth: true
         header: qsTr("Website")
         actionItem: ExternalLink {
-            description: qsTr("bitcoincore.org >")
+            description: qsTr("bitcoincore.org")
             link: "https://bitcoincore.org"
+            iconSource: "image://images/caret-right"
+            iconWidth: 18
+            iconHeight: 18
         }
     }
     Setting {
         Layout.fillWidth: true
         header: qsTr("Source code")
         actionItem: ExternalLink {
-            description: qsTr("github.com/bitcoin/bitcoin >")
+            description: qsTr("github.com/bitcoin/bitcoin")
             link: "https://github.com/bitcoin/bitcoin"
+            iconSource: "image://images/caret-right"
+            iconWidth: 18
+            iconHeight: 18
         }
     }
     Setting {
         Layout.fillWidth: true
         header: qsTr("License")
         actionItem: ExternalLink {
-            description: qsTr("MIT >")
+            description: qsTr("MIT")
             link: "https://opensource.org/licenses/MIT"
+            iconSource: "image://images/caret-right"
+            iconWidth: 18
+            iconHeight: 18
         }
     }
     Setting {
         Layout.fillWidth: true
         header: qsTr("Version")
         actionItem: ExternalLink {
-            description: qsTr("v22.99.0-1e7564eca8a6 >")
+            description: qsTr("v22.99.0-1e7564eca8a6")
             link: "https://bitcoin.org/en/download"
+            iconSource: "image://images/caret-right"
+            iconWidth: 18
+            iconHeight: 18
         }
     }
     Setting {
         Layout.fillWidth: true
         header: qsTr("Developer options")
         description: qsTr("Only use these if you have development experience")
-        actionItem: TextButton {
-            text: ">"
-            bold: false
-            rightalign: true
+        actionItem: Button {
+            icon.source: "image://images/caret-right"
+            icon.color: Theme.color.neutral9
+            icon.height: 18
+            icon.width: 18
+            background: null
             onClicked: {
                 introductions.incrementCurrentIndex()
                 swipeView.inSubPage = true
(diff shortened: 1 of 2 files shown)


shaavan (commented):
Concept ACK 

- I can see the visual difference between master and PR while testing.

**Screenshot:**

|Master|PR|
|---------|----|
|Screenshot from 2022-08-03 21-04-59|Screenshot from 2022-08-03 21-05-27|

- This PR also solves the issue of the misaligned arrow of the developer options selector.

**Suggestion:**

- There is one more instance of setting which need to be addressed here. The ">" for the Proxy settings option on the Connection Settings page should be converted to the icon.

Rspigler:
tACK commit 1c17b774e6363c53ff7e1d6249ee9abb10063f7c

I can see the change and it is all lined up on my machine.  The one missing the change is the Proxy setting under 'Connection Settings' as @shaavan mentions

jarolrod:
updated from from 1c17b77 to 1dcdac6 ([pr144.01](https://github.com/jarolrod/gui-qml/tree/pr144.01) -> [pr144.02](https://github.com/jarolrod/gui-qml/tree/pr144.02), [diff](https://github.com/jarolrod/gui-qml/compare/pr144.01..pr144.02))

Changes: 
- rebased over changes in PR's this PR is based on
- Updated icon sources to use `image://` syntax instead of `qrc` syntax, as the `image://` syntax is required for the image provider to work and provide a proper resolution image. Comparison below (top is `image://` and below that is `qrc`:

Screen Shot 2022-08-06 at 6 58 51 PM

jarolrod:
updated from [1dcdac6 ](https://github.com/bitcoin-core/gui-qml/commit/1dcdac6575eaecea07a59694435541e5f106808e)to [36bc09d ](https://github.com/bitcoin-core/gui-qml/commit/36bc09d02b4f08b3e72ed3ef79c53b7cb8971e6d)[Compare](https://github.com/bitcoin-core/gui-qml/compare/1dcdac6575eaecea07a59694435541e5f106808e..36bc09d02b4f08b3e72ed3ef79c53b7cb8971e6d)

Changes: give the developer options setting button the correct action, where it will actually open the developer options subpage

jarolrod:
updated from [36bc09d ](https://github.com/bitcoin-core/gui-qml/commit/36bc09d02b4f08b3e72ed3ef79c53b7cb8971e6d)to [f9f6921](https://github.com/bitcoin-core/gui-qml/commit/f9f692136f441ecee2686424c7fc38593bf44c82)

changes: updated over changes on master

src/qml/components/AboutOptions.qml
```diff
@@ -13,42 +13,56 @@ ColumnLayout {
...
-        actionItem: TextButton {
-            text: ">"
-            bold: false
-            rightalign: true
+        actionItem: Button {
+            icon.source: "image://images/caret-right"
+            icon.color: Theme.color.neutral9
+            icon.height: 18
+            icon.width: 18
+            background: null
```
hebasto, line 65: Why are the `icon.color` and `background` properties are specified for some elements, but not for others?

shaavan: Button is a qml type, so we need to specify the `icon.color`, background value explicitly for each time the Button is used as `actionItem`, which this PR does.

Whereas, the `ExternalLink` is a custom-defined component that already employs the respective icon.color, and background value in its definition, so we need not specify that separately when using it.

In `src/qml/controls/ExternalLink.qml`

```qml
            sourceComponent: Button {
                icon.source: root.iconSource
                icon.color: Theme.color.neutral9
                icon.height: root.iconHeight
                icon.width: root.iconWidth
                background: null
                onClicked: Qt.openUrlExternally(link)
            }
```

shaavan (commented):
ACK f9f692136f441ecee2686424c7fc38593bf44c82

The `ExternalLink` and `Button` components replace all the instances of _greater-than text symbol_ and are used appropriately.

I have two ideas to discuss that we can consider incorporating in this PR or in a follow-up.

1. The `Button{}` is used twice as an `actionItem`  in the current implementation of GUI. There is a chance that we could need to use this Button configuration multiple times as the development progresses. So we can think about extracting this code as a separate component.
2. There is only one instance of usage of `ExternalLink{}` where the default values for `iconWidth`, and `iconHeight` `(30, 30)` are used. All other usage instances modify these values to `(18, 18)`. Changing the default values in `ExternalLink.qml` for these properties will save up on redundant code.

PR #161, opened 2022-08-15
Fix options switch alignment within settings

jarolrod:
Fixes alignment of the option switch within a settings row.

| master | PR |
| --------- | ---- |
| master-option | fix-option-pr |

Closes bitcoin-core/gui-qml#159

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/161)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/161)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/161)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/161)

PR #169, opened 2022-09-08
qml: split background and indicator on option switch

promag:
Alternative to bitcoin-core/gui-qml#161.

nit, it also animates the indicator.

Links for Windows and macOS build artifacts

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/169)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/169)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/169)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/169)

### DIFF
--- a/src/qml/controls/OptionSwitch.qml
+++ b/src/qml/controls/OptionSwitch.qml
@@ -7,21 +7,23 @@ import QtQuick.Controls 2.15
 
 Switch {
     id: root
-    indicator: Rectangle {
-        implicitWidth: 45
-        implicitHeight: 28
-        x: root.leftPadding
-        y: Math.round((parent.height - height) / 2)
-        radius: 18
+    implicitWidth: 45
+    implicitHeight: 28
+    background: Rectangle {
+        radius: Math.floor(height / 2)
         color: root.checked ? Theme.color.orange : Theme.color.neutral4
-        Rectangle {
-            id: indicatorButton
-            y: parent.height / 2 - height / 2
-            x: root.checked ? parent.width - width - 4 : 0 + 4
-            width: 20
-            height: 20
-            radius: 18
-            color: Theme.color.white
+    }
+    indicator: Rectangle {
+        property real _margin: Math.round((parent.height - height) / 2)
+        y: Math.round(parent.height - height) / 2
+        x: root.checked ? parent.width - width - _margin : _margin
+        radius: 10
+        width: 20
+        height: 20
+        color: Theme.color.white
+        Behavior on x {
+            SmoothedAnimation {
+            }
         }
     }
 }


jarolrod (approved):
ACK c50e6bc3ac822306bbae84c69fcc059bb20f59bd

Confirmed that this fixes the issue, and is a better approach than 161.

| master | 161 | 169 |
| ------- | --- | ---- |
| master |  |  |

The indicator is indeed now animated but I like it. It's not clear wether we wanted some type of animation or not, what better animation than a free one? CC: @GBKS @mouxdesign

src/qml/controls/OptionSwitch.qml
```diff
@@ -7,21 +7,23 @@ import QtQuick.Controls 2.15
...
-            width: 20
-            height: 20
-            radius: 18
-            color: Theme.color.white
+    }
+    indicator: Rectangle {
+        property real _margin: Math.round((parent.height - height) / 2)
+        y: Math.round(parent.height - height) / 2
+        x: root.checked ? parent.width - width - _margin : _margin
+        radius: 10
```
shaavan, line 20:
```suggestion
        radius: 18
```

Though changing this value does not indicate visual difference, I think it is still better to make this change as this matches the value given in the [design Figma file](https://www.figma.com/file/GaCoOSNHB2yMB9ThiDtred/Bitcoin-Core-App).

shaavan (commented):
I prefer this PR's approach over #161, as this creates proper segregation between the switch's background and the indicator.
The added animation is clearly noticeable, and I like the smoothness it provides while toggling switch states. However, I would like to listen to the designers' take on it as well before giving my final ACK on this.

src/qml/controls/OptionSwitch.qml
promag: > 10 makes no sense though?

shaavan: Yes, it doesn't, and I agree with that.
But I feel a little iffy about having the value inconsistent with the design file.

However, this is just a tiny nit and can be safely ignored if you want.

promag: @GBKS thoughts?

GBKS: In Figma, the small circle in the toggle is 20px tall, the big one is 28px tall. So the radii of 10px and 14px would be the minimum.

I sometimes set them to 500px or some other ridiculously high number because it doesn't matter in Figma if you're over the exact value. But if you want to be precise, then half the height is what it should be. Looks like for the smaller size, I simply kept the value of the larger one. 10px would be the precise number.

PR #171, opened 2022-09-17
Adjust Onboarding's ContinueButton layout based on Mobile/Desktop target

johnny9:
This change registers a new QObject that can be used to check if the app is running in DESKTOP or MOBILE mode. Based on that value, adjustments to the ContinueButton's anchors will be made to match the spec for each configuration. 

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/171)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/171)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/171)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/171)

### DIFF
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -35,6 +35,7 @@ QT_FORMS_UI = \
   qt/forms/transactiondescdialog.ui
 
 QT_MOC_CPP = \
+  qml/moc_appmode.cpp \
   qml/moc_nodemodel.cpp \
   qt/moc_addressbookpage.cpp \
   qt/moc_addresstablemodel.cpp \
@@ -107,6 +108,7 @@ QT_QRC_LOCALE_CPP = qt/qrc_bitcoin_locale.cpp
 QT_QRC_LOCALE = qt/bitcoin_locale.qrc
 
 BITCOIN_QT_H = \
+  qml/appmode.h \
   qml/bitcoin.h \
   qml/imageprovider.h \
   qml/nodemodel.h \
--- a/src/qml/bitcoin.cpp
+++ b/src/qml/bitcoin.cpp
@@ -11,6 +11,7 @@
 #include <logging.h>
 #include <node/interface_ui.h>
 #include <noui.h>
+#include <qml/appmode.h>
 #include <qml/imageprovider.h>
 #include <qml/nodemodel.h>
 #include <qml/util.h>
@@ -183,6 +184,14 @@ int QmlGuiMain(int argc, char* argv[])
     engine.addImageProvider(QStringLiteral("images"), new ImageProvider{network_style.data()});
 
     engine.rootContext()->setContextProperty("nodeModel", &node_model);
+
+#ifdef __ANDROID__
+    AppMode app_mode(AppMode::MOBILE);
+#else
+    AppMode app_mode(AppMode::DESKTOP);
+#endif // __ANDROID__
+
+    qmlRegisterSingletonInstance<AppMode>("org.bitcoincore.qt", 1, 0, "AppMode", &app_mode);
 
     engine.load(QUrl(QStringLiteral("qrc:///qml/pages/main.qml")));
     if (engine.rootObjects().isEmpty()) {
--- a/src/qml/controls/OnboardingInfo.qml
+++ b/src/qml/controls/OnboardingInfo.qml
@@ -5,8 +5,9 @@
 import QtQuick 2.15
 import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.15
+import org.bitcoincore.qt 1.0
 
-Control {
+Item {
     id: root
     property alias banner: banner_loader.sourceComponent
     required property string buttonText
@@ -25,7 +26,10 @@ Control {
 
     implicitWidth: 600
 
-    contentItem: ColumnLayout {
+    ColumnLayout {
+        anchors.horizontalCenter: parent.horizontalCenter
+        width: parent.width
+        id: information
         spacing: 0
         Loader {
             id: banner_loader
@@ -49,11 +53,40 @@ Control {
             subtextMargin: root.subtextMargin
             subtextSize: root.subtextSize
         }
-        ContinueButton {
-            Layout.alignment: Qt.AlignCenter
-            Layout.topMargin: 40
-            text: root.buttonText
-            onClicked: swipeView.incrementCurrentIndex()
-        }
     }
+    ContinueButton {
+        id: continueButton
+        anchors.horizontalCenter: parent.horizontalCenter
+        anchors.topMargin: 40
+        anchors.bottomMargin: 60
+        anchors.leftMargin: 20
+        anchors.rightMargin: 20
+        text: root.buttonText
+        onClicked: swipeView.incrementCurrentIndex()
+    }
+
+    state: AppMode.state
+
+    states: [
+        State {
+            name: "MOBILE"
+            AnchorChanges {
+                target: continueButton
+                anchors.top: undefined
+                anchors.bottom: continueButton.parent.bottom
+                anchors.right: continueButton.parent.right
+                anchors.left: continueButton.parent.left
+            }
+        },
+        State {
+            name: "DESKTOP"
+            AnchorChanges {
+                target: continueButton
+                anchors.top: information.bottom
+                anchors.bottom: undefined
+                anchors.left: undefined
+                anchors.right: undefined
+            }
+        }
+    ]
 }
--- a/src/qml/appmode.h
+++ b/src/qml/appmode.h
@@ -1,0 +1,49 @@
+// Copyright (c) 2022 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef BITCOIN_QML_APPMODE_H
+#define BITCOIN_QML_APPMODE_H
+
+#include <QObject>
+
+class AppMode : public QObject
+{
+    Q_OBJECT
+    Q_PROPERTY(bool isDesktop READ isDesktop NOTIFY modeChanged)
+    Q_PROPERTY(bool isMobile READ isMobile NOTIFY modeChanged)
+    Q_PROPERTY(QString state READ state NOTIFY modeChanged)
+
+public:
+    enum Mode {
+        DESKTOP,
+        MOBILE
+    };
+
+    explicit AppMode(Mode mode) : m_mode(mode)
+    {
+    }
+
+    bool isMobile() { return m_mode == MOBILE; }
+    bool isDesktop() { return m_mode == DESKTOP; }
+    QString state()
+    {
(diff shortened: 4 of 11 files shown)


johnny9:
After looking through this a bit, I believe the mobile and desktop specs are close enough that most of the top level layouts can be reused without having to provide a different root QML for mobile.  Certain components, however, will end up with more complex anchors/layout properties due to having two different states they can be in.

**DESKTOP:**
onboarding_1

**MOBILE:**
Screenshot_20220916-212819

jarolrod:
Concept ACK on fixing up the UI to fit our mobile design

src/qml/bitcoin.cpp
```diff
@@ -13,6 +13,7 @@
 #include <noui.h>
 #include <qml/imageprovider.h>
 #include <qml/nodemodel.h>
+#include <qml/appmode.h>
```
jarolrod, line 16: For 872807a05daeae9752a39df275a3aff602c942e4, please make the following style changes:

``` diff
diff --git a/src/qml/appmode.h b/src/qml/appmode.h
index 0148a4240..e45977318 100644
--- a/src/qml/appmode.h
+++ b/src/qml/appmode.h
@@ -15,14 +15,12 @@ class AppMode : public QObject
     Q_PROPERTY(QString state READ state NOTIFY modeChanged)
 
 public:
-    enum Mode
-    {
+    enum Mode {
         DESKTOP,
         MOBILE
     };
 
-    explicit AppMode(Mode mode) :
-        m_mode(mode)
+    explicit AppMode(Mode mode) : m_mode(mode)
     {
     }
 
@@ -31,12 +29,12 @@ public:
     QString state()
     {
         switch (m_mode) {
-            case MOBILE:
-                return "MOBILE";
-            case DESKTOP:
-                return "DESKTOP";
-            default:
-                return "DESKTOP";
+        case MOBILE:
+            return "MOBILE";
+        case DESKTOP:
+            return "DESKTOP";
+        default:
+            return "DESKTOP";
         }
     }
     Mode mode() const { return m_mode; }
diff --git a/src/qml/bitcoin.cpp b/src/qml/bitcoin.cpp
index c8687b5ed..291b39f53 100644
--- a/src/qml/bitcoin.cpp
+++ b/src/qml/bitcoin.cpp
@@ -11,9 +11,9 @@
 #include <logging.h>
 #include <node/interface_ui.h>
 #include <noui.h>
+#include <qml/appmode.h>
 #include <qml/imageprovider.h>
 #include <qml/nodemodel.h>
-#include <qml/appmode.h>
 #include <qml/util.h>
 #include <qt/guiconstants.h>
 #include <qt/guiutil.h>
```

You can get the same diff locally by running the following command on 872807a05daeae9752a39df275a3aff602c942e4
```
git diff -U0 HEAD~1.. | ./contrib/devtools/clang-format-diff.py -p1 -i -v
```

src/Makefile.qt.include
```diff
@@ -110,6 +111,7 @@ BITCOIN_QT_H = \
   qml/bitcoin.h \
   qml/imageprovider.h \
   qml/nodemodel.h \
+  qml/appmode.h \
```
jarolrod, line 114: please keep the list in alphabetical order, so we would make the following change to the file:

``` diff
diff --git a/src/Makefile.qt.include b/src/Makefile.qt.include
index 096cd116e..06b642d93 100644
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -35,8 +35,8 @@ QT_FORMS_UI = \
   qt/forms/transactiondescdialog.ui
 
 QT_MOC_CPP = \
-  qml/moc_nodemodel.cpp \
   qml/moc_appmode.cpp \
+  qml/moc_nodemodel.cpp \
   qt/moc_addressbookpage.cpp \
   qt/moc_addresstablemodel.cpp \
   qt/moc_askpassphrasedialog.cpp \
@@ -108,10 +108,10 @@ QT_QRC_LOCALE_CPP = qt/qrc_bitcoin_locale.cpp
 QT_QRC_LOCALE = qt/bitcoin_locale.qrc
 
 BITCOIN_QT_H = \
+  qml/appmode.h \
   qml/bitcoin.h \
   qml/imageprovider.h \
   qml/nodemodel.h \
-  qml/appmode.h \
   qml/util.h \
   qt/addressbookpage.h \
   qt/addresstablemodel.h \
```

jarolrod (commented):
will continue testing, but some style fixups for now

johnny9:
Updated from  6412a3a to  1d10e8e

Changes:
 - fixed style issues in appmode.h
 - fixed include oridering in Makefile.qt.include to alphabetical
 - removed width changes to onboarding01-03 that are meant for a future commit

src/qml/bitcoin.cpp
johnny9: Thanks, that's really good info on the cpp style too.

src/qml/bitcoin.cpp
```diff
@@ -184,6 +185,14 @@ int QmlGuiMain(int argc, char* argv[])
 
     engine.rootContext()->setContextProperty("nodeModel", &node_model);
 
+#ifdef __ANDROID__
+    AppMode app_mode(AppMode::MOBILE);
+#else
+    AppMode app_mode(AppMode::DESKTOP);
+#endif // __ANDROID__
+
+    qmlRegisterSingletonInstance<AppMode>("com.bitcoincore.qt", 1, 0, "AppMode", &app_mode);
```
johnny9, line 194: not 100% sure that is the namespace that should be used but it is what was defined in the AndroidManifest.xml also.  org.bitcoin.qt might make more sense for both.

jarolrod:
Confirming this does as the description states

This pr on an android tablet:

| | | |
| - | - | - |
| Screenshot_20220920-210432 | Screenshot_20220920-210523 | Screenshot_20220920-210532 |

This pr on desktop:

| | | |
| - | - | - |
| Screen Shot 2022-09-20 at 9 55 57 PM | Screen Shot 2022-09-20 at 9 56 03 PM | Screen Shot 2022-09-20 at 9 56 07 PM |

src/qml/bitcoin.cpp
jarolrod: @hebasto what do you think about this URI namespace, see: https://doc.qt.io/qt-6/qqmlengine.html#qmlRegisterSingletonType

johnny9:
Updated  from 1d10e8e to eaba2d4 
Changes:
  - added newline to end of appmode.h
  - squashed commits

johnny9:
Updated from[ eaba2d4 ](https://github.com/bitcoin-core/gui-qml/commit/eaba2d4f6e52c690ab37193b1ede3173129c5274)to[ cd7f5a8](https://github.com/bitcoin-core/gui-qml/commit/cd7f5a837887366175be7ce74586bdb0fe46b6bf): 

  - Rebased with main

johnny9:
Updated from[ cd7f5a8 ](https://github.com/bitcoin-core/gui-qml/commit/cd7f5a837887366175be7ce74586bdb0fe46b6bf)to 8a3ee65

Changes:
  - Set the qml module namespace to org.bitcoincore.qt

src/qml/controls/OnboardingInfo.qml
```diff
@@ -49,11 +53,40 @@ Control {
...
+        anchors.bottomMargin: 60
+        anchors.leftMargin: 20
+        anchors.rightMargin: 20
+        text: root.buttonText
+        onClicked: swipeView.incrementCurrentIndex()
+    }
+
+    state: AppMode.state
+
+    states: [
```
jarolrod, line 70: An alternative would be to define anchor values for components inside of our theme file, and assign values in the same way we assign the color values; but when the mode changes, the anchor values change. I’ll make a branch implementing this idea and we can compare

jarolrod (commented):
Strong Concept ACK of course, just one thing that may be worth considering

src/qml/controls/OnboardingInfo.qml
jarolrod: The gist of the idea is: 

This checking of the AppMode state can just be hidden inside of the Theme file, and the theme passes the values we want to the component

johnny9: I'm not sure if it makes sense to move anchor values outside of the component definition since they are relative values that only make sense in a context that includes all of the components involved.

One thing I also learned while doing this is that Qt doesn't recommend conditional anchors bindings. "Because the evaluation order of bindings is not defined, it is not recommended to change anchors via conditional bindings, as this can lead to the ordering issue described above. In the following example the Rectangle will eventually grow to the full width of its parent, because both left and right anchors will be simultaneously set during binding update." (https://doc.qt.io/qt-6/qtquick-positioning-anchors.html#changing-anchors). That was my primary reason for using states and AnchorChanges which unfortunately requires more boilerplate than just using a tertiary.

src/qml/controls/OnboardingInfo.qml
```diff
@@ -5,8 +5,9 @@
 import QtQuick 2.15
 import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.15
+import com.bitcoincore.qt 1.0
```
hebasto, line 8: It differs from `qmlRegisterSingletonInstance` call argument.

hebasto (commented):
Great to see such a PR that introduces mobile-specific code!

> This change registers a new QObject that can be used to check if the app is running in DESKTOP or MOBILE mode. 

At some point in the past we thought that desktop/mobile code will reside in separated directories. Using dedicated states for different platforms also looks ok.

Should we distinguish Android and iOS mobile devices?

johnny9:
Update from[ 8a3ee65 ](https://github.com/bitcoin-core/gui-qml/commit/8a3ee6581ece59e8fb23debd7c8370f8d9a70617)to 038caba:
  - Rebased with main
  - Fix a bad import from "com.bitcoincore.qt" to "org.bitcoincore.qt"

jarolrod (approved):
ACK 038caba43dbdda4b6e95a1da08272559c9623a9f

I have reviewed the code and agree that this can be merged. Also tested on both desktop and Android 12:

### Desktop
#### master

|    |    |
| -- | -- |
| Screen Shot 2022-10-03 at 11 37 42 PM | Screen Shot 2022-10-03 at 11 37 47 PM |

#### PR
|    |    |
| -- | -- |
| Screen Shot 2022-10-03 at 11 30 01 PM | Screen Shot 2022-10-03 at 11 30 14 PM |

### Mobile

#### master
|    |    |
| -- | -- |
|  |  |



#### PR
|    |    |
| -- | -- |
|  |  |

PR #172, opened 2022-09-17
qml: fix Bitcoin icon size to match spec

johnny9:
The included asset has padding around the icon so the width/height of the image component should be slightly larger to match the spec.

### DIFF
--- a/src/qml/pages/onboarding/onboarding01a.qml
+++ b/src/qml/pages/onboarding/onboarding01a.qml
@@ -32,8 +32,9 @@ Page {
             Layout.fillWidth: true
             Layout.alignment: Qt.AlignCenter
             source: "image://images/app"
-            sourceSize.width: 100
-            sourceSize.height: 100
+            // Bitcoin icon has ~11% padding
+            sourceSize.width: 112
+            sourceSize.height: 112
         }
         bannerMargin: 0
         bold: true


johnny9:
Nitpick but it threw me off when I was trying to understand how dpi scaling works. 

Before:
icon_size_before

After:
icon_size_pr

Spec:
icon-size-spec

jarolrod (commented):
Concept ACK, pinging designers for review: @GBKS @rabbitholiness @mouxdesign

We aren't going to change icon ([found here](https://github.com/bitcoin-core/gui-qml/blob/main/src/qt/res/icons/bitcoin.png)), so this would be preferable. 

Confirming that the icon isn't actually taking up 100px in term of actual content because of the padding that is apart of the image, and the PR addresses that:

| master | pr |
| ------- | -- |
| Screen Shot 2022-09-19 at 1 38 29 AM | Screen Shot 2022-09-19 at 1 33 48 AM |

jarolrod:
ACK 825451b9879a12774d12f13a1389e29b3568f4cc

This change gets us to fit the design spec.

@GBKS 
> The blue tint is done dynamically, right? We can provide a hex value and it will automatically adjust the color?

Yep the colorization is done internally, we don't have a different icon for the different networks.

> I would love it if we could replace the icon, so we can remove the gradient, drop-shadow, and extra spacing.

we'll do this at a later stage :)

PR #177, opened 2022-09-26
Update block tip in the gui thread

jarolrod:
This takes commit https://github.com/bitcoin-core/gui-qml/pull/38/commits/cf8a278e7a8bb5aaabef60b9345deff4c3d3c792 from https://github.com/bitcoin-core/gui-qml/pull/38/ and rebases it over changes on master; namely that we've dropped `GUIUtil::ObjectInvoke` in favor of `QMetaObject::invokeMethod`.

This avoids non-thread-safe errors. I ran into a segfault here while running some data intensive monkey-code related to blocks, this change prevented it. I'm not going to actually use the code that caused the segfault, just to illustrate that this is useful here anyways. 

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/177)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/177)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/177)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/177)

### DIFF
--- a/src/qml/nodemodel.cpp
+++ b/src/qml/nodemodel.cpp
@@ -10,6 +10,7 @@
 #include <cassert>
 #include <chrono>
 
+#include <QMetaObject>
 #include <QTimerEvent>
 
 NodeModel::NodeModel(interfaces::Node& node)
@@ -71,7 +72,9 @@ void NodeModel::ConnectToBlockTipSignal()
 
     m_handler_notify_block_tip = m_node.handleNotifyBlockTip(
         [this](SynchronizationState state, interfaces::BlockTip tip, double verification_progress) {
-            setBlockTipHeight(tip.block_height);
-            setVerificationProgress(verification_progress);
+            QMetaObject::invokeMethod(this, [=] {
+                setBlockTipHeight(tip.block_height);
+                setVerificationProgress(verification_progress);
+            });
         });
 }


johnny9:
Does the handler need to have disconnect() called when the NodeModel cleans up possibly? I'm having a hard time seeing which object in the lambda execution would be invalid and cause the segfault.

shaavan (commented):
The code change looks acceptable to me. The reasoning for which is nicely explained in [this](https://stackoverflow.com/a/13948634) Stackoverflow answer:

> ... The reason to use `QMetaObject::invokeMethod` if the recipient object might be in another thread is that attempting to call a slot directly on an object in another thread can lead to corruption or worse if it accesses or modifies non-thread-safe data.

However, I am curious about the seg fault error you mentioned in the description. Would you please add some points on what exactly you think caused the error?

johnny9:
I agree that this change is acceptable. It's ok to do simple setters on the gui thread.  My guess is that doing so forces the juicy part of the lambda to only execute while the NodeModel object is still alive but there is also the NodeImpl that adds its own wrapper to the lambda so it could be that object's lifetime as well.

hebasto (approved):
ACK eca52529dde163dcda4903f800060edc6c40919d, I have reviewed the code and it looks OK, I agree it can be merged.

promag:
> but there is also the NodeImpl that adds its own wrapper to the lambda so it could be that object's lifetime as well.

@johnny9 I don't understand this, can you rephrase it?

johnny9:
@promag https://github.com/bitcoin-core/gui-qml/blob/39846146fc196f47d63c857cb5307767bf6afc73/src/node/interfaces.cpp#L373 is the line of code I was referring to. It looks like the function takes the lambda as input (fn) and adds its own lambda around it before registering it to the handler so it can do "GuessVerificationProgress".

promag:
The segfault comes from not calling `setBlockTipHeight` and `setVerificationProgress` on the gui thread. There is no concurrency control, and (not sure) QML direct-connects to the signals `blockTipHeightChanged ` `verificationProgressChanged`.

The alternative to having some mutex is not ideal because we don't want to block the GUI and vice-versa

So the approach is to just defer the update.

Also, with `QMetaObject::invokeMethod(this, fn)`, Qt won't call `fn` if `this` is destroyed.

johnny9:
Does having NodeModel call m_handler_notify_block_tip.disconnect() in its deconstructor fix the segfault?

promag:
It's already called on the m_handler_notify_block_tip destructor. But that's not the issue, `setBlockTipHeight` and `setVerificationProgress` have to be called on the GUI thread.

johnny9:
@promag    I'm not sure Handler cleans up properly https://github.com/bitcoin-core/gui-qml/blob/8051cef61d4a24116f91ad4620f9ffeb75dfbfef/src/interfaces/handler.h#L25

PR #184, opened 2022-10-26
Introduce `OptionsQmlModel` class to bind node settings with GUI

hebasto:
This PR adds only a basic implementation with a support for the `-prune` setting only.

### DIFF
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -37,6 +37,7 @@ QT_FORMS_UI = \
 QT_MOC_CPP = \
   qml/moc_appmode.cpp \
   qml/moc_nodemodel.cpp \
+  qml/moc_options_model.cpp \
   qt/moc_addressbookpage.cpp \
   qt/moc_addresstablemodel.cpp \
   qt/moc_askpassphrasedialog.cpp \
@@ -112,6 +113,7 @@ BITCOIN_QT_H = \
   qml/bitcoin.h \
   qml/imageprovider.h \
   qml/nodemodel.h \
+  qml/options_model.h \
   qml/util.h \
   qt/addressbookpage.h \
   qt/addresstablemodel.h \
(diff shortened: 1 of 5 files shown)


hebasto:
cc @jarolrod

johnny9:
I see 
```
2022-10-27T18:02:45Z GUI: qrc:/qml/components/StorageSettings.qml:24: TypeError: Cannot read property 'pruneSizeGB' of null
2022-10-27T18:02:45Z GUI: qrc:/qml/components/StorageSettings.qml:16: TypeError: Cannot read property 'prune' of null
```
in debug.log

johnny9:
I believe you need to use qmlRegisterSingletonType and then import the registered module in order to access the object in the StorageSettings component. It seems only the root component has access to the object when using setContextProperty.

shaavan (commented):
Concept ACK

The implementation looks standard. But as @johnny9 suggested, we can experiment using the `qmlRegisterSingletonType`.

@hebasto 
I know you are currently occupied with finishing the CMake project, so I would like to share the burden of the work with you.
I want to work forward on this branch if that's okay with you. Thank you.

jarolrod:
@shaavan @johnny9 registering with the qml macro makes qml responsible for the lifecycle of the optionsmodel, we wouldn't want that. The way around this would be to subclass the qml engine, but there's no strong use-case for that right now.

johnny9:
Is that something unique to using the macro?  Is there not an equivalent of qmlRegisterSingletonInstance using the macros to allow you to manage the lifecycle yourself?

johnny9:
Ok I found what you were referring to here https://doc-snapshots.qt.io/qt5-5.15/qqmlengine.html#QML_SINGLETON.  Sounds like we should continue to use qmlRegisterSingletonType to override ownership of the object if we decide to use the macros.

jarolrod:
@johnny9 seems that you are correct in suggesting we use `qmlRegisterSingletonInstance` for many reasons including the performance hit of context properties and the non-constant lookup time. The ownership of the object can stay with c++ and not with the qml engine by specifying the `QQmlEngine::CppOwnership` flag.

PR #203, opened 2022-12-10
Introducing the node settings views

jarolrod:
This introduces the views that would allow for [settings configuration](https://www.figma.com/file/GaCoOSNHB2yMB9ThiDtred/Bitcoin-Core-App?node-id=4588%3A101966&t=m6V59cZ2STeCRp1y-0) after the user has finished onboarding and the node has started to run.

| NodeRuner | NodeSettings |
| --------- | ------------ |
| Screen Shot 2022-12-10 at 12 58 07 AM | Screen Shot 2022-12-10 at 12 58 12 AM |

For several reasons this also entangles the About and Developer options and entangles them into their own independent SwipeView, which means that we only need to instantiate SettingsAbout. There isn't a use case for the Developer options to be shown independently.

### Follow-ups:
- The caret-right icon used within the settings rows is too large because I'm just using NavButton, the width and height should be adjusted to match closer to what the design file [specifies](https://www.figma.com/file/GaCoOSNHB2yMB9ThiDtred/Bitcoin-Core-App?node-id=4588%3A101966&t=m6V59cZ2STeCRp1y-0)
- This uses a StackView to move through the settings pages but on linux, the animation between pages isn't how we'd want it to be; it's fine on macOS. The animation should match our SwipeViews. Note: I didn't use SwipeView because while you can set an index to go to, it flips through all previous pages to get there which is visually distracting.

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/203)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/203)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/203)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/203)

### DIFF
--- a/src/qml/pages/node/NodeRunner.qml
+++ b/src/qml/pages/node/NodeRunner.qml
@@ -1,0 +1,43 @@
+// Copyright (c) 2022 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import "../../controls"
+import "../../components"
+
+Page {
+    background: null
+    clip: true
+    property alias navRightDetail: navbar.rightDetail
+    header: NavigationBar {
+        id: navbar
+    }
+    ColumnLayout {
+        spacing: 0
+        anchors.fill: parent
+        ColumnLayout {
+            width: 600
+            spacing: 0
+            anchors.centerIn: parent
+            Component.onCompleted: nodeModel.startNodeInitializionThread();
+            Image {
+                Layout.alignment: Qt.AlignCenter
+                source: "image://images/app"
+                sourceSize.width: 64
+                sourceSize.height: 64
+            }
+            BlockCounter {
+                Layout.alignment: Qt.AlignCenter
+                blockHeight: nodeModel.blockTipHeight
+            }
+            ProgressIndicator {
+                width: 200
+                Layout.alignment: Qt.AlignCenter
+                progress: nodeModel.verificationProgress
+            }
+        }
+    }
+}
--- a/src/qml/pages/node/NodeSettings.qml
+++ b/src/qml/pages/node/NodeSettings.qml
@@ -1,0 +1,120 @@
+// Copyright (c) 2022 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import "../../controls"
+import "../../components"
+import "../settings"
+
+Item {
+    id: nodeSettings
+    property alias navMiddleDetail: nodeSettingsView.navMiddleDetail
+    property alias navRightDetail: nodeSettingsView.navRightDetail
+    StackView {
+        id: nodeSettingsView
+        property alias navMiddleDetail: node_settings.navMiddleDetail
+        property alias navRightDetail: node_settings.navRightDetail
+        initialItem: Page {
+            id: node_settings
+            property alias navMiddleDetail: navbar.middleDetail
+            property alias navRightDetail: navbar.rightDetail
+            background: null
+            header: NavigationBar {
+                id: navbar
+            }
+            ColumnLayout {
+                spacing: 0
+                width: parent.width
+                ColumnLayout {
+                    spacing: 20
+                    Layout.maximumWidth: 450
+                    Layout.topMargin: 30
+                    Layout.leftMargin: 20
+                    Layout.rightMargin: 20
+                    Layout.alignment: Qt.AlignCenter
+                    Setting {
+                        Layout.fillWidth: true
+                        header: qsTr("Dark Mode")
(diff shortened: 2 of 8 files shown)


hebasto:
Rebase needed.

jarolrod:
updated from e7763aa to 942a3b4

changes: rebased over master

src/qml/pages/node/NodeSettings.qml
```diff
@@ -0,0 +1,114 @@
...
+    id: nodeSettings
+    property alias navMiddleDetail: node_settings.navMiddleDetail
+    property alias navRightDetail: node_settings.navRightDetail
+    background: null
+    StackView {
+        id: nodeSettingsView
+        initialItem: node_settings
+        anchors.fill: parent
+    }
+    Page {
```
johnny9, line 22: This should be defined as a Component so it isn't loaded twice.

src/qml/pages/node/NodeSettings.qml
```diff
@@ -0,0 +1,114 @@
...
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import "../../controls"
+import "../../components"
+import "../settings"
+
+Page {
```
johnny9, line 12: Can just be an Item since header and footer are not set at this level.

src/qml/pages/node/NodeSettings.qml
```diff
@@ -0,0 +1,114 @@
...
+        property alias navMiddleDetail: navbar.middleDetail
+        property alias navRightDetail: navbar.rightDetail
+        background: null
+        header: NavigationBar {
+            id: navbar
+        }
+        ColumnLayout {
+            spacing: 0
+            width: parent.width
+            ColumnLayout {
```
johnny9, line 33: Seems like one of these ColumnLayouts can be removed

src/qml/pages/settings/SettingsAbout.qml
```diff
@@ -8,19 +8,43 @@ import QtQuick.Layouts 1.15
...
-    headerMargin: 0
-    description: qsTr("Bitcoin Core is an open source project.\nIf you find it useful, please contribute.\n\n This is experimental software.")
-    descriptionMargin: 20
-    detailActive: true
-    detailItem: ColumnLayout {
-        spacing: 0
-        AboutOptions {
-            Layout.maximumWidth: 450
-            Layout.alignment: Qt.AlignCenter
+Page {
```
johnny9, line 11: This can be Item since header/footer are not set.

johnny9 (changes requested):

jarolrod:
updated from 942a3b4 to 06235ca ([pr203.02](https://github.com/jarolrod/gui-qml/tree/pr203.02) -> [pr203.03](https://github.com/jarolrod/gui-qml/tree/pr203.03), [diff](https://github.com/jarolrod/gui-qml/compare/pr203.02..pr203.03))

changes: addressed @johnny9 suggestions -1

src/qml/pages/node/NodeSettings.qml
jarolrod: this allows for the table to expand but be a max width of 450, we use the same nested approach in the `InformationPage` control; open to better ways of accomplishing this

jarolrod:
updated from 942a3b4 to 06235ca ([pr203.03](https://github.com/jarolrod/gui-qml/tree/pr203.03) -> [pr203.04](https://github.com/jarolrod/gui-qml/tree/pr203.04), [diff](https://github.com/jarolrod/gui-qml/compare/pr203.03..pr203.04))

changes: rebased over changes on master

src/qml/pages/node/NodeSettings.qml
johnny9: Can use width: Math.min(parent.width, 450)

src/qml/pages/node/NodeRunner.qml
```diff
@@ -0,0 +1,43 @@
...
+    background: null
+    clip: true
+    property alias navRightDetail: navbar.rightDetail
+    header: NavigationBar {
+        id: navbar
+    }
+    ColumnLayout {
+        spacing: 0
+        anchors.fill: parent
+        ColumnLayout {
```
johnny9, line 21: Here's another double ColumnLayout.  I think we should reduce the layers here or use a different top level Item.

johnny9 (changes requested):

src/qml/pages/node/NodeRunner.qml
jarolrod: maybe we can leave it to a follow-up pr to address the nested columnlayouts throughout the code?

johnny9 (approved):
ACK 6037471dbbe88a4c34daefc0095a7d183af2b38d

Changes look good. Built and ran with not qml warnings/errors and views render as expected. Will address ColumnLayout comments in a follow up PR.

PR #207, opened 2022-12-14
Introduce OptionsModel backend, Wire up Storage Amount settings to backend

jarolrod:
This re-uses https://github.com/bitcoin-core/gui-qml/pull/184, with changes, to introduce an OptionsModel backend, while also wiring the Storage Amount settings; as the first settings to be wired to the options backend.

Both the "easy configuration" and "advanced configuration" methods for configuring the storage amount are wired to the backend. While on the fifth view of the onboarding wizard, the view with the header "Storage", if you click on the "Reduce Storage" button, your `Settings.json` file will show:

```
{
    "prune": 1907
}
```

and if you click on the "Store all data" button, your `Settings.json` file will show:

```
{
    "prune": 0
}
```

Within "Detailed Settings", you can examine that the prune target specified within the `Settings.json` file will only be non-zero if a non-zero value is set in storage limit and if the "Store recent blocks only" switch is enabled, as it should be.

The values you configure while onboarding will then be visible within the NodeSettings view when you click on "Storage".

**Note:** This doesn't complete the entire feature set around the setting, enabling, or disabling of this feature. There is still work to be done, this is the initial introduction and wiring.

The screenshots below highlight the functionality of this PR.
| configuring while onboarding | correct values available within the node settings window |
| ---------------------------- | -------------------------------------------------------- |
| Screen Shot 2022-12-14 at 4 32 48 AM |  Screen Shot 2022-12-14 at 4 32 57 AM |

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/207)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/207)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/207)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/207)

### DIFF
--- a/src/qml/bitcoin.cpp
+++ b/src/qml/bitcoin.cpp
@@ -14,6 +14,7 @@
 #include <qml/appmode.h>
 #include <qml/imageprovider.h>
 #include <qml/nodemodel.h>
+#include <qml/options_model.h>
 #include <qml/util.h>
 #include <qt/guiconstants.h>
 #include <qt/guiutil.h>
@@ -184,6 +185,9 @@ int QmlGuiMain(int argc, char* argv[])
     engine.addImageProvider(QStringLiteral("images"), new ImageProvider{network_style.data()});
 
     engine.rootContext()->setContextProperty("nodeModel", &node_model);
+
+    OptionsQmlModel options_model{*node};
+    engine.rootContext()->setContextProperty("optionsModel", &options_model);
 
 #ifdef __ANDROID__
     AppMode app_mode(AppMode::MOBILE);
--- a/src/qml/components/StorageOptions.qml
+++ b/src/qml/components/StorageOptions.qml
@@ -19,11 +19,18 @@ ColumnLayout {
         description: qsTr("Uses about 2GB. For simple wallet use.")
         recommended: true
         checked: true
+        onClicked: {
+            optionsModel.prune = true
+            optionsModel.pruneSizeGB = 2
+        }
     }
     OptionButton {
         Layout.fillWidth: true
         ButtonGroup.group: group
         text: qsTr("Store all data")
         description: qsTr("Uses about 550GB. Support the network.")
+        onClicked: {
+            optionsModel.prune = false
+        }
     }
 }
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -37,6 +37,7 @@ QT_FORMS_UI = \
 QT_MOC_CPP = \
   qml/moc_appmode.cpp \
   qml/moc_nodemodel.cpp \
+  qml/moc_options_model.cpp \
   qt/moc_addressbookpage.cpp \
   qt/moc_addresstablemodel.cpp \
   qt/moc_askpassphrasedialog.cpp \
@@ -112,6 +113,7 @@ BITCOIN_QT_H = \
   qml/bitcoin.h \
   qml/imageprovider.h \
   qml/nodemodel.h \
+  qml/options_model.h \
   qml/util.h \
   qt/addressbookpage.h \
   qt/addresstablemodel.h \
@@ -291,6 +293,7 @@ BITCOIN_QML_BASE_CPP = \
   qml/bitcoin.cpp \
   qml/imageprovider.cpp \
   qml/nodemodel.cpp \
+  qml/options_model.cpp \
   qml/util.cpp
 
 QML_RES_FONTS = \
--- a/src/qml/components/StorageSettings.qml
+++ b/src/qml/components/StorageSettings.qml
@@ -11,21 +11,18 @@ ColumnLayout {
     spacing: 20
     Setting {
         Layout.fillWidth: true
-        header: qsTr("Store Recent blocks only")
-        actionItem: OptionSwitch {}
-    }
-    Setting {
-        Layout.fillWidth: true
-        header: qsTr("Storage limit")
-        actionItem: ValueInput {
-            description: qsTr("2 GB")
+        header: qsTr("Store recent blocks only")
+        actionItem: OptionSwitch {
+            checked: optionsModel.prune
+            onToggled: optionsModel.prune = checked
         }
     }
     Setting {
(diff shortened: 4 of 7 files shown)


src/qml/bitcoin.cpp
```diff
@@ -185,6 +186,9 @@ int QmlGuiMain(int argc, char* argv[])
 
     engine.rootContext()->setContextProperty("nodeModel", &node_model);
 
+    OptionsQmlModel options_model{*node};
+    engine.rootContext()->setContextProperty("options", &options_model);
```
johnny9, line 190: Might want a more specfic variable name here since its registered to the root to prevent possible conflicts with other property names.  "optionsModel" maybe.  

Also, I can't get my head around how this is working now. When hebasto did the previous PR setting a rootContext like this wasn't accessible in any of the sub components but this one is working. All I can think of is that it has something to do with how the StackView in main.qml is implemented now.

hebasto:
Rebase?

src/qml/bitcoin.cpp
shaavan: In **debug.log**, I can observe a few **TypeErrors**.

```
% cat debug.log | grep GUI

2022-12-20T13:45:16Z GUI: qrc:/qml/components/StorageSettings.qml:24: TypeError: Cannot read property 'pruneSizeGB' of null
2022-12-20T13:45:16Z GUI: qrc:/qml/components/StorageSettings.qml:16: TypeError: Cannot read property 'prune' of null
```

The reason for this **TypeError** is that the `OptionsQmlModel` is initialized after the engine.
And since objects are destroyed in reverse order of how they are initialized, the `OptionsQmlModel` is getting destroyed before the QML engine.

The same **TypeErrors** existed in #184, as pointed out by @johnny9 [here](https://github.com/bitcoin-core/gui-qml/pull/184#issuecomment-1293891380).

So you have to initialize the `OptionsQmlModel` before the engine to solve this.

shaavan (commented):
There is one bug in the code due to which the **OptionsButton** and **Setting** are not correctly connected.

Steps to reproduce:
1. Start GUI.
2. Select the "no prune" option.
3. Complete onboarding.
4. Restart GUI.

Observation:

1. The OptionsButton has reset back to the "prune" option, but the prune option is deselected in the detailed setting.

**Recording:**


https://user-images.githubusercontent.com/85434418/208688168-14c74272-b542-489c-b6e5-bf3d7fc2c59a.mp4

jarolrod:
@shaavan technically that can be addressed, but the issue here really is that we just haven't introduced detection of whether or not it's the first time that the user is running the gui. If there is a settings.json in the data directory we're looking in we wouldn't be showing the onboarding process.

jarolrod:
updated from 12cbdc9 to d72ac4c

- changes: rebased over changes on master

src/qml/bitcoin.cpp
jarolrod: will address both of these soon

hebasto:
I've re-run CI.

src/qml/bitcoin.cpp
johnny9, line 190:
```suggestion
    engine.rootContext()->setContextProperty("optionsModel", &options_model);
```

src/qml/components/StorageOptions.qml
```diff
@@ -19,11 +19,18 @@ ColumnLayout {
         description: qsTr("Uses about 2GB. For simple wallet use.")
         recommended: true
         checked: true
+        onClicked: {
+            options.prune = true
```
johnny9, line 23:
```suggestion
            optionsModel.prune = true
```

src/qml/components/StorageOptions.qml
```diff
@@ -19,11 +19,18 @@ ColumnLayout {
...
+            options.pruneSizeGB = 2
```
johnny9, line 24:
```suggestion
            optionsModel.pruneSizeGB = 2
```

jarolrod:
updated from d72ac4c to cc96bd1, [compare](https://github.com/bitcoin-core/gui-qml/compare/d72ac4c5295795301c924c4a3cad54f634531ec2..cc96bd1695c9fc4058a5ecc91728030f804bb8cd)

Changes:
- addressed feedback on [naming](https://github.com/bitcoin-core/gui-qml/pull/207#discussion_r1070390019)
- Added logic to set default prune values when the SettingsStorage page finishes loading. This is needed because if the user does not click on the prune button then the prune value will not actually get set in settings.json, even though the prune box is checked.

hebasto:
Please rebase :)

jarolrod:
Updated from cc96bd1 to 63162b0

Changes: rebased over main

A note on implications of setting the default prune values when the page is loaded and in relation to https://github.com/bitcoin-core/gui-qml/issues/219:

If we have an approach where we are only loading one page at a time, that means if a user goes back or forwards, the page gets loaded again. This means that if the user had configured a prune target of 8gb as a custom setting, continued on to the next page then decided to go back and change the prune target, the prune target would be reset to 2gb. This can be mitigated with detection of settings already set, which is needed anyways.

Alternatively, the logic here can be changed into a form, where when the user finally finishes onboarding, then the settings get written. This is probably preferable anyways, for several cases such as quiting the application during onboarding; if we write settings, we wouldn't show onboarding again, but if we only write settings when onboarding is finished, the onboarding wizard will show on subsequent gui startups until the user completes onboarding.

johnny9 (approved):
ACK 63162b0f6a1a168f158804d17ad80907caa1f44e

Built and tested. Changes look good.

PR #214, opened 2023-01-02
Add focus outlines to controls used during onboarding

johnny9:
Changes to the controls within the Onboarding flows to show which control is in active focus. This will be the main indicator when navigating with a keyboard/tabbing.

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/214)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/214)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/214)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/214)

### DIFF
--- a/src/qml/controls/ExternalLink.qml
+++ b/src/qml/controls/ExternalLink.qml
@@ -85,4 +85,14 @@ AbstractButton {
         }
     }
     onClicked: Qt.openUrlExternally(link)
+
+    background: Rectangle {
+        visible: root.visualFocus
+        anchors.fill: parent
+        anchors.margins: -4
+        border.width: 2
+        border.color: Theme.color.purple
+        radius: 9
+        color: "transparent"
+    }
 }
--- a/src/qml/controls/ContinueButton.qml
+++ b/src/qml/controls/ContinueButton.qml
@@ -43,6 +43,16 @@ Button {
         Behavior on color {
             ColorAnimation { duration: 150 }
         }
+
+        Rectangle {
+            visible: root.visualFocus
+            anchors.fill: parent
+            anchors.margins: -4
+            border.width: 2
+            border.color: Theme.color.purple
+            radius: 9
+            color: "transparent"
+        }
     }
     MouseArea {
         anchors.fill: parent
--- a/src/qml/controls/NavButton.qml
+++ b/src/qml/controls/NavButton.qml
@@ -39,7 +39,18 @@ AbstractButton {
         Behavior on color {
             ColorAnimation { duration: 150 }
         }
+
+        Rectangle {
+            visible: root.visualFocus
+            anchors.fill: parent
+            anchors.margins: -4
+            border.width: 2
+            border.color: Theme.color.purple
+            radius: 9
+            color: "transparent"
+        }
     }
+
     contentItem: RowLayout {
         anchors.fill: parent
         spacing: 0
@@ -77,7 +88,7 @@ AbstractButton {
                    text: root.text
               }
           }
-        }
+       }
     }
     MouseArea {
         anchors.fill: parent
--- a/src/qml/controls/OptionButton.qml
+++ b/src/qml/controls/OptionButton.qml
@@ -23,10 +23,9 @@ Button {
             anchors.fill: parent
             anchors.margins: -4
             border.width: 2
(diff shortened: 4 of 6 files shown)


jarolrod:
concept ACK, going in the right direction here. 

Not a blocker for this PR but: seems we can cycle into components in pages that aren't currently in view, right now this is an issue for proper tab-focus functionality and for accessibility features like a screenreader.

johnny9:
|||||
|---|---|---|---|
|Screenshot from 2023-01-02 21-22-00|Screenshot from 2023-01-02 21-22-14|Screenshot from 2023-01-03 07-54-54|Screenshot from 2023-01-03 07-54-13|

johnny9:
> concept ACK, going in the right direction here.
> 
> Not a blocker for this PR but: seems we can cycle into components in pages that aren't currently in view, right now this is an issue for proper tab-focus functionality and for accessibility features like a screenreader.

Yeah, I've been working on the KeyNavigation to prevent the focus from going to outside pages.  It's more complicated than I thought it would be but it seems to be workable. Good motivation to get some simple qml testing going too since its something that could be broken fairly easily with changes to the controls.

src/qml/controls/ExternalLink.qml
```diff
@@ -46,4 +46,17 @@ Control {
...
         }
     }
+
+    background: Rectangle {
+        visible: root.visualFocus
+        anchors.fill: parent
+        anchors.leftMargin: -15
+        anchors.rightMargin: -4
+        anchors.topMargin: -4
+        anchors.bottomMargin: -5
```
jarolrod, line 56: Why have different values like here? The border should be nicely defined and uniform, it is the elements themselves that differ in how much space they take up, and we need to address those issues in follow-ups. 

I would keep the definitions in all of these to just specify `anchors.margins: -4` and address the issues in the elements themselves, instead of changing how the border works to accommodate the size issues.

It might be possible to extrapolate this out into our theme file, but we can leave that for much later.

jarolrod (changes requested):

src/qml/controls/ExternalLink.qml
johnny9: Ok, I like that approach. I'll try it out and see if it can be adjusted properly with margins/padding values on the elements.

jarolrod: I'm going to ACK this

this pr doesn't have to address the issues with the elements, only to keep the declaration of the margins of the border to only `anchors.margins: -4` for now, alternatively we can merge like this and address the issues in a follow-up and clean up the definition of the border.

hebasto:
Tested on Ubuntu 22.04. Having multiple warnings:
```
qrc:/qml/controls/ValueInput.qml:22:9: Unable to assign [undefined] to bool
```

hebasto:
@johnny9 

Mind rebasing?

And https://github.com/bitcoin-core/gui-qml/pull/214#issuecomment-1385082098 remains unaddressed...

johnny9:
Update from[ 6f00bc3 ](https://github.com/bitcoin-core/gui-qml/commit/6f00bc31f8d8c7fbbb7df131769a991161a11d89)to[ 85317d3](https://github.com/bitcoin-core/gui-qml/commit/85317d38f1e92d79a878dfba402f2b41c95eae91):
  - Rebase with main
  - Address Jarol's comment by using `margins: -4` for all components
  - Make ValueInput a Control so it can receive tab focus and fix the error seen by hebasto

jarolrod (commented):
The tab focus for Setting should be the setting row itself, we'd have to move out the separator rectangle from the setting control into the ColumnLayout containing the settings though

Screen Shot 2023-02-04 at 10 47 01 PM

jarolrod:
Replaced by https://github.com/bitcoin-core/gui-qml/pull/263

PR #140, opened 2022-08-01
qml: Make links, data location and numerical description value un-tra…

shaavan:
This is a follow-up to #124 and addresses [this](https://github.com/bitcoin-core/gui-qml/pull/124#discussion_r930629844) comment.

This PR removed the `qsTr()` tag for the description strings, which were links, disk locations, or numerical values, and hence wouldn't be translated from their source language.

PR #217, opened 2023-01-16
Make untranslatable values, untranslatable

jarolrod:
This picks up and replaces https://github.com/bitcoin-core/gui-qml/pull/140

Rebases and removes one more occurrence of a value qsTr() on an untranslatable value.

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/217)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/217)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/217)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/217)

### DIFF
--- a/src/qml/components/StorageSettings.qml
+++ b/src/qml/components/StorageSettings.qml
@@ -25,7 +25,7 @@ ColumnLayout {
         Layout.fillWidth: true
         header: qsTr("Data location")
         actionItem: ValueInput {
-            description: qsTr("c://.../data")
+            description: "c://.../data"
         }
     }
 }
--- a/src/qml/components/AboutOptions.qml
+++ b/src/qml/components/AboutOptions.qml
@@ -13,7 +13,7 @@ ColumnLayout {
         Layout.fillWidth: true
         header: qsTr("Website")
         actionItem: ExternalLink {
-            description: qsTr("bitcoincore.org")
+            description: "bitcoincore.org"
             link: "https://bitcoincore.org"
             iconSource: "image://images/caret-right"
         }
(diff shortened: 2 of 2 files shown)


src/qml/components/StorageSettings.qml
```diff
@@ -18,14 +18,14 @@ ColumnLayout {
         Layout.fillWidth: true
         header: qsTr("Storage limit")
         actionItem: ValueInput {
-            description: qsTr("2 GB")
+            description: "2 GB"
```
hebasto, line 21: Not sure about this.

hebasto (commented):
Concept ACK.

src/qml/components/StorageSettings.qml
johnny9: Yeah this should still be translated. The translator should be parameterized in this case to convert the GB but not the number.

jarolrod: ok reverted this line, Still not sure how this will actually end up. The component that ends up being the real number input for our options can either split up the value and the units or paramterize the value. We will see what plays best into the validation that we need to do on the input value.

jarolrod:
updated from 35fe19d to d528468 

Changes: Leave the prune target text as a qsTr()

src/qml/components/StorageSettings.qml
johnny9: Depending on how you implement it. You can use format strings and add args

text: qsTr("File %1 of %2").arg(counter).arg(total)
 https://doc.qt.io/qt-5.15/qtquick-internationalization.html

hebasto (approved):
ACK d528468f75d298aea9f94526dba574e4610b4acb

PR #235, opened 2023-01-31
Fix layout issue with checkmark in OptionButton

jarolrod:
This fixes a UI issue within the OptionButton, when the option button is smaller than 450px; the checkmark will cause the text to wrap to be smaller. The issue then is that they UI is toggling in a distracting manner between two different widths for text within the OptionButton.

The issue is illustrated below:
### Master
| select 1 | select 2 |
| -------- | -------- |
| Screen Shot 2023-01-30 at 8 10 13 PM | Screen Shot 2023-01-30 at 8 10 26 PM |

### PR
This PR fixes this by setting the box containing the Header to take up 90% of the internal width of the OptionButton and giving the remaining 10% to the Checkmark icon button. The math for this is as follows:
- given a width of 450, and padding of 15, we have an internal max width of 420 🔥
- The design file specifies that when the OptionButton has a width of 450, the box containing the header has a width of 380
- 380/420 = ~0.9
Screen Shot 2023-01-30 at 7 50 34 PM

| select 1 | select 2 |
| -------- | -------- |
| Screen Shot 2023-01-30 at 8 06 43 PM | Screen Shot 2023-01-30 at 8 06 55 PM |

### Why a 15pt Right Margin

This sets `Layout.rightMargin: 15` on the box containing the header because otherwise the header text would get too close to the checkmark

| without | with |
| ------- | ---- |
| Screen Shot 2023-01-30 at 8 04 54 PM | Screen Shot 2023-01-30 at 8 03 59 PM |

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/235)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/235)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/235)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/235)

### DIFF
--- a/src/qml/controls/OptionButton.qml
+++ b/src/qml/controls/OptionButton.qml
@@ -33,6 +33,7 @@ Button {
         spacing: 3
         ColumnLayout {
             spacing: 3
+            Layout.fillWidth: true
             Header {
                 Layout.fillWidth: true
                 Layout.preferredWidth: 0
@@ -64,11 +65,12 @@ Button {
                 }
             }
         }
-        Loader {
-            id: detail_loader
-            visible: button.checked
-            active: true
-            sourceComponent: Button {
+        Item {
+            height: parent.height
+            width: 40
+            Button {
+                anchors.centerIn: parent
+                visible: button.checked
                 icon.source: "image://images/check"
                 icon.color: Theme.color.neutral9
                 icon.height: 24


jarolrod:
cc @GBKS

GBKS:
Thanks for addressing this. Regarding the 15px margin being the right number... perfect numbers don't really exist in responsive design since we have so many screen sizes and content configurations (especially across languages). Maybe we could tighten it up to 5-10px... but what you have looks great, let's go for it.

Probably irrelevant here, but in CSS, I'd have the icon a fixed width and the text area fill the remaining horizontal space (so not a fixed percentage). And always render the icon, just with opacity 0 when it should not be visible to the user to avoid layout reflow. Flex box layouts make this super easy these days.

jarolrod:
Updated from 1f85ce4 to 259ca0a, [compare](https://github.com/bitcoin-core/gui-qml/compare/1f85ce4c50de85f8e0991414aab13913cd23eefc..259ca0a0b53ece7714e8621cd475ed3fb348c86b)

changes: incorporated @GBKS suggestion of setting the icon to constant width (40) then letting the text fill up the rest of the space

johnny9:
ACK 259ca0a0b53ece7714e8621cd475ed3fb348c86b

PR #256, opened 2023-02-11
ValueInput "EMPTY" state

jarolrod:
This adds an `"EMPTY"` state to `ValueInput`, as specified in the [design file](https://www.figma.com/file/ek8w3n3upbluw5UL2lGhRx/Bitcoin-Core-App-Design?node-id=6079%3A166395&t=zvbm4HW9rxQf4iBb-4).

This state is only added to `ValueInput` because the `Setting` control should not care about this. All other states continue to function as normal*.

### Light

| EMPTY | FILLED | HOVER | ACTIVE |
| ----- | ------ | ----- | ------ |
| Screen Shot 2023-02-10 at 5 26 04 PM | Screen Shot 2023-02-10 at 5 24 42 PM | Screen Shot 2023-02-10 at 5 24 57 PM | Screen Shot 2023-02-10 at 5 25 18 PM |

### Dark

| EMPTY | FILLED | HOVER | ACTIVE |
| ----- | ------ | ----- | ------ |
| Screen Shot 2023-02-10 at 5 23 43 PM | Screen Shot 2023-02-10 at 5 24 00 PM | Screen Shot 2023-02-10 at 5 24 11 PM | Screen Shot 2023-02-10 at 5 24 28 PM |

*the Setting control loses the active state when the mouse moves away from it, will be addressed in a follow-up.

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/256)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/256)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/256)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/256)

PR #255, opened 2023-02-10
Properly relinquish activeFocus on ValueInput when editing is finished

jarolrod:
On master, pressing `enter` while editing a setting that has a ValueInput does not properly relinquish focus from the ValueInput, still leaving it in an editing state. This fixes that by:

a. Making ValueInput a [TextInput](https://doc.qt.io/qt-5/qml-qtquick-textinput.html#text-prop) instead of a [TextEdit](https://doc.qt.io/qt-5/qml-qtquick-textedit.html)
  - TextEdit allows for multi-line input, so an `enter` in this would add a new line, which we don't want. There is no use case for a ValueInput to be multiple lines, it should only be one. Additionally, this prevents `enter` from signaling that editing is finished
  - TextInput is specifically for single line input. This QML type also allows us to do necessary validations with a [validator](https://doc.qt.io/qt-5/qml-qtquick-textinput.html#validator-prop), [inputMask](https://doc.qt.io/qt-5/qml-qtquick-textinput.html#inputMask-prop), and signal what kind of keyboard we want with [inputMethodHints](https://doc.qt.io/qt-5/qml-qtquick-textinput.html#inputMethodHints-prop). Additionally, this allows for `enter` to signal that editing is finished.

b. Relinquishing focus on the ValueInput itself, by forcing focus back to the parent Setting.

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/255)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/255)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/255)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/255)

PR #245, opened 2023-02-02
Custom Prune OptionButton

jarolrod:
This introduces a new OptionButton in StorageOptions that represents a custom prune target option. This is shown only when the user has gone into detailed settings and chosen a custom prune target.

Additionally, this moves the logic of setting the default prune value from SettingsStorage into StorageOptions. It is the easy storage option button shown during onboarding that should set this. With the original logic, when onboarding is not shown because the user has already gone through it, the app will set prune to 2 gb again.

| light | dark |
| ----- | ---- |
| Screen Shot 2023-02-01 at 10 30 36 PM | Screen Shot 2023-02-01 at 10 30 16 PM |

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/<PR>)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/<PR>)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/<PR>)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/<PR>)

PR #257, opened 2023-02-11
Display an error text when ValueInput has invalid user input

jarolrod:
based on https://github.com/bitcoin-core/gui-qml/pull/255 and https://github.com/bitcoin-core/gui-qml/pull/256

This allows for a ValueInput to signal when it currently has an invalid input from the user and display an error text to the user. IntValidator values are essentially arbitrary right now until we get some more intelligent detection of system resources.

| invalid input | valid input |
| ------------- | ----------- |
| Screen Shot 2023-02-10 at 7 08 22 PM | Screen Shot 2023-02-10 at 5 24 42 PM |

Draft until we resolve the following concerns @GBKS @johnny9 @hebasto :
- The TextInput field can be left in an empty state if the user deletes all of the text and then clicks on another TextInput field. This doesn't write any value, but the TextInput should not be left empty
  - We could fill in the current value set in settings.json for the setting if the user leaves the TextInput field in an empty set, the stored value is what the GUI is going to use.
- Should the error text be shown in a new row, so that the TextInput field doesn't get moved around when showing/hiding the error text?

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/<PR>)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/<PR>)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/<PR>)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/<PR>)


GBKS:
We talked about this a bit on Slack, so just repeating my thoughts (IIRC). If we evaluate the validity of the input on each key stroke, it's pretty annoying if the layout changes. So it's better to evaluate it when the input field loses focus, and in that case, the layout reflow should be OK.

When entering nothing, a reset to the current value also makes sense.

A more helpful error text would be "Enter a value between X and Y".

Is it possible to use MB instead of MiB? I'd assume that more people know what MB are and might confuse MiB with the Men in Black.

hebasto:
> based on #255

It has just been merged. Rebase?

jarolrod:
Updated from[ 1e89928 ](https://github.com/bitcoin-core/gui-qml/commit/1e8992805459b275c5565306c8235b8cbcbf550d)to[ 00fac08](https://github.com/bitcoin-core/gui-qml/commit/00fac088265af0e0c94499e619de44f32a24d1fe), [compare](https://github.com/bitcoin-core/gui-qml/compare/1e8992805459b275c5565306c8235b8cbcbf550d..00fac088265af0e0c94499e619de44f32a24d1fe)

Changes: rebased to resolve conlicts

jarolrod:
Updated from[ 00fac08 ](https://github.com/bitcoin-core/gui-qml/commit/00fac088265af0e0c94499e619de44f32a24d1fe)to [8bfafde](https://github.com/bitcoin-core/gui-qml/commit/8bfafdecdcf51430b3745daad1c1bb8adab465bb), [compare](https://github.com/bitcoin-core/gui-qml/compare/00fac088265af0e0c94499e619de44f32a24d1fe..8bfafdecdcf51430b3745daad1c1bb8adab465bb)

Changes: Addressed Review [feedback](https://github.com/bitcoin-core/gui-qml/pull/257#issuecomment-1427506593)
- Validation happens on editing finished instead of every input keystroke
- Display informative error text with what the valid values are


TODO in follow-ups:
- Restrict the characters that can be input into the field, best option is [inputMask](https://doc.qt.io/qt-6/qml-qtquick-textinput.html#inputMask-prop)
- Restrict the width of the error text
- Address review feedback that the default value or the value that options model has for a setting should be set if the user leaves a textinput in an empty state

hebasto:
> based on https://github.com/bitcoin-core/gui-qml/pull/271

https://github.com/bitcoin-core/gui-qml/pull/271 has just been merged.

jarolrod:
Updated from [8bfafde ](https://github.com/bitcoin-core/gui-qml/commit/8bfafdecdcf51430b3745daad1c1bb8adab465bb)to [d8cf8d0](https://github.com/bitcoin-core/gui-qml/commit/d8cf8d018a9dcb7efa541df3e4b421649b2644b5), [compare](https://github.com/bitcoin-core/gui-qml/compare/8bfafdecdcf51430b3745daad1c1bb8adab465bb..d8cf8d018a9dcb7efa541df3e4b421649b2644b5)

Changes: rebased over main

GBKS:
Just tested it and I think the input validation needs a bit more work. While it worked well when I entered a low-ish number, I was able to instantly break the app by entering random text (or a really high number like 100 trillion).

image

hebasto:
> based on https://github.com/bitcoin-core/gui-qml/pull/256

It has been just merged.

jarolrod:
Updated from from [d8cf8d0 ](https://github.com/bitcoin-core/gui-qml/commit/d8cf8d018a9dcb7efa541df3e4b421649b2644b5)to [a51820f](https://github.com/bitcoin-core/gui-qml/commit/a51820f9e704b36aff0e67695b33201dd85310ea), [compare](https://github.com/bitcoin-core/gui-qml/compare/d8cf8d018a9dcb7efa541df3e4b421649b2644b5..a51820f9e704b36aff0e67695b33201dd85310ea)

changes: rebased over main

jarolrod:
Updated from [a51820f ](https://github.com/bitcoin-core/gui-qml/commit/a51820f9e704b36aff0e67695b33201dd85310ea)to [524decd](https://github.com/bitcoin-core/gui-qml/commit/524decd192d45192933c26b8e665a7565d78e6ab), [compare](https://github.com/bitcoin-core/gui-qml/compare/a51820f9e704b36aff0e67695b33201dd85310ea..524decd192d45192933c26b8e665a7565d78e6ab)

Changes: @GBKS your concerns should now be addressed here. I hadn't addressed those two points yet because I was thinking about what changes would be needed to ValueInput to allow for it to be used within the Proxy screen, but I'll leave that thinking to another PR.

- Set a maximum input length of 5 to the ValueInput control. This is enough to enter a valid value for dbcache, threads, and prune target
- Set an IntValidator without specifying the valid bottom or top values. What this does is it prevents any characters that are not numeric from being entered, while allowing a character like `-` to enter negative numbers (can be used in threads).

hebasto:
Merged as a part of #245.

Issue #297, opened 2023-03-22
File name of download


mouxdesign:
Hey team,

Just a quick suggestion would be to change the file name of the download:

Suggestion: From insecure to unsecure. 

https://github.com/bitcoin-core/gui-qml/tree/main/src/qml

hebasto:
Just leaving this here (from https://chat.openai.com):

> What is the difference between 2 words: "insecure" and "unsecure"?

> The words "insecure" and "unsecure" are often confused, but they have different meanings.
>
> "Insecure" means not confident or uncertain, and it is often used to describe a person's emotional state or a situation where there is a lack of security.
> 
> For example: "She felt insecure about her abilities to perform the task." or "The website's insecure password policy left the user's data vulnerable."
> 
> On the other hand, "unsecure" is not a commonly used word in English. If it is used, it means not firmly fixed in place or not secured properly.
> 
> For example: "The ladder is unsecure and could fall at any moment." or "The bolts on the door are unsecure, allowing easy access for intruders."
> 
> However, it's worth noting that "insecure" and "unsecure" are sometimes used interchangeably in certain contexts, particularly in reference to computer systems and networks. In this context, both words can be used to describe a lack of security or vulnerability to unauthorized access.

:smile:

PR #301, opened 2023-03-25
doc: Refer to CI assets as unsecure, instead of insecure

jarolrod:
Fixes https://github.com/bitcoin-core/gui-qml/issues/297


hebasto (approved):
ACK 12abfb0b0dc4ee753be0b57ad4ff4319f10bd710

PR #303, opened 2023-03-25
doc: document qttools5-dev-tools qml build requirement

jarolrod:
This was needed on a fresh Ubuntu 22.04 installation running on ARM64. The package provides `lrelease`. Not needed on fedora.


jarolrod:
This is already documented in build-unix :D

PR #343, opened 2023-06-03
Ask before opening external links

jarolrod:
Introduces an external link confirm popup window that prompts the user to acknowledge that they really want to visit the external link. This is based off the [design file](https://www.figma.com/file/ek8w3n3upbluw5UL2lGhRx/Bitcoin-Core-App-Design?type=design&node-id=6265-24513&t=PyRCN7AM9KdhuP8x-4)

This contains a menu bar with a title, but is missing the close button. It is also missing the "Don't ask again" button. These require some more conversation and can be addressed in a follow-up.

This deviates from the design file in providing a mobile layout that positions the buttons in a column layout instead of a rowlayout. cc @GBKS @mouxdesign

| Desktop Layout | Mobile Layout |
| -------------- | ------------- |
| Screen Shot 2023-06-03 at 2 11 30 AM | Screen Shot 2023-06-03 at 2 12 43 AM |

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/unsecure_win_gui.zip?branch=pull/343)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/unsecure_mac_gui.zip?branch=pull/343)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/unsecure_mac_arm64_gui.zip?branch=pull/343)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/unsecure_android_apk.zip?branch=pull/343)
[ARM32 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android32/unsecure_android_32bit_apk.zip?branch=pull/343)

PR #350, opened 2023-06-28
modified NavigationBar.qml so text is centered

D33r-Gee:
RE: PR [343](https://github.com/bitcoin-core/gui-qml/pull/343)

Modified the NavigationBar.qml file so that the "External Link" text would be centered in `ExternalPopup.qml` file


johnny9 (commented):
No need for comments.

jarolrod (commented):
Thanks for investigating this, the current approach would currently mess up the layout across different pages in the UI. Pictured below. I'll address this in https://github.com/bitcoin-core/gui-qml/pull/343

PR #355, opened 2023-08-23
Icon control

promag:
`Button` control is used to show icons because it has support to colorize the image. `Button` is packed with more stuff like states, event handling, label... eventually a light component should be used instead. For now, abstract the usage of `Button` in the new `Icon` control and use it in trivial places. There are more places where `Icon` can be used but require more changes, so I'm leaving them for follow-ups.

### DIFF
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -362,6 +362,7 @@ QML_RES_QML = \
   qml/controls/ExternalLink.qml \
   qml/controls/FocusBorder.qml \
   qml/controls/Header.qml \
+  qml/controls/Icon.qml \
   qml/controls/InformationPage.qml \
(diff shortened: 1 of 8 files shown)


promag:
Forgot to mention that this is a small step in fixing focus navigation since the previous `Button` instances had by default `enabled: true` and so were in the focus chain.

jarolrod (approved):
ACK 000837be04343cbe8b17f3c933a69c0ca78c14ab

Agree with encapsulating the Button Icon workaround into a dedicated component. This does make some progress in fixing the tab focus chain; can especially be seen in tabbing on a settings page.

johnny9 (approved):
ACK 000837be04343cbe8b17f3c933a69c0ca78c14ab

Code looks cleaner with the new component, good change.

PR #356, opened 2023-08-23
CaretRightIcon control

promag:
Replace `CaretRightButton` with a more appropriate `CaretRightIcon`, and refactor the `clicked` signed.

This also is a step towards removing the `loadedItem` property from `Setting`.


jarolrod (approved):
ACK 74d4a9740a39720fb001716c1f9b4ed7eaf8f3f2

PR #263, opened 2023-02-14
Separate out Separator from Setting control

jarolrod:
This is required in order to properly support the feature attempted by https://github.com/bitcoin-core/gui-qml/pull/214

### Master

| About | Dev | Prune | Connection | Node |
| ----- | --- | ----- | ---------- | ---- |
| about-master |dev-master |prune-master |connection-master |node-master |

### PR 

| About | Dev | Prune | Connection | Node |
| ----- | --- | ----- | ---------- | ---- |
| about-pr |dev-pr |prune-pr |connection-pr |node-pr |

This also gets us in line with the design file in regards to the height of a setting
| master | pr | design |
| ------ | -- | ------ |
| Screen Shot 2023-02-13 at 7 30 58 PM | Screen Shot 2023-02-13 at 7 21 49 PM | Screen Shot 2023-02-13 at 7 31 54 PM |

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/263)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/263)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/263)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/263)

PR #357, opened 2023-08-25
qml: drop unused last property from Setting

promag:
Remove unused property `last` from `Setting` component.


jarolrod (approved):
ACK c0e45bcf05f4ee1ee5a409174d522fd001b02976

This was overlooked in https://github.com/bitcoin-core/gui-qml/pull/263

PR #368, opened 2023-10-25
ci, gha: Build Android artifacts

hebasto:
This PR reintroduces downloadable binary artifacts for macOS and Windows as it was on now decommissioned Cirrus CI.

Unfortunately, the implementation is suboptimal due to various reasons. For example,
- A container is used as building the `qt` package in depends fails using the default Android installation in Ubuntu [image](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md).
- Caching, both for depends and ccache, does not work as expected within a container.

---

The artifacts are available for downloading from https://github.com/bitcoin-core/gui-qml/actions/workflows/artifacts.yml.

Also refer to GitHub documentation: https://docs.github.com/en/actions/managing-workflow-runs/downloading-workflow-artifacts.

Links for build artifacts:

[Android ARM 32-bit APK](https://github.com/bitcoin-core/gui-qml/suites/17611385186/artifacts/1006627194)
[Android ARM 64-bit APK](https://github.com/bitcoin-core/gui-qml/suites/17611385186/artifacts/1006627197)
[Android x86-64 APK](https://github.com/bitcoin-core/gui-qml/suites/17611385186/artifacts/1006627201)
[macOS Apple Silicon](https://github.com/bitcoin-core/gui-qml/suites/17611385186/artifacts/1006627204)
[macOS Intel](https://github.com/bitcoin-core/gui-qml/suites/17611385186/artifacts/1006627206)
[Windows](https://github.com/bitcoin-core/gui-qml/suites/17611385186/artifacts/1006627209)

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -122,3 +122,68 @@ jobs:
         with:
           name: ${{ matrix.host.artifact }}
           path: src/qt/${{ matrix.host.gui_exe }}
+
+  android:
+    name: ${{ matrix.host.name }}
+    runs-on: ubuntu-latest
+    container: ubuntu:22.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        host:
+          - name: 'Android ARM 64-bit APK'
+            triplet: 'aarch64-linux-android'
+            artifact: 'unsecure_android_arm64'
+          - name: 'Android ARM 32-bit APK'
+            triplet: 'armv7a-linux-android'
+            artifact: 'unsecure_android_arm32'
+          - name: 'Android x86_64 APK'
+            triplet: 'x86_64-linux-android'
+            artifact: 'unsecure_android_x86_64'
+
+    env:
+      ANDROID_HOME: '/tmp/Android'
+      ANDROID_API_LEVEL: '28'
+      ANDROID_BUILD_TOOLS_VERSION: '28.0.3'
+      ANDROID_NDK_VERSION: '23.2.8568313'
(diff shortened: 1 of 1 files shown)


.github/workflows/artifacts.yml
```diff
@@ -122,3 +122,68 @@ jobs:
...
+          make -j$(nproc) -C src/qt bitcoin-qt
+
+      - name: Build APK
+        run: |
+          make -j$(nproc) -C src/qt apk
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: ${{ matrix.host.artifact }}
+          path: src/qt/android/build/outputs/apk/debug/android-debug.apk
```
jarolrod, line 189: Do we want to call this debug? config isn't being built as debug here. 

Is it possible to give each android build a unique name? like android-arm64.apk, etc...

jarolrod (commented):
Concept ACK, want to also edit the pr template here?

.github/workflows/artifacts.yml
hebasto: > Do we want to call this debug?

It is named by the build system.

>  config isn't being built as debug here.

The other artifact is unsigned, btw.

> Is it possible to give each android build a unique name? like android-arm64.apk, etc...

Artifacts already have unique names. Do you think a name collision here is a real concern for users of those artifacts?

hebasto:
> want to also edit the pr template here?

Maybe someone else could adjust all related docs, including the pr template?

jarolrod (approved):
ACK 8d3533a7d2702207f86472f6d058270c1868eb08

Improvements to this and documentation mixups can be done in follow-ups. Tested that the arm64 android binary works.

PR #374, opened 2023-11-30
Don't call update during paint

promag:


jarolrod (approved):
ACK caa0f4a2d04a23c9cfef337b46b4db9c6e9483de

Confirmed this line is unnecessary

PR #377, opened 2023-12-01
Update build-unix.md to ensure clang is installed when changing configure flag for compilation

jpgaviria2:
make sure clang is installed otherwise ./configure command will fail

Added a step to make sure clang is installed on the linux machine before using the ./configure command to use clang instead of gcc, clang is not listed as the dependencies needed for linux.

suggestion for improvement only as I spent some time troubleshooting and didn't realise this was the error, was trying to follow instructions to build android and the make apk would fail due to not having a clang.

### DIFF
--- a/doc/build-unix.md
+++ b/doc/build-unix.md
@@ -33,7 +33,9 @@ Alternatively, or in addition, debugging information can be skipped for compilat
     ./configure CXXFLAGS="-O2"
 
 Finally, clang (often less resource hungry) can be used instead of gcc, which is used by default:
+Make sure clang is installed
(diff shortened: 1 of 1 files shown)


jarolrod (commented):
@jpgaviria2 Thanks for looking into contributing here. 

This isn't related to what we're building here. Additionally, this is an optional step in building core in general, so if you opt into this optional step using a different compiler than gcc, you should have installed this alternative compiler.

I'll close this here, but if you feel strongly that this is still of value, then I would suggest you open it on the main repo.

Issue #382, opened 2023-12-07
Set minimum block clock size to 200px width


GBKS:
### Issues, reports or feature requests related to the GUI should be opened directly on the GUI repo

- [X] I still think this issue should be opened here

### Report

On my Android phone, the clock is too small. Minimum size should be 200px width. It is ~131px wide.

![image](https://github.com/bitcoin-core/gui-qml/assets/695901/6aa973ec-91f2-4f38-abba-3f274cfa56d8)

D33r-Gee:
Hi @GBKS , what android device are you using?

GBKS:
It's a Galaxy A32 5G.

D33r-Gee:
> It's a Galaxy A32 5G.

Great I will test with a virtual device...

Also to make sure I'm able to reproduce it, which build are you using? Latest on main branch?

GBKS:
The math that defines the width is located [here](https://github.com/bitcoin-core/gui-qml/blob/67150efc17a17820193fbd5715e7fc92d417556a/src/qml/components/BlockClock.qml#L45).

`Math.min((root.parentWidth * dial.scale), (root.parentHeight * dial.scale))`

The Galaxy A32 has a logical resolution of 360x800. The dial scale is either 1/2 or 1/3. `parentWidth` is set [here](https://github.com/bitcoin-core/gui-qml/blob/67150efc17a17820193fbd5715e7fc92d417556a/src/qml/pages/node/NodeRunner.qml#L28) as the `parent.width - 40`. So that puts the width at `(360-40)/2 = 160` or `(360-40)/3 = ~107`. Not exactly the ~131 I roughly measured in my screenshot above, but both values are below the 200 minimum that we want. So maybe we change the line above to this?

`Math.max(200, Math.min((root.parentWidth * dial.scale), (root.parentHeight * dial.scale)))`

But that misses the scenario where the clock is shown in a widget. So you really don't want to go smaller than the screen width minus some padding. So maybe this?

`Math.max(Math.min(200, root.parentWidth - 30), Math.min((root.parentWidth * dial.scale), (root.parentHeight * dial.scale)))`

I am probably missing something, since my math above doesn't totally add up with my screenshot measurement.

Thanks for looking into this. I believe I am on the latest main branch.

D33r-Gee:
@GBKS, thanks for your comment above, made my life very easy :)

build and tested with the change you suggested:

[...]

PR #385, opened 2023-12-20
Fix For BlockClock Min Size

D33r-Gee:
Issue and Fix:

This fixes https://github.com/bitcoin-core/gui-qml/issues/382

[GBKS](https://github.com/GBKS) reported an issue where the blockclock on their Galaxy A32 5G phone displayed incorrectly under the min size of 200 pixels.
They provided a fix implemented in this pull request:

`Math.max(Math.min(200, root.parentWidth - 30), Math.min((root.parentWidth * dial.scale), (root.parentHeight * dial.scale)))`

Verification:

I tested the fix on both a virtual device Galaxy A32 5G in Android Studio and a physical Galaxy A13 5G (armv7) device. The fix successfully resolves the pixel-related display issue on both devices.

### DIFF
--- a/src/qml/components/BlockClock.qml
+++ b/src/qml/components/BlockClock.qml
@@ -42,7 +42,8 @@ Item {
         id: dial
         anchors.horizontalCenter: root.horizontalCenter
         scale: Theme.blockclocksize
-        width: Math.min((root.parentWidth * dial.scale), (root.parentHeight * dial.scale))
+        width: {Math.max(Math.min(200, Math.min(root.parentWidth - 30, root.parentHeight - 30)), 
+                Math.min((root.parentWidth * dial.scale), (root.parentHeight * dial.scale)))}
         height: dial.width
         penWidth: dial.width / 50
         timeRatioList: chainModel.timeRatioList


johnny9 (changes requested):
Remove the commented out line of code

johnny9:
Tested the qml logic on desktop by removing the minimum window restraints
Before:
Screenshot from 2023-12-20 23-49-00
After
Screenshot from 2023-12-20 23-48-43

Looks like it fixes the issue.

GBKS:
Awesome. I realized there's one thing missing from the code I proposed. I only ensures that the block fits the horizontal screen width, but not the vertical height. So here's the logic.

    // The maximum size.
    // This is our hardest constraint - the physical screen size.
    screenConstraintSize = Math.min(root.parentWidth, root.parentHeight) - 30

    // The minimum size (for legibility).
    minimumSize = 200

    // The dynamic target size based on our display mode scale.
    // Basically just scaling up for larger non-mobile screens.
    scaleTargetSize = Math.min(root.parentWidth, root.parentHeight) * dial.scale

    // Mashing them all up
    result = Math.min(screenConstraintSize, Math.max(minimumSize, scaleTargetSize))

    // Or in one line
    result = Math.min(Math.min(root.parentWidth, root.parentHeight) - 30, Math.max(200, Math.min(root.parentWidth, root.parentHeight) * dial.scale))

@jarolrod you probably spent the most time on this, how does that logic look to you?

johnny9:
Good catch with the missing minimum height factor

| Current PR | Updated Logic | Main |
| ---------- | ------------- | ---- |
| current pr extreme | updated pr extreme | main extreme |

Updated code can be
```
        width: {
            minimumSize = Math.min(200, Math.min(root.parentWidth - 30, root.parentHeight - 30)
            scaledSize = Math.min((root.parentWidth * dial.scale), (root.parentHeight * dial.scale))
            return Math.max(minimumSize, scaledSize)
        }
```

or

```
        width: {
            Math.max(Math.min(200, Math.min(root.parentWidth - 30, root.parentHeight - 30)),
                     Math.min((root.parentWidth * dial.scale), (root.parentHeight * dial.scale)))
        }
```

GBKS:
Phew, what are those screenshots from? Android split-screen? That gets really tiny with a huge signet indicator. We might even need something custom for that, like this design:

image

Would that be possible?

D33r-Gee:
> Remove the commented out line of code

Done in new commit

D33r-Gee:
Thanks @GBKS and @johnny9 for looking into the math.

I have update the PR with a new commit https://github.com/bitcoin-core/gui-qml/pull/385/commits/3f455ee10807a0663b833afd94f864a7b7c23c75 which replaces the line suggested by @GBKS with @johnny9's:

> Math.max(Math.min(200, Math.min(root.parentWidth - 30, root.parentHeight - 30)),
                     Math.min((root.parentWidth * dial.scale), (root.parentHeight * dial.scale)))

D33r-Gee:
Also here are some screenshots of the new commit running on different platforms/devices:

Ubuntu 22.04 showcase mode:

Ubuntu2204_showcase

Ubuntu 22.04 compact mode:

Ubuntu2204_compact

D33r-Gee:
Galaxy A13 5G physical device screenshots:

Portrait:

physical_GalaxyA13_portrait

Landscape:

physical_GalaxyA13_landscape

D33r-Gee:
Virtual device Galaxy A32 5G:

Landscape:

virtual_GalaxyA32_landscape

Portrait:

virtual_GalaxyA32_portrait

D33r-Gee:
Changing the mode from `compact` to `showcase` didn't have a noticeable effect on the Android builds.

src/qml/components/BlockClock.qml
```diff
@@ -42,7 +42,7 @@ Item {
         id: dial
         anchors.horizontalCenter: root.horizontalCenter
         scale: Theme.blockclocksize
-        width: Math.min((root.parentWidth * dial.scale), (root.parentHeight * dial.scale))
+        width: Math.max(Math.min(200, Math.min(root.parentWidth - 30, root.parentHeight - 30)), Math.min((root.parentWidth * dial.scale), (root.parentHeight * dial.scale)))
```
johnny9, line 45: Can we break this up into two lines? It's difficult to read.

johnny9 (changes requested):
One last thing. Give a better commit message and include the context in your commit header. Something like "qml: set minimum size for the BlockClock".

src/qml/components/BlockClock.qml
D33r-Gee: Yes on both counts... new commit with changes https://github.com/bitcoin-core/gui-qml/pull/385/commits/d6331aa60f46ceb8d6d5c2019b7d64d6ec16310a

D33r-Gee:
@GBKS regarding your question about the `signet` indicator:

> Would that be possible?

Yes, would you like to explore this in this PR or should we create a new one?

johnny9 (approved):
ACK d6331aa60f46ceb8d6d5c2019b7d64d6ec16310a

src/qml/components/BlockClock.qml
```diff
@@ -42,7 +42,8 @@ Item {
         id: dial
         anchors.horizontalCenter: root.horizontalCenter
         scale: Theme.blockclocksize
-        width: Math.min((root.parentWidth * dial.scale), (root.parentHeight * dial.scale))
+        width: {Math.max(Math.min(200, Math.min(root.parentWidth - 30, root.parentHeight - 30)), 
+                Math.min((root.parentWidth * dial.scale), (root.parentHeight * dial.scale)))}
```
pablomartin4btc, line 46: nit: I would extract the formula to a function where I'd structure the logic as @GBKS suggested [above](https://github.com/bitcoin-core/gui-qml/pull/385#issuecomment-1865777711).

<details>
<summary>How about this...</summary>**

```diff --git a/src/qml/components/BlockClock.qml b/src/qml/components/BlockClock.qml
--- a/src/qml/components/BlockClock.qml
+++ b/src/qml/components/BlockClock.qml
@@ -44,7 +44,7 @@ Item {
         id: dial
         anchors.horizontalCenter: root.horizontalCenter
         scale: Theme.blockclocksize
-        width: Math.min((root.parentWidth * dial.scale), (root.parentHeight * dial.scale))
+        width: calculateConstrainedWidth()
         height: dial.width
         penWidth: dial.width / 50
         timeRatioList: chainModel.timeRatioList
@@ -239,6 +239,24 @@ Item {
         }
     ]

+    function calculateConstrainedWidth() {
+        // Margin size used to create a buffer around the constrained area.
+        var marginSize = 30;
+
+        // The maximum size.
+        // This is our hardest constraint - the physical screen size.
+        var screenConstraintSize = Math.min(root.parentWidth, root.parentHeight) - marginSize;
+
+        // The minimum size (for legibility).
+        var minimumSize = 200;
+
+        // The dynamic target size based on our display mode scale.
+        // Basically just scaling up for larger non-mobile screens.
+        var scaleTargetSize = Math.min(root.parentWidth, root.parentHeight) * dial.scale;
+
+        // Mashing them all up
+        return Math.min(screenConstraintSize, Math.max(minimumSize, scaleTargetSize))
+    }

     function formatProgressPercentage(progress) {
         if (progress >= 1) {
```
</details>

pablomartin4btc (commented):
tACK d6331aa60f46ceb8d6d5c2019b7d64d6ec16310a

Tested on an Android Wear smart watch and experienced a significant improvement.

|           | Light Theme | Dark Theme |
| --------- | ----------- | ---------- |
| **Before**| Screenshot_20231227_223355 | Screenshot_20231227_223546 |
| **After** | Screenshot_20231228_110333 | Screenshot_20231228_105503
 
As it can be noticed, the network indicator has almost disappeared in such devices, therefor as @GBKS [pointed out](https://github.com/bitcoin-core/gui-qml/pull/385#issuecomment-1866339697) perhaps the network indicator could be moved up to the top-left, but that's where the "back" button is when we navigate through the settings, maybe it can be accommodated a bit up at the bottom reducing its size?

Another observation, out of scope of this PR, the settings button it's almost at the rounded border and it has been cut a bit, but I'm not sure if we support any/ all devices.

D33r-Gee:
Thanks @GBKS , @johnny9 , and @pablomartin4btc for testing and working on this PR

Thanks @hebasto for merging it!

src/qml/components/BlockClock.qml
D33r-Gee: thanks for testing on a wearable device,  maybe let’s open a new PR if this feels like a priority, what do you think @GBKS?

GBKS: I mocked things up and added them to the design docs, [see here](https://bitcoincore.app/block-clock/#layout-logic). Feedback appreciated.

GBKS: Also had a question about how this works on Android Wear. Is the full app running on the watch? Or does it run in some sort of stripped down way? I'd be surprised of watches are powerful enough. I also assume we can detect if the app runs on a watch and potentially adjust the layout accordingly (specifically for the rounded frame)?

pablomartin4btc: > Is the full app running on the watch? Or does it run in some sort of stripped down way? 

I haven't passed the "Connecting..." phase, even the watch is connected to the internet and everything, I'd need to debug and check deeper, but I doubt it, I was just going a bit further with the testing of this PR and since I was going thru the [Android guide](https://github.com/bitcoin-core/gui-qml/pull/366) it seems I got a bit overexcited :laughing:.

> I'd be surprised of watches are powerful enough. 

Well, this is an old one, 2017-ish, nowadays resources on smartwatches are still limited but they are getting there.

> I also assume we can detect if the app runs on a watch and potentially adjust the layout accordingly (specifically for the rounded frame)?

I'd think so, I'll check when I do something related with it.

PR #375, opened 2023-11-30
Improve peer delegate

promag:
Use Qt `ItemDelegate` and use properties `down` and `hovered`, making it possible to drop the `MouseArea` Also improve the layout and elide the label that can overflow.

fixes https://github.com/bitcoin-core/gui-qml/issues/372.

PR #387, opened 2024-01-29
qml: Add PeerDetails page

johnny9:
This commit introduces a new page, PeerDetails, that shows additional information on a Peer after it is selected from the list of Peers. After clicking on the peer, the new page gets pushed onto the settings stack. Two conditions will cause the view to pop, the user clicking the Back at the top left or the peer disconnecting.

For reviewers, I would very much like comments on the conversion of the Stats struct to the QML compatible object that is used.

This commit does not include the disconnect/ban buttons, the top right buttons to cycle through the peers, or the icons that are adjacent to the values.

Link to github actions build artifacts.

[Build Artifacts]()

### DIFF
--- a/src/qml/bitcoin.cpp
+++ b/src/qml/bitcoin.cpp
@@ -24,6 +24,7 @@
 #include <qml/models/networktraffictower.h>
 #include <qml/models/nodemodel.h>
 #include <qml/models/options_model.h>
+#include <qml/models/peerdetailsmodel.h>
 #include <qml/models/peerlistsortproxy.h>
 #include <qml/imageprovider.h>
 #include <qml/util.h>
@@ -292,6 +293,8 @@ int QmlGuiMain(int argc, char* argv[])
     qmlRegisterSingletonInstance<AppMode>("org.bitcoincore.qt", 1, 0, "AppMode", &app_mode);
     qmlRegisterType<BlockClockDial>("org.bitcoincore.qt", 1, 0, "BlockClockDial");
     qmlRegisterType<LineGraph>("org.bitcoincore.qt", 1, 0, "LineGraph");
+    qmlRegisterUncreatableType<PeerDetailsModel>("org.bitcoincore.qt", 1, 0, "PeerDetailsModel", "");
+
 
     engine.load(QUrl(QStringLiteral("qrc:///qml/pages/main.qml")));
     if (engine.rootObjects().isEmpty()) {
--- a/src/qml/pages/node/PeerDetails.qml
+++ b/src/qml/pages/node/PeerDetails.qml
@@ -1,0 +1,137 @@
+// Copyright (c) 2024 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import org.bitcoincore.qt 1.0
+import "../../controls"
+import "../../components"
+
+Page {
+    signal backClicked()
+
+    property PeerDetailsModel details
+
+    Connections {
+        target: details
+        function onDisconnected() {
+            root.backClicked()
+        }
+    }
+
+    id: root
+    background: null
+
+    header: NavigationBar2 {
+        leftItem: NavButton {
+            iconSource: "image://images/caret-left"
+            text: qsTr("Back")
+            onClicked: root.backClicked()
+        }
+        centerItem: Header {
+            headerBold: true
+            headerSize: 18
+            header: qsTr("Peer " + details.nodeId)
+        }
+    }
+
+    component PeerKeyValueRow: Row {
+        width: parent.width
+        property string key: ""
+        property string value: ""
+        height: 21 * valueText.lineCount
+        spacing: 10
+        CoreText {
+            color: Theme.color.neutral9;
+            text: key;
+            width: 125;
+            horizontalAlignment: Qt.AlignLeft;
+        }
+        CoreText {
+            id: valueText
+            color: Theme.color.neutral9;
+            elide: Text.ElideRight;
+            wrapMode: Text.WordWrap;
+            text: value
+            width: parent.width - 125;
+            horizontalAlignment: Qt.AlignLeft;
+        }
+    }
+
+    ScrollView {
+        id: scrollView
+        width: parent.width
+        height: parent.height
+        clip: true
+        contentWidth: width
+
+        Column {
+            width: Math.min(parent.width - 40, 450)
+            anchors.horizontalCenter: parent.horizontalCenter
+            spacing: 10
+            topPadding: 30
+            bottomPadding: 30
+
+            CoreText {
+                text: "Information";
+                bold: true;
+                font.pixelSize: 18;
+                horizontalAlignment: Qt.AlignLeft;
+                color: Theme.color.neutral9;
+            }
+            Column {
+                width: parent.width
+                bottomPadding: 5
+                PeerKeyValueRow { key: qsTr("Address"); value: details.address }
+                PeerKeyValueRow { key: qsTr("VIA"); value: details.addressLocal }
+                PeerKeyValueRow { key: qsTr("Type"); value: details.type }
+                PeerKeyValueRow { key: qsTr("Permissions"); value: details.permission }
+                PeerKeyValueRow { key: qsTr("Version"); value: details.version }
+                PeerKeyValueRow { key: qsTr("User agent"); value: details.userAgent }
+                PeerKeyValueRow { key: qsTr("Services"); value: details.services }
+                PeerKeyValueRow { key: qsTr("Transaction relay"); value: details.transactionRelay }
+                PeerKeyValueRow { key: qsTr("Address relay"); value: details.addressRelay }
+                PeerKeyValueRow { key: qsTr("Mapped AS"); value: details.mappedAS }
+            }
+
+            CoreText {
+                text: "Block data";
+                bold: true;
+                font.pixelSize: 18;
+                horizontalAlignment: Qt.AlignLeft;
+                color: Theme.color.neutral9;
+            }
+            Column {
+                width: parent.width
+                bottomPadding: 5
+                PeerKeyValueRow { key: qsTr("Starting block"); value: details.startingHeight }
+                PeerKeyValueRow { key: qsTr("Synced headers"); value: details.syncedHeaders }
+                PeerKeyValueRow { key: qsTr("Synced blocks"); value: details.syncedBlocks }
+            }
+
+            CoreText {
+                text: "Network traffic";
+                bold: true;
+                font.pixelSize: 18;
+                horizontalAlignment: Qt.AlignLeft;
+                color: Theme.color.neutral9;
+            }
+            Column {
+                width: parent.width
+                bottomPadding: 5
+                PeerKeyValueRow { key: qsTr("Direction"); value: details.direction }
+                PeerKeyValueRow { key: qsTr("Connection time"); value: details.connectionDuration }
+                PeerKeyValueRow { key: qsTr("Last send"); value: details.lastSend }
+                PeerKeyValueRow { key: qsTr("Last receive"); value: details.lastReceived }
+                PeerKeyValueRow { key: qsTr("Sent"); value: details.bytesSent }
+                PeerKeyValueRow { key: qsTr("Received"); value: details.bytesReceived }
+                PeerKeyValueRow { key: qsTr("Ping time"); value: details.pingTime }
+                PeerKeyValueRow { key: qsTr("Ping wait"); value: details.pingWait }
+                PeerKeyValueRow { key: qsTr("Min ping"); value: details.pingMin }
+                PeerKeyValueRow { key: qsTr("Time offset"); value: details.timeOffset }
+            }
+        }
+    }
+}
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -41,6 +41,7 @@ QT_MOC_CPP = \
   qml/models/moc_networktraffictower.cpp \
   qml/models/moc_nodemodel.cpp \
   qml/models/moc_options_model.cpp \
+  qml/models/moc_peerdetailsmodel.cpp \
   qml/models/moc_peerlistsortproxy.cpp \
   qml/moc_appmode.cpp \
   qt/moc_addressbookpage.cpp \
@@ -120,6 +121,7 @@ BITCOIN_QT_H = \
   qml/models/networktraffictower.h \
   qml/models/nodemodel.h \
   qml/models/options_model.h \
+  qml/models/peerdetailsmodel.h \
   qml/models/peerlistsortproxy.h \
   qml/appmode.h \
   qml/bitcoin.h \
@@ -307,6 +309,7 @@ BITCOIN_QML_BASE_CPP = \
   qml/models/networktraffictower.cpp \
   qml/models/nodemodel.cpp \
   qml/models/options_model.cpp \
+  qml/models/peerdetailsmodel.cpp \
   qml/models/peerlistsortproxy.cpp \
   qml/imageprovider.cpp \
   qml/util.cpp
@@ -384,6 +387,7 @@ QML_RES_QML = \
   qml/pages/node/NodeRunner.qml \
   qml/pages/node/NodeSettings.qml \
   qml/pages/node/Peers.qml \
+  qml/pages/node/PeerDetails.qml \
   qml/pages/node/Shutdown.qml \
   qml/pages/onboarding/OnboardingBlockclock.qml \
   qml/pages/onboarding/OnboardingConnection.qml \
--- a/src/qml/bitcoin_qml.qrc
+++ b/src/qml/bitcoin_qml.qrc
@@ -45,6 +45,7 @@
         <file>pages/node/NodeRunner.qml</file>
         <file>pages/node/NodeSettings.qml</file>
         <file>pages/node/Peers.qml</file>
+        <file>pages/node/PeerDetails.qml</file>
         <file>pages/node/Shutdown.qml</file>
         <file>pages/onboarding/OnboardingBlockclock.qml</file>
         <file>pages/onboarding/OnboardingConnection.qml</file>
--- a/src/qml/models/peerdetailsmodel.cpp
+++ b/src/qml/models/peerdetailsmodel.cpp
@@ -1,0 +1,56 @@
+// Copyright (c) 2024 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include <qml/models/peerdetailsmodel.h>
+
+PeerDetailsModel::PeerDetailsModel(CNodeCombinedStats* nodeStats, PeerTableModel* parent)
+: m_combinedStats{nodeStats}
+, m_model{parent}
+, m_disconnected{false}
+{
+    for (int row = 0; row < m_model->rowCount(); ++row) {
+        QModelIndex index = m_model->index(row, 0);
+        int nodeIdInRow = m_model->data(index, PeerTableModel::NetNodeId).toInt();
+        if (nodeIdInRow == m_combinedStats->nodeStats.nodeid) {
+            m_row = row;
+            break;
+        }
+    }
+    connect(parent, &PeerTableModel::rowsRemoved, this, &PeerDetailsModel::onModelRowsRemoved);
+    connect(parent, &PeerTableModel::dataChanged, this, &PeerDetailsModel::onModelDataChanged);
+}
+
+void PeerDetailsModel::onModelRowsRemoved(const QModelIndex&  parent, int first, int last)
+{
+    for (int row = first; row <= last; ++row) {
+        QModelIndex index = m_model->index(row, 0, parent);
+        int nodeIdInRow = m_model->data(index, PeerTableModel::NetNodeId).toInt();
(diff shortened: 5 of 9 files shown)


src/qml/controls/SettingsButton.qml
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2022 The Bitcoin Core developers
```
jarolrod, line 1:
```suggestion
// Copyright (c) 2024 The Bitcoin Core developers
```
🫣

src/qml/controls/PeerRadioButton.qml
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2022 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import org.bitcoincore.qt 1.0
+
+RadioButton {
```
jarolrod, line 9: > This commit does not include the disconnect/ban buttons, the top right buttons to cycle through the peers, or the icons that are adjacent to the values.

We should remove this file then and the references in bitcoin_qml.qrc and the makefile

src/qml/bitcoin.cpp
```diff
@@ -292,6 +294,8 @@ int QmlGuiMain(int argc, char* argv[])
     qmlRegisterSingletonInstance<AppMode>("org.bitcoincore.qt", 1, 0, "AppMode", &app_mode);
     qmlRegisterType<BlockClockDial>("org.bitcoincore.qt", 1, 0, "BlockClockDial");
     qmlRegisterType<LineGraph>("org.bitcoincore.qt", 1, 0, "LineGraph");
+    qmlRegisterUncreatableType<PeerDetailsModel>("org.bitcoincore.qt", 1, 0, "PeerDetailsModel", "");
```
jarolrod, line 297: for reviewers: https://doc.qt.io/qt-6/qqmlengine.html#qmlRegisterUncreatableType

src/qml/pages/node/PeerDetails.qml
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    ScrollView {
+        id: scrollView
+        width: parent.width
+        height: parent.height
+        clip: true
+        contentWidth: width
+
+        Column {
```
jarolrod, line 62: we need rules and way to wrap these values if for whatever reason they get too long: 

Screenshot 2024-01-29 at 9 11 16 AM

src/qml/pages/node/PeerDetails.qml
```diff
@@ -0,0 +1,102 @@
...
+            headerSize: 18
+            header: qsTr("Peer " + details.nodeId)
+        }
+    }
+
+    component PeerKeyValueRow: Row {
+        property string key: ""
+        property string value: ""
+        CoreText {
+            color: Theme.color.white;
```
jarolrod, line 44:
```suggestion
            color: Theme.color.neutral9
```

and everywhere else where there is white for text currently

src/qml/controls/PeerRadioButton.qml
johnny9: Yes, I missed removing these ones from this commit

johnny9:
From[ 845a878 ](https://github.com/bitcoin-core/gui-qml/commit/845a8781d47e69c87f0233bb1d0091502bacb44b)to a660b9c
  - Removed unused controls (RadioButton and SettingsButton for disconnect/ban)

johnny9:
Update from[ a660b9c ](https://github.com/bitcoin-core/gui-qml/commit/a660b9c348f8f413d37ced95fbd3293468a88968)to 8c2a918
  - Replaced color white with color neutral9 for Dark mode switching

src/qml/pages/node/PeerDetails.qml
johnny9: i think a max width with an elipses should cover it. I'll see what that looks like.

johnny9:
Update  from ef6e826 to b5e2de78b5c19f43100a0f49949a3eac4b96ca8a
  - Fixed lint issues
  - Constrained text values with elide

src/qml/pages/node/PeerDetails.qml
johnny9: values now elide

MarnixCroes: user might want to view all _Services_
maybe be good to put the missing ones in a tooltip?

MarnixCroes:
b5e2de78b5c19f43100a0f49949a3eac4b96ca8a weird jumping when clicking on peer, at _Direction_, _User_ agent_, _Type_, _Ip_ and _Network_
cannot repro at _ID_


[jumping.webm](https://github.com/bitcoin-core/gui-qml/assets/93143998/31d31a7c-a512-4b01-bdb9-4c40120443bf)

johnny9:
> [b5e2de7](https://github.com/bitcoin-core/gui-qml/commit/b5e2de78b5c19f43100a0f49949a3eac4b96ca8a) weird jumping when clicking on peer, at _Direction_, _User_ agent_, _Type_, _Ip_ and _Network_ cannot repro at _ID_
> jumping.webm

This appears to be due to using the wrong index for the original model. I have updated the branch with a working solution

johnny9:
Update from b5e2de7 to b7161c5 
  - Don't use the index from the proxy table, instead find the index for the details from the original table model.

src/qml/pages/node/PeerDetails.qml
```diff
@@ -0,0 +1,111 @@
...
+
+        Column {
+            width: Math.min(parent.width - 40, 450)
+            anchors.horizontalCenter: parent.horizontalCenter
+            spacing: 20
+
+            CoreText { text: "Information"; bold: true; font.pixelSize: 18; horizontalAlignment: Qt.AlignLeft; }
+            Column {
+                width: parent.width
+                PeerKeyValueRow { key: qsTr("IP"); value: details.address }
```
MarnixCroes, line 76:
```suggestion
                PeerKeyValueRow { key: qsTr("Address"); value: details.address }
```
can also be onion service/ other networks

src/qml/pages/node/PeerDetails.qml
```diff
@@ -0,0 +1,111 @@
...
+                PeerKeyValueRow { key: qsTr("VIA"); value: details.addressLocal }
+                PeerKeyValueRow { key: qsTr("Type"); value: details.type }
+                PeerKeyValueRow { key: qsTr("Permission"); value: details.permission }
+                PeerKeyValueRow { key: qsTr("Version"); value: details.version }
+                PeerKeyValueRow { key: qsTr("User agent"); value: details.userAgent }
+                PeerKeyValueRow { key: qsTr("Services"); value: details.services }
+                PeerKeyValueRow { key: qsTr("Transport relay"); value: details.transactionRelay }
```
MarnixCroes, line 83:
```suggestion
                PeerKeyValueRow { key: qsTr("Transaction relay"); value: details.transactionRelay }
```

src/qml/pages/node/PeerDetails.qml
```diff
@@ -0,0 +1,111 @@
...
+                PeerKeyValueRow { key: qsTr("Address relay"); value: details.addressRelay }
+                PeerKeyValueRow { key: qsTr("Mapped AS"); value: details.mappedAS }
+            }
+
+            CoreText { text: "Block data"; bold: true; font.pixelSize: 18; horizontalAlignment: Qt.AlignLeft; }
+            Column {
+                width: parent.width
+                PeerKeyValueRow { key: qsTr("Starting block"); value: details.startingHeight }
```
MarnixCroes, line 91: `Starting Block` doesn't seem to work correctly, for me this always shows 0
(testing with a synced signet node, normal bitcoin qt it works fine)

MarnixCroes (commented):
what was the motivation behind current shown details?
some things like _Connection Time_ (Age), _Network_, _Last Block_, _Last Transaction_ are missing

src/qml/pages/node/PeerDetails.qml
MarnixCroes: also things like onion address should be fully visible imo
 https://github.com/bitcoin-core/gui-qml/pull/375#pullrequestreview-1759750423

GBKS:
Great work. I tested it out on Android. Looks great overall, just see some inconsistencies to the design in terms of vertical spacing and missing formatting.

For text that does not fit in one line, I think it should just go multi-line. If you shorten it, you then need a way to expand it (which could be a simple tap).

In the image below, the Android screenshot is on the right, and the Figma mock-ups are on the left.

image

A few questions:
- What does a "Synced headers" value of "-1" mean? That the process is not initialized?
- What are permissions? What does "N/A" mean?
- What does "Mapped AS" mean?
- For sent/received, does it maybe make sense to show the percentage that is of the total send/receive? As an indicator of how much your node relies on this peer?

MarnixCroes:
@GBKS 

> What does a "Synced headers" value of "-1" mean? That the process is not initialized?

yes, it means that that peer didn't announce that info yet

> What are permissions? What does "N/A" mean?

any special permissions that have been granted to that peer. N/A means _Not Applicable_ / _Not Assessed_

> What does "Mapped AS" mean?

_Mapped Autonomous System_, which can be used for diversifying peer selection

> For sent/received, does it maybe make sense to show the percentage that is of the total send/receive? As an indicator of how much your node relies on this peer?

As peers are connecting and disconnecting all the time this is quite hard: Do you measure it for all currently connected peers, since the node is up, since first launch... etc.
**and** these values aren't that meaningful: a higher amount sent/received doesn't necessarily mean that the node relies more on this peer than another peer with a lower amount.

So, I get what you mean, but this shouldn't be added imo

johnny9:
> Great work. I tested it out on Android. Looks great overall, just see some inconsistencies to the design in terms of vertical spacing and missing formatting.
> 

Ill adjust that spacing. Looks like mine is too tight.

> For text that does not fit in one line, I think it should just go multi-line. If you shorten it, you then need a way to expand it (which could be a simple tap).

I'll use word wrap on the next update

> A few questions:
> 
>     * What does a "Synced headers" value of "-1" mean? That the process is not initialized

It means that it hasn't synced any headers with this peer yet. -1 is the default vault. I think this should just probably just be a minus as well until this data updates.

>     * What are permissions? What does "N/A" mean?

These are all of the permission strings "bloomfilter", "noban", "forcerelay" "relay" "mempool" "download" "addr". N/A just means that none are set. These permissions are assigned by using the "-whitelist" configuration option. Here are the description of these flags. 
    "bloomfilter (allow requesting BIP37 filtered blocks and transactions)",
    "noban (do not ban for misbehavior; implies download)",
    "forcerelay (relay transactions that are already in the mempool; implies relay)",
    "relay (relay even in -blocksonly mode, and unlimited transaction announcements)",
    "mempool (allow requesting BIP35 mempool contents)",
    "download (allow getheaders during IBD, no disconnect after maxuploadtarget limit)",
    "addr (responses to GETADDR avoid hitting the cache and contain random records with the most up-to-date info)"

>     * What does "Mapped AS" mean?
Mapped AS means Mapped Autonomous System which is a configuration option that lets you map peers to AS numbers. AS numbers are assigned by top level internet registries to organizations with a bunch of network addresses. You specify a mapping file to bitcoin core with a configuration option.

>     * For sent/received, does it maybe make sense to show the percentage that is of the total send/receive? As an indicator of how much your node relies on this peer?
Yeah I think there is a lot of value in that.

src/qml/pages/node/PeerDetails.qml
johnny9: I noticed that the Qt gui uses a value from a different struct. I will switch to that and see if that resovles it.

johnny9:
Update from[ b7161c5 ](https://github.com/bitcoin-core/gui-qml/commit/b7161c5035b7dbfae55bd447abb60d7cb3587112)to d737ea3
  - Used correct struct for m_starting_height value to fix "Starting Height"
  - Changed IP to Address
  - Adjusted spacing to match figma
  - Allowed Text to WordWrap

MarnixCroes:
running d737ea3406c3e8e99a5a93f31dc03725a06a0823, I now had twice that the peer details  page closed itself randomly and returned to the peers page.
peer didn't disconnect, not sure what is causing it.

> Changed IP to Address

forgot to add the change?

> Allowed Text to WordWrap

this looks good 👍

src/qml/pages/node/PeerDetails.qml
MarnixCroes: yes it works now

GBKS:
Thanks for the explainers and tweaking the layout.

> These permissions are assigned by using the "-whitelist" configuration option

Does that mean permissions are application-wide and the same for all peers? If so, maybe they can be shown elsewhere...?

> I think this should just probably just be a minus as well until this data updates.

Are we using "N/A" and "-" interchangeably? Looks like we need states for
- Pending: we're still waiting to find out (maybe "..."?)
- Not available: we're not waiting anymore, and have no idea ("N/A")
- None: none were set ("-")

Does that seem about right? What's the likelihood of getting "not available"? Is that even possible?

GBKS:
> I now had twice that the peer details page closed itself randomly and returned to the peers page

I also had this happen.

yashrajd:
Good stuff guys.

In line with external feedback we received in past Bitcoin Core calls, a little tooltip and explainers for these items would be extremely useful.

Even the tip that @johnny9 provided about how peer permissions can be changed would be useful (on this page or somewhere else)

> @GBKS
> 
> > What does a "Synced headers" value of "-1" mean? That the process is not initialized?
> 
> yes, it means that that peer didn't announce that info yet
> 
> > What are permissions? What does "N/A" mean?
> 
> any special permissions that have been granted to that peer. N/A means _Not Applicable_ / _Not Assessed_
> 
> > What does "Mapped AS" mean?
> 
> _Mapped Autonomous System_, which can be used for diversifying peer selection
> 
> > For sent/received, does it maybe make sense to show the percentage that is of the total send/receive? As an indicator of how much your node relies on this peer?
> 
> As peers are connecting and disconnecting all the time this is quite hard: Do you measure it for all currently connected peers, since the node is up, since first launch... etc. **and** these values aren't that meaningful: a higher amount sent/received doesn't necessarily mean that the node relies more on this peer than another peer with a lower amount.
> 
> So, I get what you mean, but this shouldn't be added imo

GBKS:
For this PR generally, I'm happy to ACK it whenever the functionality is solid and we can tweak visualization details and tips as we live with it during practical usage.

MarnixCroes (changes requested):
_Transport_ version and _Connection Time_ with the peer should be added imo

MarnixCroes:
>  I now had twice that the peer details page closed itself randomly and returned to the peers page.
peer didn't disconnect, not sure what is causing it.

running b4f8fc140487649e6d368941ad1a85cad9adfade this didn't happen

johnny9:
> _Transport_ version and _Connection Time_ with the peer should be added imo

I think we need to re-sync with main before we add transport type but I can add Connection Time to the Network traffic section

johnny9:
Update from[ b4f8fc1 ](https://github.com/bitcoin-core/gui-qml/commit/b4f8fc140487649e6d368941ad1a85cad9adfade)to c8344d8
 - Prevented disconnect signal from triggering twice
 - Set color correctly on the section titles
 - Added "Connection time"

MarnixCroes:
> _Transport_ version and _Connection Time_ with the peer should be added imo

and the _Network_ (ipv4, onion etc), 
sorry forgot about that one

MarnixCroes (approved):
apart from some minor things, like the dialog sometimes randomly closing.
I think this is good enough and improvements can be made later on

ACK c8344d8a54de3a0b44788f304c6e134227e98e87

nice work

src/qml/pages/node/PeerDetails.qml
```diff
@@ -0,0 +1,132 @@
...
+                horizontalAlignment: Qt.AlignLeft;
+                color: Theme.color.neutral9;
+            }
+            Column {
+                width: parent.width
+                spacing: 5
+                PeerKeyValueRow { key: qsTr("Address"); value: details.address }
+                PeerKeyValueRow { key: qsTr("VIA"); value: details.addressLocal }
+                PeerKeyValueRow { key: qsTr("Type"); value: details.type }
+                PeerKeyValueRow { key: qsTr("Permission"); value: details.permission }
```
jarolrod, line 85:
```suggestion
                PeerKeyValueRow { key: qsTr("Permissions"); value: details.permission }
```

src/qml/pages/node/PeerDetails.qml
```diff
@@ -0,0 +1,132 @@
...
+            CoreText {
+                text: "Information";
+                bold: true;
+                font.pixelSize: 18;
+                horizontalAlignment: Qt.AlignLeft;
+                color: Theme.color.neutral9;
+            }
+            Column {
+                width: parent.width
+                spacing: 5
```
jarolrod, line 81: The spacing between sections of information is actually 15, so we'd want to add a bottom margin of 5

```diff

diff --git a/src/qml/pages/node/PeerDetails.qml b/src/qml/pages/node/PeerDetails.qml
index 224aef8c7..b32c2a0c7 100644
--- a/src/qml/pages/node/PeerDetails.qml
+++ b/src/qml/pages/node/PeerDetails.qml
@@ -79,6 +79,7 @@ Page {
             Column {
                 width: parent.width
                 spacing: 5
+                bottomPadding: 5
                 PeerKeyValueRow { key: qsTr("Address"); value: details.address }
                 PeerKeyValueRow { key: qsTr("VIA"); value: details.addressLocal }
                 PeerKeyValueRow { key: qsTr("Type"); value: details.type }
@@ -101,6 +102,7 @@ Page {
             Column {
                 width: parent.width
                 spacing: 5
+                bottomPadding: 5
                 PeerKeyValueRow { key: qsTr("Starting block"); value: details.startingHeight }
                 PeerKeyValueRow { key: qsTr("Synced headers"); value: details.syncedHeaders }
                 PeerKeyValueRow { key: qsTr("Synced blocks"); value: details.syncedBlocks }
@@ -116,6 +118,7 @@ Page {
             Column {
                 width: parent.width
                 spacing: 5
+                bottomPadding: 5
                 PeerKeyValueRow { key: qsTr("Direction"); value: details.direction }
                 PeerKeyValueRow { key: qsTr("Connection time"); value: details.connectionDuration }
                 PeerKeyValueRow { key: qsTr("Last send"); value: details.lastSend }

```

jarolrod: See here, this starts to look like the design file
| master | patch |
| ------ | ----- |
| section-spacing-master | section-spacing-patch |

Design file screenshot of section spacing: 

spacing-between-sections-15

src/qml/pages/node/PeerDetails.qml
```diff
@@ -0,0 +1,132 @@
...
+        id: scrollView
+        width: parent.width
+        height: parent.height
+        clip: true
+        contentWidth: width
+
+        Column {
+            width: Math.min(parent.width - 40, 450)
+            anchors.horizontalCenter: parent.horizontalCenter
+            spacing: 10
```
jarolrod, line 70: we want to add top and bottom padding of 30 to adhere to the design file

```suggestion
            spacing: 10
            topPadding: 30
            bottomPadding: 30
```

jarolrod: See here, this starts to look like the design file
| master | patch |
| ------ | ----- |
| top-bottom-padding-master | top-bottom-padding-patch |



Design file screenshot of top padding:

Screenshot 2024-02-13 at 10 30 16 PM

jarolrod (commented):
Almost there for the basic functionality of this page, some spacing is not adherent to the design file

src/qml/pages/node/PeerDetails.qml
```diff
@@ -0,0 +1,132 @@
...
+            headerBold: true
+            headerSize: 18
+            header: qsTr("Peer " + details.nodeId)
+        }
+    }
+
+    component PeerKeyValueRow: Row {
+        width: parent.width
+        property string key: ""
+        property string value: ""
```
jarolrod, line 43: also
```suggestion
        property string value: ""
        spacing: 10
```

johnny9:
Update from[ c8344d8 ](https://github.com/bitcoin-core/gui-qml/commit/c8344d8a54de3a0b44788f304c6e134227e98e87)to[ eb5030f](https://github.com/bitcoin-core/gui-qml/commit/eb5030f55465d3f10fdbe59ff30c57915edf794c)
 - incorporated all of @jarolrod suggestions (good eye)
 - set height of PeerKeyValueRow to 21 as specified by the Figma and removed the spacing value for the Columns that hold those Rows.

MarnixCroes (commented):
eb5030f55465d3f10fdbe59ff30c57915edf794c the new spacing can cause some overlap

image

johnny9:
> [eb5030f](https://github.com/bitcoin-core/gui-qml/commit/eb5030f55465d3f10fdbe59ff30c57915edf794c) the new spacing can cause some overlap
> 
> image

ok updated with the height attached to the lineCount of the value

MarnixCroes (approved):
ack fa4e080f7261fa66a8db0e6fbc3449ad305e37e4

yashrajd:
Just ran the latest artefact, and it looks pretty good. Didn't see any spacing issues.

GBKS:
> Thank you, updated with the height attached to the lineCount of the value

Do I see it correctly that the line height is hard-coded? Users can typically adjust text size via accessibility settings, which means the line height will also differ. Is there a way to automatically calculate this value?

jarolrod:
https://github.com/bitcoin-core/gui-qml/pull/388 is proposed as an alternative which handles some of the outstanding issues, implements more functionality

johnny9:
Closing this as it is superseded with #388

PR #391, opened 2024-03-22
Introducing an improved IP Address and Port input control in ProxySettings component

pablomartin4btc:
Adding a new improved `ValueInput` control to handle both `IP` address and port values combined in the same field (as it was designed for `QML`, current `QT` project have them separately) with the correct visual formatting (`255.255.255.255:65535`).

In order to add/ change the `IP` address and port value, user needs to enable the fields (currently in `main` branch this doesn't work and this PR fixes it). Currently the `IP` address and port input fields don't allow more than 5 digits and this PR also correct it so the user can enter a complete `IP` address and port value.

<details>
<summary>If an invalid <code>IP</code> address (e.g. <code>127.0.:9050</code>) and/ or port (e.g. <code>127.0.0.1:</code>) are entered an error message will be shown as we do in other fields.</summary>

image

</details>

<details>
<summary>Current <code>master</code>/ <code>main</code> branch screenshot.</summary>

image

</details>

<details>
<summary>Current master branch in <code>QT</code> <a href="https://github.com/bitcoin-core/gui">repo</a> (just for reference).</summary>

image

</details>

<details>
<summary>This PRs branch screenshot.</summary>

image

</details>

<details>
<summary><a href="https://www.figma.com/file/ek8w3n3upbluw5UL2lGhRx/Bitcoin-Core-App-Design?type=design&node-id=7635-19847&mode=design&t=Ns1KRBputDp3sBjs-0">Design</a> in Figma.</summary>

image

</details>

---

**Implementation details**:

In this iteration, validation for this control is implemented using both a function within the control itself focused solely on ensuring correct formatting and a simple regex validator for the valid input characters.
Future versions will see integration with network model classes, enabling parsing of IP addresses and ports
and additional checks, such as those outlined in [`doc/p2p-bad-ports.md`](https://github.com/bitcoin/bitcoin/tree/master/doc/p2p-bad-ports.md).

**Note**: 

Persistence/ wiring of the settings and creating the proper network connection thru the Proxy will be performed in a different PR, the idea is to isolate UI experience/ testing in case we need to make changes on that aspect separately and we concentrate on the functionality behind later.

### DIFF
--- a/src/qml/components/ProxySettings.qml
+++ b/src/qml/components/ProxySettings.qml
@@ -20,9 +20,18 @@ ColumnLayout {
     }
     Separator { Layout.fillWidth: true }
     Setting {
+        id: defaultProxyEnable
         Layout.fillWidth: true
         header: qsTr("Enable")
-        actionItem: OptionSwitch {}
+        actionItem: OptionSwitch {
+            onCheckedChanged: {
+                if (checked == false) {
+                    defaultProxy.state = "DISABLED"
+                } else {
+                    defaultProxy.state = "FILLED"
+                }
+            }
+        }
         onClicked: {
             loadedItem.toggle()
             loadedItem.toggled()
@@ -33,6 +42,8 @@ ColumnLayout {
         id: defaultProxy
         Layout.fillWidth: true
         header: qsTr("IP and Port")
+        errorText: qsTr("Invalid IP address or port format. Please use the format '255.255.255.255:65535'.")
+        state: !defaultProxyEnable.loadedItem.checked ? "DISABLED" : "FILLED"
         actionItem: ValueInput {
             parentState: defaultProxy.state
             description: "127.0.0.1:9050"
@@ -55,10 +66,18 @@ ColumnLayout {
     }
     Separator { Layout.fillWidth: true }
     Setting {
+        id: torProxyEnable
         Layout.fillWidth: true
         header: qsTr("Enable")
-        actionItem: OptionSwitch {}
-        description: qsTr("When disabled, Tor connections will use the default proxy (if enabled).")
+        actionItem: OptionSwitch {
+            onCheckedChanged: {
+                if (checked == false) {
+                    torProxy.state = "DISABLED"
+                } else {
+                    torProxy.state = "FILLED"
+                }
+            }
+        }
         onClicked: {
             loadedItem.toggle()
             loadedItem.toggled()
@@ -69,6 +88,8 @@ ColumnLayout {
         id: torProxy
         Layout.fillWidth: true
         header: qsTr("IP and Port")
+        errorText: qsTr("Invalid IP address or port format. Please use the format '255.255.255.255:65535'.")
+        state: !torProxyEnable.loadedItem.checked ? "DISABLED" : "FILLED"
         actionItem: ValueInput {
             parentState: torProxy.state
             description: "127.0.0.1:9050"
--- a/src/qml/components/ProxySettings.qml
+++ b/src/qml/components/ProxySettings.qml
@@ -44,14 +44,16 @@ ColumnLayout {
         header: qsTr("IP and Port")
         errorText: qsTr("Invalid IP address or port format. Please use the format '255.255.255.255:65535'.")
         state: !defaultProxyEnable.loadedItem.checked ? "DISABLED" : "FILLED"
-        actionItem: ValueInput {
+        showErrorText: !defaultProxy.loadedItem.validInput && defaultProxyEnable.loadedItem.checked
+        actionItem: IPAddressValueInput {
             parentState: defaultProxy.state
             description: "127.0.0.1:9050"
-            onEditingFinished: {
-                defaultProxy.forceActiveFocus()
-            }
+            activeFocusOnTab: true
         }
-        onClicked: loadedItem.forceActiveFocus()
+        onClicked: {
+            loadedItem.filled = true
+            loadedItem.forceActiveFocus()
+        }
     }
     Separator { Layout.fillWidth: true }
     Header {
@@ -90,14 +92,16 @@ ColumnLayout {
         header: qsTr("IP and Port")
         errorText: qsTr("Invalid IP address or port format. Please use the format '255.255.255.255:65535'.")
         state: !torProxyEnable.loadedItem.checked ? "DISABLED" : "FILLED"
-        actionItem: ValueInput {
+        showErrorText: !torProxy.loadedItem.validInput && torProxyEnable.loadedItem.checked
+        actionItem: IPAddressValueInput {
             parentState: torProxy.state
             description: "127.0.0.1:9050"
-            onEditingFinished: {
-                torProxy.forceActiveFocus()
-            }
+            activeFocusOnTab: true
         }
-        onClicked: loadedItem.forceActiveFocus()
+        onClicked: {
+            loadedItem.filled = true
+            loadedItem.forceActiveFocus()
+        }
     }
     Separator { Layout.fillWidth: true }
 }
--- a/src/qml/controls/IPAddressValueInput.qml
+++ b/src/qml/controls/IPAddressValueInput.qml
@@ -1,0 +1,82 @@
+// Copyright (c) 2024 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+
+TextInput {
+    id: root
+    required property string parentState
+    property string description: ""
+    property bool filled: false
+    property int descriptionSize: 18
+    property color textColor: root.filled ? Theme.color.neutral9 : Theme.color.neutral5
+    // Expose a property to indicate validity, initial value will be true (no error message displayed)
+    property bool validInput: true
+    enabled: true
+    state: root.parentState
+    validator: RegExpValidator { regExp: /[0-9.:]*/ } // Allow only digits, dots, and colons
+
+    maximumLength: 21
+
+    states: [
+        State {
+            name: "ACTIVE"
+            PropertyChanges { target: root; textColor: Theme.color.orange }
+        },
+        State {
+            name: "HOVER"
+            PropertyChanges {
+                target: root
+                textColor: root.filled ? Theme.color.orangeLight1 : Theme.color.neutral5
+            }
+        },
+        State {
+            name: "DISABLED"
+            PropertyChanges {
+                target: root
+                enabled: false
+                textColor: Theme.color.neutral4
+            }
+        }
+    ]
+
+    font.family: "Inter"
+    font.styleName: "Regular"
+    font.pixelSize: root.descriptionSize
+    color: root.textColor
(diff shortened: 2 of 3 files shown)


src/qml/components/ProxySettings.qml
```diff
@@ -33,14 +42,26 @@ ColumnLayout {
...
+        errorText: qsTr("Invalid IP address or port format. Please use the format '255.255.255.255:65535'.")
+        showErrorText: !loadedItem.isValidIPPort(loadedItem.description)
+        state: !defaultProxyEnable.loadedItem.checked ? "DISABLED" : "ACTIVE"
+        actionItem: IPAddressValueInput {
             parentState: defaultProxy.state
             description: "127.0.0.1:9050"
+            activeFocusOnTab: true
             onEditingFinished: {
-                defaultProxy.forceActiveFocus()
+                if (isValidIPPort(text)) {
```
johnny9, line 53: I think it might be a good idea to encapsulate the validation logic in the IPAddressValueInput component itself. I think one possible way to do that is to add a Connections element to onEditingFinished signal in the IPAddressValueInput definition so the validation can't be overridden. Then just expose a bool for if the input is valid or not.

pablomartin4btc: Ok, it makes sense, I'll try to do that. Thanks!

GBKS:
Just tested on Android mobile. This is a nice update.

I don't think the input field needs to turn orange when it goes from disabled to enabled state (by toggling the line above).

Also interesting how the "Enable" title briefly turns orange when tapping. Feels a bit surprising and unnecessary. I think the title could just remain white.

The copy for the Tor proxy is too confusing still. 

> Enable to run Tor connections through a dedicated proxy.

> When disabled, Tor connections will use the default proxy (if enabled).

Can we remove that second line?

src/qml/components/ProxySettings.qml
pablomartin4btc: Updated it with the `Connection` element and removed the `onEditingFinished` from `IPAddressValueInput`, please let me know what you think.

pablomartin4btc:
@GBKS

> I don't think the input field needs to turn orange when it goes from disabled to enabled state (by toggling the line above).
> 
> Also interesting how the "Enable" title briefly turns orange when tapping. Feels a bit surprising and unnecessary. I think the title could just remain white.

I've incorrectly used a different `State` in the code of the input control (`ACTIVE`), I've fixed it with the correct one now (`FILLED`).
<details>
<summary>The new <code>IPAddressValueInput</code> control follows now the same pattern as the <code>ValueInput</code> control ("Block Storage Limit (GB)") in the "Storage Settings" page.</summary>

image


</details>

> Can we remove that second line?

Yes, that wasn't touched by this PR, but corrected it as requested.

pablomartin4btc:
- Updates: I've addressed feedback from @johnny9, @GBKS and @D33r-Gee. Thanks for your valuable reviews!

src/qml/components/ProxySettings.qml
```diff
@@ -69,14 +89,17 @@ ColumnLayout {
         id: torProxy
         Layout.fillWidth: true
         header: qsTr("IP and Port")
+        errorText: qsTr("Invalid IP address or port format. Please use the format '255.255.255.255:65535'.")
+        state: !torProxyEnable.loadedItem.checked ? "DISABLED" : "FILLED"
         actionItem: ValueInput {
```
johnny9, line 94:
```suggestion
        actionItem: IPAddressValueInput {
```

src/qml/components/ProxySettings.qml
```diff
@@ -69,14 +89,17 @@ ColumnLayout {
         id: torProxy
         Layout.fillWidth: true
         header: qsTr("IP and Port")
+        errorText: qsTr("Invalid IP address or port format. Please use the format '255.255.255.255:65535'.")
+        state: !torProxyEnable.loadedItem.checked ? "DISABLED" : "FILLED"
```
johnny9, line 93:
```suggestion
        state: !torProxyEnable.loadedItem.checked ? "DISABLED" : "FILLED"
        showErrorText: !torProxy.loadedItem.validInput
```

src/qml/components/ProxySettings.qml
```diff
@@ -33,14 +42,17 @@ ColumnLayout {
         id: defaultProxy
         Layout.fillWidth: true
         header: qsTr("IP and Port")
+        errorText: qsTr("Invalid IP address or port format. Please use the format '255.255.255.255:65535'.")
+        state: !defaultProxyEnable.loadedItem.checked ? "DISABLED" : "FILLED"
```
johnny9, line 46:
```suggestion
        state: !defaultProxyEnable.loadedItem.checked ? "DISABLED" : "FILLED"
        showErrorText: !defaultProxy.loadedItem.validInput
```

src/qml/components/ProxySettings.qml
```diff
@@ -33,14 +42,17 @@ ColumnLayout {
...
         actionItem: ValueInput {
```
johnny9, line 47:
```suggestion
        actionItem: IPAddressValueInput {
```

src/qml/controls/IPAddressValueInput.qml
```diff
@@ -0,0 +1,83 @@
...
+TextInput {
+    id: root
+    required property string parentState
+    property string description: ""
+    property bool filled: false
+    property int descriptionSize: 18
+    property color textColor: root.filled ? Theme.color.neutral9 : Theme.color.neutral5
+    // Expose a property to indicate validity, initial value will be true (no error message displayed)
+    // if both properties errortText and description are set
+    property bool validInput: description!=="" && errorText!==""
```
johnny9, line 17: errorText isn't within this scope. I think just initializing to true works here.
```suggestion
    property bool validInput: true
```

johnny9 (changes requested):
Noticed some missing pieces and a bug in initial state. I left comments and suggested changes. Wrong actionItem type was used and showErrorText needed to be connected to the loadedItem.validInput property that you added.

src/qml/controls/IPAddressValueInput.qml
pablomartin4btc: Ok.

pablomartin4btc:
> Noticed some missing pieces and a bug in initial state. I left comments and suggested changes. Wrong actionItem type was used and showErrorText needed to be connected to the loadedItem.validInput property that you added.

It seems I did something wrong when I updated the 1st commit. I'll correct it in a bit with your `validInput` initialisation value suggestion.

pablomartin4btc:
- Updates: I've solved the issue reported by @johnny9.

MarnixCroes (commented):
- it's unintuitive imo that the errorText shows up only after enabling/disabling and/or closing the dialog or window
- errorText is displayed at both, while only one is invalid
Screenshot from 2024-03-26 15-32-33

src/qml/controls/IPAddressValueInput.qml
```diff
@@ -0,0 +1,82 @@
...
+        }
+        var port = parseInt(parts[1]);
+        if (port < 1 || port > 65535) return false;
+        return true;
+    }
+
+    // Connections element to ensure validation on editing finished
+    Connections {
+        target: root
+        onEditingFinished: {
```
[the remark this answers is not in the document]
pablomartin4btc: Ok, I tried that syntax before, somehow QT didn't like it, but works now:
```

    // Connections element to ensure validation on editing finished
    Connections {
        target: root
        function onEditingFinished() {
            // Validate the input whenever editing is finished
            validInput = isValidIPPort(root.text);
        }
    }
```

I'll update the code, thanks!

By the way, out of curiosity, how did you get that deprecated message displayed? Perhaps I need to update my QT version:

```
qmake -v
QMake version 3.1
Using Qt version 5.15.3 in /usr/lib/x86_64-linux-gnu

```

Did you build `depends`? (I think that's `5.15.5`)

pablomartin4btc:
> * errorText is displayed at both, while only one is invalid

Sorry, you've found another bug in the code, the Tor IP input points to the Default Proxy one. I'll fix it immediately.
> * it's unintuitive imo that the errorText shows up only after enabling/disabling and/or closing the dialog or window

The `errorText` will be shown only `onEditingFinished` (another way could be `onTextChanged`, but it's too intrusive I think as the user perhaps didn't finish to enter the value and already sees the error message), but it's a good point that even the IP is invalid we should not display the error if we disable the input field. Is that your point?

pablomartin4btc:
- Updates: I've addressed @MarnixCroes's feedback.

src/qml/controls/IPAddressValueInput.qml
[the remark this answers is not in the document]
MarnixCroes: I'm running the exact same as you.
it only happened to me when I ran with `-resetguisettings`

MarnixCroes:
> but it's too intrusive I think as the user perhaps didn't finish to enter the value and already sees the error message)

I understand your point, but for me I'd prefer that behavior. The few users who use this and enter it will only enter it rarely. (in contrast to maybe regularly manually typing in a BTC address, then yes such behavior might be annoying.)
Because now I enter it, make a typo and navigate back. There was no indication for me that I made a mistake, until I navigate back to the dialog.

> is invalid we should not display the error if we disable the input field. Is that your point?

no, the above was.
no strong feelings about this one, both are ok imo

> If so, should we go back to the default value 127.0.0.1:9050 or leave it there even invalid 

yes I'd say to leave it, as user intent was to enter something else

pablomartin4btc:
> Because now I enter it, make a typo and navigate back. There was no indication for me that I made a mistake, until I navigate back to the dialog.

Ok, let me update that. It would be also good to see others reviewers takes on this.

pablomartin4btc:
> is a bit glitchy when hovering

Thanks, I'll investigate what's going on.

pablomartin4btc:
- <ins>Updates</ins>: I've addressed latest @MarnixCroes's feedback.
  - Fixed the [reported issue](https://github.com/bitcoin-core/gui-qml/pull/391#issuecomment-2021255688) on new `IPAddressValueInput` field being glitchy when hovering;
  - Moved validation logic from the `onEditingFinished` event into `onTextChanged` one, [as requested according to the specified use case](https://github.com/bitcoin-core/gui-qml/pull/391#issuecomment-2021250132).

johnny9 (approved):
ACK d47bc1ae97faab8be3bee3e0c20d92eb3cef5e0a

MarnixCroes (approved):
ack d47bc1ae97faab8be3bee3e0c20d92eb3cef5e0a

stackingsaunter:
Tested on macOS M1, very nice!

One thing I would change:

> Tor connections can also be run through a seperate Tor proxy

I feel like this is redundant here a bit, and Tor proxy description says it all

pablomartin4btc:
>       * Moved validation logic from the `onEditingFinished` event into `onTextChanged` one, [as requested according to the specified use case](https://github.com/bitcoin-core/gui-qml/pull/391#issuecomment-2021250132).

@GBKS could you please comment on this? If you agree on where the validation takes place at the moment? Otherwise we might have to update it in #430. Many thanks!

PR #273, opened 2023-02-22
DRAFT: Custom data directory support

jarolrod:
Opened to showcase initial support. Lacks system settings wiring.

| Dark | Light |
| ---- | ----- |
| Screen Shot 2023-02-22 at 3 37 32 AM | Screen Shot 2023-02-22 at 3 36 54 AM |

[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/<PR>)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/<PR>)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/<PR>)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/insecure_android_apk.zip?branch=pull/<PR>)

PR #284, opened 2023-03-09
Only write options after onboarding

johnny9:
[Windows](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/win64/insecure_win_gui.zip?branch=pull/284)
[Intel macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos/insecure_mac_gui.zip?branch=pull/284)
[Apple Silicon macOS](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/macos_arm64/insecure_mac_arm64_gui.zip?branch=pull/284)
[ARM64 Android](https://api.cirrus-ci.com/v1/artifact/github/bitcoin-core/gui-qml/android/unsecure_android_apk.zip?branch=pull/284)

PR #397, opened 2024-04-11
UI Only Custom Datadir Display

D33r-Gee:
This pull request builds upon #392 and introduces enhancements to display the data directory information within the UI. This functionality encompasses both default and custom data directory paths, fulfilling the UI requirements for user-defined data directory selection initiated in #273. 

Also the custom datadir is not persistent at the moment it will be once the back end wiring is added.

<details>

datadir_desktop

</details>

<details>
<summary>Android Screenshots</summary>

datadir_mobile_720

</details>

As a potential follow-up enhancement, consider incorporating mechanisms for saving the data directory path. This could be achieved through:

- Double-click functionality: Allow users to save the displayed path by simply double-clicking on it. This provides a convenient and intuitive method for desktop environments.

- Dedicated button: For mobile use cases or scenarios where double-clicking might not be feasible, introduce a dedicated "Save Path" button. This ensures a clear and accessible action for users on various devices.

### DIFF
--- a/src/qml/components/StorageLocations.qml
+++ b/src/qml/components/StorageLocations.qml
@@ -17,33 +17,52 @@ ColumnLayout {
     }
     spacing: 15
     OptionButton {
+        id: defaultDirOption
         Layout.fillWidth: true
         ButtonGroup.group: group
         text: qsTr("Default")
         description: qsTr("Your application directory.")
-        recommended: true
-        checked: true
+        customDir: optionsModel.getDefaultDataDirString
+        checked: optionsModel.dataDir === optionsModel.getDefaultDataDirString
+        onClicked: {
+            defaultDirOption.checked = true
+            optionsModel.dataDir = optionsModel.getDefaultDataDirString
+        }
     }
     OptionButton {
+        id: customDirOption
         Layout.fillWidth: true
         ButtonGroup.group: group
         text: qsTr("Custom")
         description: qsTr("Choose the directory and storage device.")
+        customDir: customDirOption.checked ? fileDialog.folder : ""
+        checked: optionsModel.dataDir !== optionsModel.getDefaultDataDirString
         onClicked: fileDialog.open()
     }
     FileDialog {
         id: fileDialog
         selectFolder: true
-        folder: optionsModel.getDefaultDataDirectory
+        folder: shortcuts.home
         onAccepted: {
             optionsModel.setCustomDataDirString(fileDialog.fileUrls[0].toString())
             var customDataDir = fileDialog.fileUrl.toString();
             if (customDataDir !== "") {
-                optionsModel.setCustomDataDirArgs(customDataDir);
+                optionsModel.setCustomDataDirArgs(customDataDir)
+                customDirOption.customDir = optionsModel.getCustomDataDirString()
+                if (optionsModel.dataDir !== optionsModel.getDefaultDataDirString) {
+                    customDirOption.checked = true
+                    defaultDirOption.checked = false
+                }
             }
         }
         onRejected: {
             console.log("Custom datadir selection canceled")
+            if (optionsModel.dataDir !== optionsModel.getDefaultDataDirString) {
+                customDirOption.checked = true
+                defaultDirOption.checked = false
+            } else {
+                defaultDirOption.checked = true
+            }
         }
     }
 }
--- a/src/qml/components/StorageSettings.qml
+++ b/src/qml/components/StorageSettings.qml
@@ -58,4 +58,17 @@ ColumnLayout {
             loadedItem.forceActiveFocus()
         }
     }
+    Separator { Layout.fillWidth: true }
+    Setting {
+        id: customDataDirSetting
+        Layout.fillWidth: true
+        header: qsTr("Data Directory")
+    }
+    CoreText {
+        Layout.fillWidth: true
+        text: optionsModel.dataDir
+        color: Theme.color.neutral7
+        font.pixelSize: 15
+        horizontalAlignment: Text.AlignLeft
+    }
 }
--- a/src/qml/controls/OptionButton.qml
+++ b/src/qml/controls/OptionButton.qml
@@ -11,6 +11,7 @@ Button {
     property string description
     property bool recommended: false
     property string image: ""
+    property string customDir: ""
     padding: 15
     checkable: true
     implicitWidth: 450
@@ -24,6 +25,12 @@ Button {
             borderRadius: 14
         }
     }
+
+    MouseArea {
+        anchors.fill: parent
+        onClicked: button.clicked()
+    }
+
     contentItem: RowLayout {
         spacing: 3
         Loader {
@@ -77,6 +84,28 @@ Button {
 
                     Behavior on color {
                         ColorAnimation { duration: 150 }
+                    }
+                }
+            }
+            Loader {
+                Layout.topMargin: 12
+                Layout.fillWidth: true
+                active: button.customDir.length > 0
+                visible: active
+                sourceComponent: Button {
+                    id: container
+                    background: Rectangle {
+                        color: Theme.color.neutral2
+                        radius: 5
+                    }
+                    font.family: "Inter"
+                    font.styleName: "Semi Bold"
+                    font.pixelSize: 13
+                    contentItem: Text {
+                        font: container.font
+                        color: Theme.color.neutral9
+                        text: button.customDir
+                        wrapMode: Text.WordWrap
                     }
                 }
             }
--- a/src/qml/models/options_model.cpp
+++ b/src/qml/models/options_model.cpp
@@ -43,6 +43,8 @@ OptionsQmlModel::OptionsQmlModel(interfaces::Node& node, bool is_onboarded)
     m_server = SettingToBool(m_node.getPersistentSetting("server"), false);
 
     m_upnp = SettingToBool(m_node.getPersistentSetting("upnp"), DEFAULT_UPNP);
+
+    m_dataDir = getDefaultDataDirString();
 }
 
 void OptionsQmlModel::setDbcacheSizeMiB(int new_dbcache_size_mib)
@@ -156,11 +158,46 @@ QUrl OptionsQmlModel::getDefaultDataDirectory()
     return QUrl::fromLocalFile(path);
 }
 
-void OptionsQmlModel::setCustomDataDirArgs(QString path)
+bool OptionsQmlModel::setCustomDataDirArgs(QString path)
 {
     if (!path.isEmpty()) {
-        // TODO: add actual custom data wiring
+    // TODO: add actual custom data wiring
+#ifdef __ANDROID__
+    QString uri = path;
+    QString originalPrefix = "content://com.android.externalstorage.documents/tree/primary%3A";
+    QString newPrefix = "/storage/self/primary/";
+    QString path = uri.replace(originalPrefix, newPrefix);
+#else
+    path = QUrl(path).toLocalFile();
+#endif // __ANDROID__
         qDebug() << "PlaceHolder: Created data directory: " << path;
+
+        m_custom_datadir_string = path;
+        Q_EMIT customDataDirStringChanged(path);
+        setDataDir(path);
+        return true;
+    }
+    return false;
+}
+
+QString OptionsQmlModel::getCustomDataDirString()
+{
+#ifdef __ANDROID__
+    m_custom_datadir_string = m_custom_datadir_string.replace("content://com.android.externalstorage.documents/tree/primary%3A", "/storage/self/primary/");
+#endif // __ANDROID__
+    return m_custom_datadir_string;
+}
+
+void OptionsQmlModel::setDataDir(QString new_data_dir)
+{
+    if (new_data_dir != m_dataDir) {
+        m_dataDir = new_data_dir;
+        if (!getCustomDataDirString().isEmpty() && (new_data_dir != getDefaultDataDirString())) {
+            m_dataDir = getCustomDataDirString();
+        } else {
+            m_dataDir = getDefaultDataDirString();
+        }
+        Q_EMIT dataDirChanged(new_data_dir);
     }
 }
 
--- a/src/qml/models/options_model.h
+++ b/src/qml/models/options_model.h
@@ -34,6 +34,7 @@ class OptionsQmlModel : public QObject
     Q_PROPERTY(int scriptThreads READ scriptThreads WRITE setScriptThreads NOTIFY scriptThreadsChanged)
     Q_PROPERTY(bool server READ server WRITE setServer NOTIFY serverChanged)
     Q_PROPERTY(bool upnp READ upnp WRITE setUpnp NOTIFY upnpChanged)
+    Q_PROPERTY(QString dataDir READ dataDir WRITE setDataDir NOTIFY dataDirChanged)
     Q_PROPERTY(QString getDefaultDataDirString READ getDefaultDataDirString CONSTANT)
     Q_PROPERTY(QUrl getDefaultDataDirectory READ getDefaultDataDirectory CONSTANT)
 
@@ -60,14 +61,16 @@ public:
     void setServer(bool new_server);
     bool upnp() const { return m_upnp; }
     void setUpnp(bool new_upnp);
+    QString dataDir() const { return m_dataDir; }
+    void setDataDir(QString new_data_dir);
     QString getDefaultDataDirString();
     QUrl getDefaultDataDirectory();
-    Q_INVOKABLE void setCustomDataDirArgs(QString path);
+    Q_INVOKABLE bool setCustomDataDirArgs(QString path);
+    Q_INVOKABLE QString getCustomDataDirString();
 
 public Q_SLOTS:
     void setCustomDataDirString(const QString &new_custom_datadir_string) {
         m_custom_datadir_string = new_custom_datadir_string;
-        m_signalReceived = true;
     }
     Q_INVOKABLE void onboard();
 
@@ -81,6 +84,7 @@ Q_SIGNALS:
     void serverChanged(bool new_server);
     void upnpChanged(bool new_upnp);
     void customDataDirStringChanged(QString new_custom_datadir_string);
+    void dataDirChanged(QString new_data_dir);
 
 private:
     interfaces::Node& m_node;
@@ -100,7 +104,7 @@ private:
     bool m_server;
     bool m_upnp;
     QString m_custom_datadir_string;
-    bool m_signalReceived = false;
+    QString m_dataDir;
 
     common::SettingsValue pruneSetting() const;
 };


src/qml/controls/OptionButton.qml
```diff
@@ -80,6 +81,28 @@ Button {
...
+                visible: active
+                sourceComponent: Button {
+                    id: container
+                    background: Rectangle {
+                        color: Theme.color.neutral2
+                        radius: 5
+                    }
+                    font.family: "Inter"
+                    font.styleName: "Semi Bold"
+                    font.pixelSize: (Qt.platform.os === "android") ? Math.max(12, Qt.screenDpi / 100) : 13
```
johnny9, line 97: What's the reason for having this different for android?  Android also scales the fonts based on the dpi already so I'm not sure I understand the dpi factor here

src/qml/models/options_model.cpp
```diff
@@ -130,10 +130,32 @@ QUrl OptionsQmlModel::getDefaultDataDirectory()
...
+
+QString OptionsQmlModel::getCustomDataDirString()
+{
+    #ifdef __ANDROID__
+        m_custom_datadir_string = m_custom_datadir_string.replace("content://com.android.externalstorage.documents/tree/primary%3A", "/storage/self/primary/");
+    #else
+        m_custom_datadir_string = QUrl(m_custom_datadir_string).toLocalFile();
+    #endif // __ANDROID__
+    return m_custom_datadir_string;
 }
```
johnny9, line 161: add a newline to the end of the file

src/qml/models/options_model.cpp
```diff
@@ -130,10 +130,32 @@ QUrl OptionsQmlModel::getDefaultDataDirectory()
...
+
+        Q_EMIT customDataDirStringChanged(path);
+        return path;
     }
+    return QString();
+}
+
+QString OptionsQmlModel::getCustomDataDirString()
+{
+    #ifdef __ANDROID__
```
johnny9, line 155: #ifdef are never indented

src/qml/models/options_model.cpp
```diff
@@ -130,10 +130,32 @@ QUrl OptionsQmlModel::getDefaultDataDirectory()
     return QUrl::fromLocalFile(path);
 }
 
-void OptionsQmlModel::setCustomDataDirArgs(QString path)
+QString OptionsQmlModel::setCustomDataDirArgs(QString path)
 {
     if (!path.isEmpty()) {
         // TODO: add actual custom data wiring
+        #ifdef __ANDROID__
```
johnny9, line 137: #ifdef shouldnt be indented

johnny9 (changes requested):
Based on both the figma and the web prototype, the path should only be displayed on the custom selection after the user has made a choice.

https://lively-kashata-cfde7e.netlify.app/screen/first-use/storage-location?state=default

src/qml/components/StorageLocations.qml
```diff
@@ -21,25 +21,28 @@ ColumnLayout {
         ButtonGroup.group: group
```
pablomartin4btc, line 21: Part of suggestion (2), assign an identifier to set `checked = true` on it when user cancels to set custom data dir.

```suggestion
        id: defaultDirOption
        ButtonGroup.group: group
```

src/qml/components/StorageLocations.qml
```diff
@@ -21,25 +21,28 @@ ColumnLayout {
...
         checked: true
     }
     OptionButton {
+        id: customDirOption
         Layout.fillWidth: true
         ButtonGroup.group: group
         text: qsTr("Custom")
         description: qsTr("Choose the directory and storage device.")
+        customDir: fileDialog.folder
         onClicked: fileDialog.open()
```
pablomartin4btc, line 35: Part of suggestion (2), if the users checks the default option, we don't show the custom data dir anymore.

```suggestion
        onClicked: fileDialog.open()
        onCheckedChanged: {
            if (!customDirOption.checked) {
                customDirOption.customDir = ""
            }
        }
```

src/qml/components/StorageLocations.qml
```diff
@@ -21,25 +21,28 @@ ColumnLayout {
...
+        folder: shortcuts.home
         onAccepted: {
             optionsModel.setCustomDataDirString(fileDialog.fileUrls[0].toString())
             var customDataDir = fileDialog.fileUrl.toString();
             if (customDataDir !== "") {
-                optionsModel.setCustomDataDirArgs(customDataDir);
+                customDirOption.customDir = optionsModel.setCustomDataDirArgs(customDataDir);
             }
         }
         onRejected: {
```
pablomartin4btc, line 48: Part of suggestion (2), when users cancels the file dialog, depending if the users changed the folder (stays in custom) or not (goes back to default).

```suggestion
        onRejected: {
            console.log("Custom datadir selection canceled")
            
            // if the user cancels to set a custom dir from the file dialog
            // default custom datadir check should be ticked
            if (fileDialog.folder == shortcuts.home) { 
                defaultDirOption.checked = true
            } else {
                // unless user accepted first another custom folder and then 
                // going back and forth from default cancels changing custom
                customDirOption.customDir = optionsModel.setCustomDataDirArgs(fileDialog.folder);
            }
```

pablomartin4btc (commented):
tACK 0025e91fe4ac53e0d8485f3dedb65e8ffff8b50c

- [Noticed](https://github.com/bitcoin-core/gui-qml/pull/397#pullrequestreview-1998557872) the same as @johnny9, proposed a change below (1) which also fixes the prefix of the `customDir` "`file://`" before the file dialog gets open or if the user cancels it.
  - <details>
    <summary>screenshot</summary>

    Screenshot from 2024-04-12 20-36-30

    </details>

- Currently when the user cancels to set a custom datadir the custom option is still checked when perhaps we should go back to the default datadir being checked. So I proposed another change to that (2) with a caveat where if the user sets the custom, then wants to change it again but cancels it, it remains there in the custom checked.

- We need to tackle the [double click issue](https://github.com/bitcoin-core/gui-qml/pull/392#pullrequestreview-1973457090) at some point. I spent some time with it but couldn't solved it yet.

- First time I run it I got the [crash reported](https://github.com/bitcoin-core/gui-qml/pull/390#issuecomment-2018656948) by @MarnixCroes on #390, and more often I get another errors related with Gtk image in the console, which I [reported](https://github.com/bitcoin-core/gui-qml/pull/390#pullrequestreview-1940546706) (3.) already in #390 as well, both errors are produced as soon as I open the file dialog.
  - <details>
    <summary>crash error</summary>
    
    `**
Gtk:ERROR:../../../../gtk/gtkpathbar.c:1342:set_button_image_get_info_cb: assertion failed: (cancellable == data->button_data->cancellable)
Bail out! Gtk:ERROR:../../../../gtk/gtkpathbar.c:1342:set_button_image_get_info_cb: assertion failed: (cancellable == data->button_data->cancellable)
Aborted (core dumped)`
    
    </details>
  - <details>
    <summary>Gtk image warning</summary>
    
    `(bitcoin-qt:1496162): Gtk-CRITICAL **: 20:32:43.950: gtk_image_set_from_gicon: assertion 'GTK_IS_IMAGE (image)' failed`
        
    </details>

D33r-Gee:
> Based on both the figma and the web prototype, the path should only be displayed on the custom selection after the user has made a choice.
> 
> https://lively-kashata-cfde7e.netlify.app/screen/first-use/storage-location?state=default

To be clear we don't want to display the default datadir path here, correct?

However do we want to display the default path in StorageSettings.qml?

pablomartin4btc:
> > Based on both the figma and the web prototype, the path should only be displayed on the custom selection after the user has made a choice.
> > https://lively-kashata-cfde7e.netlify.app/screen/first-use/storage-location?state=default
> 
> To be clear we don't want to display the default datadir path here, correct?

As of figma, and the web prototype (link above) no, but it won't hurt to show it, I'm not sure.

Also, sorry if I'm adding confusion here but on a previous PR #390 there was a "Recommended" highlighted label on the default choice [when I reviewed it](https://github.com/bitcoin-core/gui-qml/pull/390#pullrequestreview-1940546706), was that dropped (if ever discussed)?

image

 
> However do we want to display the default path in StorageSettings.qml?

I think so, whatever it is set default or custom.

cc @GBKS

D33r-Gee:
> > To be clear we don't want to display the default datadir path here, correct?
> 
> As of figma, and the web prototype (link above) no, but it won't hurt to show it, I'm not sure.

@GBKS and @johnny9 what your thoughts regarding the above?

src/qml/components/StorageLocations.qml
D33r-Gee: great suggestions... done

src/qml/components/StorageLocations.qml
```diff
@@ -21,25 +21,28 @@ ColumnLayout {
...
         recommended: true
         checked: true
     }
     OptionButton {
+        id: customDirOption
         Layout.fillWidth: true
         ButtonGroup.group: group
         text: qsTr("Custom")
         description: qsTr("Choose the directory and storage device.")
+        customDir: fileDialog.folder
```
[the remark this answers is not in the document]
D33r-Gee: done

src/qml/components/StorageLocations.qml
D33r-Gee: also great suggestion

src/qml/controls/OptionButton.qml
D33r-Gee: good call folks... done

src/qml/models/options_model.cpp
D33r-Gee: done

src/qml/models/options_model.cpp
D33r-Gee: done

src/qml/models/options_model.cpp
D33r-Gee: done

src/qml/components/StorageLocations.qml
D33r-Gee: done

johnny9:
> > > To be clear we don't want to display the default datadir path here, correct?
> > 
> > 
> > As of figma, and the web prototype (link above) no, but it won't hurt to show it, I'm not sure.
> 
> @GBKS and @johnny9 what your thoughts regarding the above?

I don't have a strong opinion about this so i would defer to the current figma.

I do think the "Recommended" highlight is redundant as "Default" has the same meaning so I would remove that to match the figma

GBKS:
Just tested on desktop (Mac) and mobile (Android).

image

I agree with Johnny's rationale about ditching the `Recommended` label.

Something I noticed is that the directory display of the custom option disappears if I select the default one again (center). It feels unexpected, could we keep that visible?

I could go either way on showing the default directory. Might be more consistent if we just show it, so users know what they are changing from.

For the rightmost screenshot, could we switch to a vertical layout for the directory field? So the directory goes underneath the label? It's a bit awkward to have it squeezed on the right side.

mouxdesign:
Storage error message: What would be the error message that a user would see if they choose to proceed but their device does not have sufficient storage?

pablomartin4btc (commented):
re tACK dfc002306b6e6445703c75a87b0725786db33133

Verified my previous [feedback](https://github.com/bitcoin-core/gui-qml/pull/397#pullrequestreview-1998733695) got addressed. Usability seems more consistent.

<ins>**Some observations**</ins> (perhaps we need to address them with a new "onboarding related" PR):
- <details><summary> Storage space estimation looks wrong during onboarding (e.g. starting the app simply with <code>bitcoin-qt</code> vs. <code>bitcoin-qt -signet</code>, please expand to see the screenshots) - Also, should we keep "Recommended" highlighted here?</summary>

  - `bitcoin-qt` ("Reduce storage" looks wrong)

    Screenshot from 2024-04-23 12-23-04
  
  - `bitcoin-qt signet` ("Store all data" looks wrong)

    Screenshot from 2024-04-23 12-24-39


</details>

- Using `-datadir` behaviour doesn't trigger the onboarding (neither combined with `-chain=` option - e.g. `-signet` with no `-datadir` triggers the onboarding when signet subdir doesn't exist on default datadir structure but setting `-datadir` option + `-signet` with no subdire doesn't) and it should when that custom datadir is empty otherwise:

  ```
  ./src/qt/bitcoin-qt -datadir="/home/pablo/custom"
  Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
  Running initialization in thread
  Warning: Disk space for "/home/pablo/custom/blocks" may not accommodate the block files. Approximately 540 GB of data will be stored in this directory.


  ```

@D33r-Gee, could you please try to investigate root of the crash mentioned at the end of my previous review? It seems it happens at the very first time you run `bitcoin-qt` (need to remove datadir and bitcoin config dir) using the file dialog of course, otherwise it doesn't happen. Most of the time the user would get this error message in the terminal:

`(bitcoin-qt:756208): Gtk-CRITICAL **: 10:57:38.210: gtk_image_set_from_gicon: assertion 'GTK_IS_IMAGE (image)' failed
`

pablomartin4btc:
> Just tested on desktop (Mac) and mobile (Android).

@GBKS for future reference, please add the commit version you've tested if possible so we can verify if it's the latest or a previous one.

> I agree with Johnny's rationale about ditching the `Recommended` label.

Ok, please check the next page when user has to select the amount of space to store, there's also a "Recommended" highlighted label.
 
> Something I noticed is that the directory display of the custom option disappears if I select the default one again (center). It feels unexpected, could we keep that visible?

I've proposed to hide it when a user goes back to default datadir to avoid confusion, as in when a user sees the page for the first time won't see the custom datadir until the user clicks on it and set it. Is it necessary if that's not the datadir that's going to be used?

> I could go either way on showing the default directory. Might be more consistent if we just show it, so users know what they are changing from.

I'd show it but I've also thought if we can add an external link, not sure if with an information icon or something, to the bitcoin datadir [documentation](https://github.com/bitcoin/bitcoin/blob/master/doc/files.md#data-directory-location).

pablomartin4btc:
> Storage error message: What would be the error message that a user would see if they choose to proceed but their device does not have sufficient storage?

I think that should be after accepting (clicking on next) the following page "Reduce store"/ "Store all data" (before starting the initial download sync).

GBKS:
How about one of these two options for the error message?

image

GBKS:
Can we detect how much space is available on each location and show it? And can we estimate the required space needed (the mock-up uses 500GB as a placeholder, which is meant to be replaced with something accurate).

D33r-Gee:
Rebased over main [dfc0023 ](https://github.com/bitcoin-core/gui-qml/commit/dfc002306b6e6445703c75a87b0725786db33133)to [2fbb27a](https://github.com/bitcoin-core/gui-qml/commit/2fbb27a37a7c666f49a25465863dcafd9d0713f5)

D33r-Gee:
@GBKS and @pablomartin4btc Thanks for the feedback regarding the storage amount and the error messaging.
 those will be addressed in a different PR since this is only for datadir display
 
 Meanwhile please me know your thoughts on the latest update update [8e8aadf](https://github.com/bitcoin-core/gui-qml/commit/8e8aadf2396e54428c0ca5ae562030734d0650cf)?

MarnixCroes (commented):
8e8aadf2396e54428c0ca5ae562030734d0650cf
I cannot repro the double-click issue anymore https://github.com/bitcoin-core/gui-qml/pull/399 
friendly ping @pablomartin4btc

---

one issue/inconsistency: 
the selected custom data dir is not updated/displayed correctly during the onboarding, but it as at settings:
onboarding -> select custom data dir -> next -> click `Detailed settings` -> default data dir is displayed -> next -> next -> arrive at home screen -> go to `settings` -> `storage` ->  data directory displays the custom data dir

D33r-Gee:
> [8e8aadf](https://github.com/bitcoin-core/gui-qml/commit/8e8aadf2396e54428c0ca5ae562030734d0650cf) I cannot repro the double-click issue anymore #399 friendly ping @pablomartin4btc
> 
That's great to hear...
> one issue/inconsistency: the selected custom data dir is not updated/displayed correctly during the onboarding, but it as at settings: onboarding -> select custom data dir -> next -> click `Detailed settings` -> default data dir is displayed -> next -> next -> arrive at home screen -> go to `settings` -> `storage` -> data directory displays the custom data dir
Could you post a screenshot please?

pablomartin4btc (commented):
> [8e8aadf](https://github.com/bitcoin-core/gui-qml/commit/8e8aadf2396e54428c0ca5ae562030734d0650cf) I cannot repro the double-click issue anymore #399 friendly ping @pablomartin4btc

Ok, thanks for letting me know... I'll do another test on the latest version of this PR.

MarnixCroes:
> Could you post a screenshot please?

during onboarding:

Screenshot from 2024-05-01 23-45-12

Screenshot from 2024-05-01 23-45-19

after onboarding, at settings
Screenshot from 2024-05-01 23-45-28

D33r-Gee:
@MarnixCroes Thanks for the screenshots... I see what you mean... Looking into it

pablomartin4btc (commented):
tACK 8e8aadf2396e54428c0ca5ae562030734d0650cf

I'm not sure about showing the default path when the user has selected the custom one but not a blocker.

> [8e8aadf](https://github.com/bitcoin-core/gui-qml/commit/8e8aadf2396e54428c0ca5ae562030734d0650cf)
I cannot repro the double-click issue anymore #399
friendly ping @pablomartin4btc

<details>
<summary>I can still do it (being on custom data dir selected, just double clicking on it).</summary>
 
image

</details>

> one issue/inconsistency:
the selected custom data dir is not updated/displayed correctly during the onboarding, but it as at settings:
onboarding -> select custom data dir -> next -> click Detailed settings -> default data dir is displayed -> next -> next -> arrive at home screen -> go to settings -> storage -> data directory displays the custom data dir

I could repro this ^; it looks like the model hasn't been refreshed? Are we sure it's the same instance of the options model (not a copy)? cc @johnny9.

D33r-Gee:
Thanks @pablomartin4btc for testing it

> I'm not sure about showing the default path when the user has selected the custom one but not a blocker.

I could go either way, however it feels consistent to show the default datadir regardless since if the user is going through onboarding there's a high chance it might be their first time installing Bitcoin Core, and knowing where the default data is store might be helpful. Any thoughts @GBKS?

RE: @MarnixCroes comment about displaying the default datadir instead of custom, it seems StorageSettings is not refreshing until the node launches so it's not registering the change from default to custom datadir, I believe that this get addressed with #284. Going to test to confirm on  separate branch...

D33r-Gee:
with the latest commit [2a301e1](https://github.com/bitcoin-core/gui-qml/commit/2a301e117f6e83611b08a3d6242ed6aaa1f4caa5) addressed @MarnixCroes comment about the datadir not displaying properly during onboarding. Turns out made the mistake of not using the setter function for the customDataDir bool.

Please let me know if it displays correctly for ya'll @pablomartin4btc @johnny9 @GBKS @MarnixCroes ?

src/qml/models/options_model.h
```diff
@@ -62,7 +63,10 @@ class OptionsQmlModel : public QObject
     void setUpnp(bool new_upnp);
     QString getDefaultDataDirString();
     QUrl getDefaultDataDirectory();
-    Q_INVOKABLE void setCustomDataDirArgs(QString path);
+    Q_INVOKABLE bool setCustomDataDirArgs(QString path);
+    Q_INVOKABLE QString getCustomDataDirString();
+    bool customDataDir() const { return m_custom_datadir; }
+    void setCustomDataDir(bool new_custom_datadir);
```
pablomartin4btc, line 69: It seems confusing the current naming convention here, also: do we need to store a bool?; perhaps @johnny9 has a better naming since he has been working with the options model already, mine is just an idea.

```suggestion
    bool isCustomDataDirSet() const { return (!getCustomDataDirString().isEmpty() && (getDefaultDataDirString() != getCustomDataDirString())); }
```

pablomartin4btc (commented):
tACK 2a301e117f6e83611b08a3d6242ed6aaa1f4caa5

<details>
<summary>The issue/ inconsistency <a href="https://github.com/bitcoin-core/gui-qml/pull/397#pullrequestreview-2034467887">reported</a> by @MarnixCroes has been fixed.</summary>

Screenshot from 2024-05-06 18-42-42

</details>

<details>
<summary>I noticed another bug which is: if the user clicks on custom, the file dialogs opens, the user creates a folder or moves around dirs but cancels the dialog, the custom option will be kept ticked/ selected but with no path and if the user continues the default will be stored.</summary>

image

</details>

On the default path shown when custom is selected:
> it feels consistent to show the default datadir regardless

I'm still not sure because when the user lands into this page, no custom data dir has been selected so nothing is shown. When the 2 paths are shown someone could think that some data is still stored on the default path? Again, not a blocker.

src/qml/models/options_model.h
D33r-Gee: yes a bool is needed here so that StorageSettings.qml relfects the proper data directory (either default or custom).

D33r-Gee:
Thanks @pablomartin4btc for testing and reviewing

> I noticed another bug which is: if the user clicks on custom...

addressed this with the latest push from [2a301e1 ](https://github.com/bitcoin-core/gui-qml/commit/2a301e117f6e83611b08a3d6242ed6aaa1f4caa5)to [61e775b](https://github.com/bitcoin-core/gui-qml/commit/61e775b0d7f18328e6f04e6a99cb70017fe00539)

Now when a user cancels after having picked a custom datadir it reverts back to default

src/qml/models/options_model.h
pablomartin4btc: ok but do we need that instance variable (`m_custom_datadir`) to store it or it can be dynamically calculated as suggested?

GBKS:
Just tested the [latest build](https://github.com/bitcoin-core/gui-qml/actions/runs/8989560816), which I believe is 61e775b , on MacOS and Android.

image

Nice to see this come together. Just have some small comments:

- Tapping anywhere in the directory option should activate it. This does not seem to work when tapping the light grey directory display box inside the item. Does it maybe intercept the click/tap event and prevent it from propagating to its parent?
- On the settings page, the directory name seems to have the wrong color and too much spacing to the option title above it 

Regarding the point about consistency for showing the directory box, I'd show it if it has a value. The default one can always be visible. And if the user has set one for the custom option, it can also be shown independent of whether that option is still selected or not. And if the user goes to the directory picker and cancels out of it, the previously picked option can remain active. That seems intuitive to me.

Quite the long thread, hope I didn't miss anything. Maybe good to address things like the error message in a separate PR to keep things manageable.

src/qml/models/options_model.h
D33r-Gee: Certainly, adhering to the established conventions for options would suggest utilizing m_custom_datadir. However, if a more refined approach emerges, we could potentially explore its implementation in a subsequent pull request.

D33r-Gee:
Thanks @GBKS for testing and reviewing...

This:
> Tapping anywhere in the directory option should activate it. This does not seem to work when tapping the light grey directory display box inside the item. Does it maybe intercept the click/tap event and prevent it from propagating to its parent?

Is addressed with the latest push [b941932](https://github.com/bitcoin-core/gui-qml/commit/b941932926a55144d78e2ae07b721cf38ad48639)

> On the settings page, the directory name seems to have the wrong color

what would be the correct color?

> And if the user goes to the directory picker and cancels out of it, the previously picked option can remain active. That seems intuitive to me.

this use case may be problematic since if the user cancels it currently reverts back to default, are you suggesting that if there was a previously selected folder then the custom option should remain selected even after the user cancels a secondary attempt at setting the custom datadir?

D33r-Gee:
with the latest push from [b941932 ](https://github.com/bitcoin-core/gui-qml/commit/b941932926a55144d78e2ae07b721cf38ad48639)to [0c46cae](https://github.com/bitcoin-core/gui-qml/commit/0c46caeac5c9f9166ae8891987f1f9bf626dbd4d) addressed @GBKS comment regarding the proper color for datadir display

Next is addressing this:

> And if the user goes to the directory picker and cancels out of it, the previously picked option can remain active. That seems intuitive to me.

D33r-Gee:
new commit [855cea5](https://github.com/bitcoin-core/gui-qml/commit/855cea5fd27d227ce595cb06a3fd281438a6d0b6) addressing this use case 

> And if the user goes to the directory picker and cancels out of it, the previously picked option can remain active. That seems intuitive to me.

Now when the user has picked/accepted a custom datadir and then decides to change it but cancels the selection it stays with the custom datadir previously selected unless they click on the default option

D33r-Gee:
rebased new head is ce6866a9adf408625f6506497f9e78b6b663cadf

src/qml/models/options_model.h
johnny9: The confusion here is around the name of the variable. customDataDir implies it is a string. A better name for a bool would be isCustomDataDirSet or usingCustomDataDir and customDataDir should just be the name for the string.

johnny9: I prefer the dynamic calculation of the state myself. It is clearer to me

src/qml/models/options_model.cpp
```diff
@@ -156,14 +158,45 @@ QUrl OptionsQmlModel::getDefaultDataDirectory()
...
 }
 
-void OptionsQmlModel::setCustomDataDirArgs(QString path)
+bool OptionsQmlModel::setCustomDataDirArgs(QString path)
 {
     if (!path.isEmpty()) {
-        // TODO: add actual custom data wiring
+// TODO: add actual custom data wiring
+#ifdef __ANDROID__
+    QString uri = path;
```
johnny9, line 166: spacing is off here

src/qml/components/StorageSettings.qml
```diff
@@ -58,4 +58,21 @@ ColumnLayout {
             loadedItem.forceActiveFocus()
         }
     }
+    Separator { Layout.fillWidth: true }
+    Setting {
+        id: customDataDirSetting
+        Layout.fillWidth: true
+        header: qsTr("Data Directory")
+    }
+    ValueInput {
```
johnny9, line 67: This isnt an Input so just use CoreText

src/qml/controls/ValueInput.qml
```diff
@@ -36,6 +36,14 @@ TextInput {
                 enabled: false
                 textColor: Theme.color.neutral4
             }
+        },
+        State {
```
johnny9, line 40: Remove this state as its not appropriate

johnny9 (commented):
Don't overload ValueInput for non input items. Just use CoreText or Label

src/qml/controls/ValueInput.qml
D33r-Gee: done

src/qml/models/options_model.cpp
D33r-Gee: done

src/qml/models/options_model.h
D33r-Gee: Thanks for the suggestions, decided to go with a dynamic string variable instead of managing the customDataDir bool... Let me know your thoughts?

src/qml/components/StorageSettings.qml
D33r-Gee: done

D33r-Gee:
Updated from [ce6866a ](https://github.com/bitcoin-core/gui-qml/commit/ce6866a9adf408625f6506497f9e78b6b663cadf)to [2e2fe79](https://github.com/bitcoin-core/gui-qml/commit/2e2fe79b5af3262fca050d98fbbb639ccb0d6343)

Main changes were implementing @johnny9 and @pablomartin4btc feedback...

RE: `customDataDir` bool, decided to go with a dynamic setting of `dataDir` since the bool options was giving me grief...

Please let me know your thoughts?

johnny9 (approved):
ACK 2e2fe79b5af3262fca050d98fbbb639ccb0d6343

johnny9 (changes requested):
Seems to be an issue with building the Android target

"qml/models/options_model.cpp:186:55: error: no matching member function for call to 'replace'
    m_custom_datadir_string = m_custom_datadir_string.replace("content://com.android.externalstorage.documents/tree/primary%3A", "/storage/self/primary/");"

D33r-Gee:
> Seems to be an issue with building the Android target
> 
> "qml/models/options_model.cpp:186:55: error: no matching member function for call to 'replace' m_custom_datadir_string = m_custom_datadir_string.replace("content://com.android.externalstorage.documents/tree/primary%3A", "/storage/self/primary/");"

Thanks for discovering this... Found the cause of the problem, I mistakenly left the added `const` to the getter function. I removed them and now it builds properly for both Android and Linux

johnny9 (approved):
ACK 7341bdf839040bba2895290f3a59adc7a48132c2

MarnixCroes (approved):
tested lgtm 7341bdf839040bba2895290f3a59adc7a48132c2

one nit/UX thing: select/have a custom data dir -> select default data dir -> select custom data dir -> it prompts the folder location selection again.
for me it would be more intuitive to not prompt the storage location for one click -> only when double-clicking. when clicking on the custom datadir again

MarnixCroes:
FTR, can repro the double-click in background issue on latest commit. not sure why previously I couldn't repro https://github.com/bitcoin-core/gui-qml/pull/397#pullrequestreview-2034467887

GBKS:
The MacOS workflow runs seem to fail for this PR (also for 401). Would it be possible to fix those to simplify testing?

D33r-Gee:
> one nit/UX thing: select/have a custom data dir -> select default data dir -> select custom data dir -> it prompts the folder location selection again. for me it would be more intuitive to not prompt the storage location for one click -> only when double-clicking. when clicking on the custom datadir again

Thanks for testing and the feedback. That's a good point, will add it to the polishing features alongside the error messaging. One thing to keep in mind is that on mobile "double-tapping" is uncommon so we'll have to play with it to see what feels most natural.

pablomartin4btc (approved):
tACK 7341bdf839040bba2895290f3a59adc7a48132c2

Another <ins>UX nit</ins>: when a user selects a custom datadir and accepts it, clicks again on custom datadir to open the file dialog, creates a different custom dir but cancels the dialog, it goes wrongly to tick/ check the default datadir. Perhaps we can correct that on a follow-up.


Standing issues to follow-up:
- Custom Datadir double click (check workaround at #399).
- custom data dir -> select default data dir -> select custom data dir -> it prompts the folder location selection again ([reported](https://github.com/bitcoin-core/gui-qml/pull/397#pullrequestreview-2070826537) by @MarnixCroes)
  - not sure about differentiation between 1 click (for just tick the custom dir) or double to open a file dialog, perhaps adding a button/ icon to open the dialog but would break this design, I'd leave this to the design experts.


I still get this error the first time I run QT and open the File Dialog... could be something related with my initial env, after the first time it's not reproducible, leaving the comment here to follow-up later.

pablomartin4btc:
> Regarding the point about consistency for showing the directory box, I'd show it if it has a value. The default one can always be visible. And if the user has set one for the custom option, it can also be shown independent of whether that option is still selected or not. And if the user goes to the directory picker and cancels out of it, the previously picked option can remain active. That seems intuitive to me.

@GBKS: Agreed. Just found another [case](https://github.com/bitcoin-core/gui-qml/pull/397#pullrequestreview-2074161733) (ref.: "Another <ins>UX nit</ins>").

D33r-Gee:
[c2e73b2](https://github.com/bitcoin-core/gui-qml/commit/c2e73b259802c5644859542c785eb46f93911841) rebased over main

src/qml/models/options_model.cpp
```diff
@@ -156,11 +158,46 @@ QUrl OptionsQmlModel::getDefaultDataDirectory()
...
+    m_custom_datadir_string = m_custom_datadir_string.replace("content://com.android.externalstorage.documents/tree/primary%3A", "/storage/self/primary/");
+#endif // __ANDROID__
+    return m_custom_datadir_string;
+}
+
+void OptionsQmlModel::setDataDir(QString new_data_dir)
+{
+    if (new_data_dir != m_dataDir) {
+        m_dataDir = new_data_dir;
+        if (!getCustomDataDirString().isEmpty() && (new_data_dir != getDefaultDataDirString())) {
```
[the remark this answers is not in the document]
D33r-Gee: Thanks for your suggestion... after testing it became clear that not depending on a new `bool` would be the way to go. And so implemented yours and Johnny9 suggestions as part of a new `dataDir` variable.

src/qml/models/options_model.cpp
```diff
@@ -156,11 +158,46 @@ QUrl OptionsQmlModel::getDefaultDataDirectory()
     return QUrl::fromLocalFile(path);
 }
 
-void OptionsQmlModel::setCustomDataDirArgs(QString path)
+bool OptionsQmlModel::setCustomDataDirArgs(QString path)
```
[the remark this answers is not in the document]
D33r-Gee: The `Args` suffix will come in play when the actual backend is connected since it will modify the `gArgs` so that the new `datadir` is set properly.

MarnixCroes (commented):
c2e73b259802c5644859542c785eb46f93911841

~it can happen that the GUI completely freezes~
~can't do anything, need to close the app.~
~can repro most of the time, but not 100%~

~maybe a bit of an edge case, as it's not the usual user flow for how to repro~

<details>
<summary>repro steps</summary>
<br>

- run with resetguisettings
- at Storage location dialog click `custom`
- click back (so don't select a location) (can also repro when selecting the location)
- click on default
- then double-click on `custom`

</details>

[qmlguifreeze.webm](https://github.com/bitcoin-core/gui-qml/assets/93143998/379343bb-b3c5-49fc-ad24-922cf2b9a626)

D33r-Gee:
> it can happen that the GUI completely freezes can't do anything, need to close the app. can repro most of the time, but not 100%
> 
> maybe a bit of an edge case, as it's not the usual user flow for how to repro
> 
Couldn't repro on WSL Ubuntu 22.04. Kept on double/triple clicking and nothing froze... 

However was able to approximately repro the "double click" (#399) from my experience the first click opens the dialog and if it opens away from the `custom` area then the second click  will push the dialog window behind the app window which I believe is proper file dialog behavior.

MarnixCroes:
ok please ignore my previous comment about freezing.
This time I was running in full window, and the file dialog was opened in background without me noticing.
sry

MarnixCroes (approved):
tACK 7a8fb19db1db138e50d7762aef2658737a3d6f8c on Ubuntu desktop

the double-click issue can be fixed later imo

pablomartin4btc (commented):
reACK 7a8fb19

Verified that the following issue has been fixed:
> Another UX nit: when a user selects a custom datadir and accepts it, clicks again on custom datadir to open the file dialog, creates a different custom dir but cancels the dialog, it goes wrongly to tick/ check the default datadir. Perhaps we can correct that on a follow-up.

D33r-Gee:
Looks like the mac os artifact issue has been addressed... Could you please confirm on your end when you get the chance, @GBKS?

GBKS:
ACK 7a8fb19. Looks and feels great, just tested again on MacOS.

Regarding the follow-ups in the description about double-clicking to save the path - why would users want to save the path in this specific situation?

D33r-Gee:
> Regarding the follow-ups in the description about double-clicking to save the path - why would users want to save the path in this specific situation?

It was an idea in case a user wants to quickly copy the `datadir` so they can either navigate to it in the cli or browse the folder using the OS gui

GBKS:
> It was an idea in case a user wants to quickly copy the datadir so they can either navigate to it in the cli or browse the folder using the OS gui

I like this thinking about secondary actions and I think we'll have a lot of opportunity to implement them throughout the app for more regular and power users (via tap-and-hold and right-click menus on mobile and desktop respectively). They are not always easy to discover, but repeat users should appreciate them.

PR #403, opened 2024-05-28
Introduce Create Single-Sig wallet flow

johnny9:
These changes add more pages after the initial onboarding.

To test, run the application with the `-resetguisettings` option.

Link to github actions build artifacts.

[Build Artifacts](https://github.com/bitcoin-core/gui-qml/actions/runs/9263281845)


src/qml/pages/wallet/CreateIntro.qml
```diff
@@ -0,0 +1,114 @@
...
+                width: 30
+                height: width
+                anchors.centerIn: circle
+            }
+        }
+
+        CoreText {
+            Layout.topMargin: 25
+            Layout.alignment: Qt.AlignCenter
+            text: qsTr("You are about to create \na single-key bitcoin wallet")
```
MarnixCroes, line 56:
```suggestion
            text: qsTr("You are about to create \na single-sig bitcoin wallet")
```

MarnixCroes (commented):
- every entrybox should automatically be selected, so user doesn't have to move the cursor there and click it to enter something
- at Create password, good to give feedback when user can't continue because passwords don't match
-  `CreateConfirm` dialog shouldn't have a back button imo, the wallet is already created so going back doesn't make sense, same for `CreateBackup`

---

some other things, but maybe for a follow-up:
- check and don't allow to enter an already existing wallet name
- at every entry it is intuitive to press enter to trigger the continue button, maybe good time to introduce the behavior that enter keystroke triggers it
- the transition to the wallet home page after clicking done at CreateConfirm is in the opposite direction of the rest of the create wallet flow, why?

src/qml/pages/wallet/CreateIntro.qml
pablomartin4btc: It says "single-key" also in [figma](https://www.figma.com/design/ek8w3n3upbluw5UL2lGhRx/Bitcoin-Core-App-Design?node-id=7516-13170).

pablomartin4btc: It says "single-key" also in [figma](https://www.figma.com/design/ek8w3n3upbluw5UL2lGhRx/Bitcoin-Core-App-Design?node-id=7516-13170).

pablomartin4btc (commented):
tACK 4c047e0b977e1c104777467c72ec8cde347b5469

It creates the wallet successfully, with or without password (skip it when it's offered).

I agree with all [observations](https://github.com/bitcoin-core/gui-qml/pull/403#pullrequestreview-2082083987) from @MarnixCroes as, tested on desktop platform, I experienced the same issues.
 
> some other things, but maybe for a follow-up:
> 
>     * check and don't allow to enter an already existing wallet name

This is what I first tested, when user enters the same wallet name doesn't do anything/ warn about the situation (doesn't override existent ones at least 🙏🏼) and it goes to the `CreateConfirm` saying it was created.


>     * the transition to the wallet home page after clicking done at CreateConfirm is in the opposite direction of the rest of the create wallet flow, why?

I'd say opposite direction to the rest of the **onboarding flow** even, so create wallet flow follows the same pattern (screen passed to the left as we move forward to the next form/ screen to the right) until the user clicks on "Done" or even if the user "skip"s the wallet creation. Unless it's intentional, it seems there's an inconsistency here.

On `CreateBackup.qml`, there's a "View file" button (that's doesn't work yet?), is this the backup feature? If so I see in [figma](https://www.figma.com/design/ek8w3n3upbluw5UL2lGhRx/Bitcoin-Core-App-Design?node-id=7516-13170) says "Save wallet file" on that button, otherwise other than "CreateBackup" I'd name the `.qml` file as "BackupIntro" o "BackupWarning"/ Info.


I see other discrepancies (missing options/ different wording) against [figma](https://www.figma.com/design/ek8w3n3upbluw5UL2lGhRx/Bitcoin-Core-App-Design?node-id=7516-13170), is this intentional?

src/qml/pages/wallet/CreateIntro.qml
MarnixCroes: yeah, "single-key" is wrong and also confusing imo

D33r-Gee: I concur "single-sig" is more intuitive than "single-key". It seems "single-key" is referenced in the docs although most people/users only refer to "single-sig"

D33r-Gee (commented):
tACK 4c047e0b977e1c104777467c72ec8cde347b5469 on WSL Ubuntu 22.04. 

Wallet creation worked.  In general agree with Marnix's comments.

And to be clear is this the first iteration, then layering more functionality (i.e. error messages, wallet name collision and more specific passwords hints )?

johnny9:
> tACK [4c047e0](https://github.com/bitcoin-core/gui-qml/commit/4c047e0b977e1c104777467c72ec8cde347b5469) on WSL Ubuntu 22.04.
> 
> Wallet creation worked. In general agree with Marnix's comments.
> 
> And to be clear is this the first iteration, then layering more functionality (i.e. error messages, wallet name collision and more specific passwords hints )?

Yeah this is just following what was ready in the Figma. I think building it helps identify all of the gaps. I think all of these comments are great and I'll do my best to get these in on this first change-set.

src/qt/bitcoin.cpp
```diff
@@ -622,7 +622,7 @@ int GuiMain(int argc, char* argv[])
     qApp->installNativeEventFilter(new WinShutdownMonitor());
 #endif
     // Install qDebug() message handler to route to debug.log
-    qInstallMessageHandler(DebugMessageHandler);
+    //qInstallMessageHandler(DebugMessageHandler);
```
hebasto, line 625: Why is that commented out?

hebasto:
cc @GBKS

src/qml/pages/wallet/CreateIntro.qml
GBKS: We generally use multi-key in the design community ([example](https://bitcoin.design/guide/how-it-works/private-key-management/multi-key/)). Some reasons:

- "sig" is an abbreviation that many people may not be familiar with
- Avoiding "signature" altogether in the applications removes another technical concept that people don't really need to be aware of. For example, "sign transaction" can be just be the more common term "approve transaction". Just part of the overall idea to use fewer technical terms if possible
- If we want to be super specific in technical terms, then a "single-sig" wallet could even be a "multi-key" wallet. For example, in a wallet with 3 keys that only requires 1 signature

I understand that the technical terms can be more intuitive for us who are super deep into the tech, and they allow us to discuss things more precisely. For broader audiences, especially global ones where english may be the second language, it's often best to simplify things a little (you even have tools for that like [Hemingway](https://hemingwayapp.com)). Does that sound reasonable?

D33r-Gee:
tACK [a3de255](https://github.com/bitcoin-core/gui-qml/commit/a3de25576f1ae86c60176885c0466e081059161b) on WSL Ubuntu 22.04

Works as expected, was able to create a new wallet in the default `datadir`

PR #410, opened 2024-08-05
create wallet: set focus on TextField so it is automatically selected

MarnixCroes:
add `focus: true` to make it automatically selected so user doesn't have to first manually select it before it can type something

addresses https://github.com/bitcoin-core/gui-qml/pull/403#pullrequestreview-2082083987


johnny9 (approved):
tACK 0beac80087f6143e9ef846026f2eb8aa22c1852c

Works as expected

PR #414, opened 2024-08-08
Uncomment DebugMessageHandler install

johnny9:
DebugMessageHandler is needed to route output to debug.log and should not have been commented out.

Mistakenly commented out in https://github.com/bitcoin-core/gui-qml/pull/403


MarnixCroes (approved):
ACK 456d28a63992aa3916c7c2d30aeb0050db7ccd7c

hebasto:
> Mistakenly commented out in #403

From https://github.com/bitcoin-core/gui-qml/pull/403#discussion_r1631956804:

> Why is that commented out?

PR #418, opened 2024-08-24
Introduce AddWalletButton and connect it to the AddWallet flow

johnny9:
These commits enable the "Add Wallet" button at the bottom of the Wallet Select menu on Desktop

### DIFF
--- a/src/qml/pages/main.qml
+++ b/src/qml/pages/main.qml
@@ -80,13 +80,17 @@ ApplicationWindow {
 
     Component {
         id: desktopWallets
-        DesktopWallets {}
+        DesktopWallets {
+            onAddWallet: {
+                main.push(createWalletWizard)
+            }
+        }
     }
 
     Component {
         id: createWalletWizard
         CreateWalletWizard {
-            onFinished: {
+           onFinished: {
                 main.pop()
             }
         }
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -382,6 +382,7 @@ QML_RES_QML = \
   qml/components/ThemeSettings.qml \
   qml/components/TotalBytesIndicator.qml \
   qml/components/Tooltip.qml \
+  qml/controls/AddWalletButton.qml \
   qml/controls/CaretRightIcon.qml \
   qml/controls/ContinueButton.qml \
   qml/controls/CoreText.qml \
--- a/src/qml/bitcoin_qml.qrc
+++ b/src/qml/bitcoin_qml.qrc
@@ -20,6 +20,7 @@
         <file>components/ThemeSettings.qml</file>
         <file>components/TotalBytesIndicator.qml</file>
         <file>components/Tooltip.qml</file>
+        <file>controls/AddWalletButton.qml</file>
         <file>controls/ContinueButton.qml</file>
         <file>controls/CoreText.qml</file>
         <file>controls/CoreTextField.qml</file>
--- a/src/qml/controls/AddWalletButton.qml
+++ b/src/qml/controls/AddWalletButton.qml
@@ -1,0 +1,83 @@
+// Copyright (c) 2024 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+
+import org.bitcoincore.qt 1.0
+
+Button {
+    id: root
+
+    property color bgActiveColor: Theme.color.neutral2
+    property color textColor: Theme.color.neutral7
+    property color textHoverColor: Theme.color.orange
+    property color textActiveColor: Theme.color.orange
+
+    hoverEnabled: AppMode.isDesktop
+    implicitHeight: 30
+    implicitWidth: 220
+
+    MouseArea {
+        anchors.fill: parent
+        enabled: false
+        hoverEnabled: true
+        cursorShape: Qt.PointingHandCursor
+    }
+
+    contentItem: Item {
+        anchors.fill: parent
+        RowLayout {
+            anchors.centerIn: parent
+            spacing: 3
+            Icon {
+                id: addIcon
+                Layout.alignment: Qt.AlignRight
+                source: "image://images/plus"
+                color: Theme.color.neutral7
+                size: 16
+                Layout.minimumWidth: 16
+                Layout.preferredWidth: 16
+                Layout.maximumWidth: 16
+            }
+            CoreText {
+                id: addText
+                Layout.fillHeight: true
+                Layout.alignment: Qt.AlignLeft
+                text: qsTr("Add Wallet")
+                color: Theme.color.neutral7
+                font.pixelSize: 15
+            }
+        }
+    }
+
+    background: Rectangle {
+        id: bg
+        height: 30
+        width: 220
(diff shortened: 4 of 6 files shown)


MarnixCroes (commented):
4b79351937f0228504ff2c3dfdbb20d1a1915059
the space between the icon and the text is different than from master

<details>
<summary>Master</summary>
<br>

Screenshot from 2024-08-24 11-06-36

</details>

<details>
<summary>PR</summary>
<br>

Screenshot from 2024-08-24 11-06-48

</details>

---

maybe for follow up:
At the AddWallet page, the `Skip` button in in the right top should get a condition to only have it when in the onboarding flow. Now in the app when you click add wallet button -> Add Wallet page has a skip button, which makes no sense. a cancel would be more appropriate

hebasto:
cc @GBKS

GBKS:
Nice work adding this functionality. I did a review of the full flow (MacOS) and added my notes in the visual below. Some of them will not be relevant to the focus of this PR and should probably be better handled in a separate one. Let me know how you want to handle that.

Bitcoin Core App PR 418 review 240911

And in written form:

**1. Wallet selector dropdown**
1.1 All font sizes should be 15px (not 13, 14, and 15)
1.2 Reduce spacing between "+" icon and label in the "+ Add Wallet" button

**2. Add a wallet screen**
2.1 In the transition to this screen, the new screen should slide in from below (not from the right)
2.2 Replace "Skip" button in the top-right, but a "Cancel" button in the top-left (keep "Skip" on this screen when it is shown in the onboarding sequence)

**3. Choose a wallet name screen**
3.1 Replace "Continue" button label with "Next"

**4. Your wallet has been created screen**
4.1 Do we want to offer a "Back" button here? It is convenient for users, yet also creates a new wallet every time they go back and forth - is that a problem?

**5. Back up your wallet screen**
5.1 Same question about the "Back" button as above
5.2 "View file" button label should be white
5.3 Reduce spacing between buttons to 20px
5.4 When exiting this screen, it should slide down

**6. Activity screen post wallet creation**  
6.1 The newly created wallet should now be active

rabbitholiness:
On macOS, the wallet name screen would not allow me to add spaces to the wallet name
Screenshot 2024-09-18 at 10 38 10

pablomartin4btc (approved):
tACK 694025a8b4d202a6edf91d59e9c3c9d0c24314ab

@rabbitholiness, this is only the UI link to the existent flow (`CreateName.qml` - not included in this PR), where there's a validator (dev's note: regexp qt class is deprecated and needs to be updated) which needs to be modified to include spaces as you said and I tested it too. Also to that flow we need to add functionality to the "Import wallet" and "View file" buttons, make the encryption not mandatory and perhaps as in Qt switch to the recently created wallet.

MarnixCroes (approved):
tACK 694025a8b4d202a6edf91d59e9c3c9d0c24314ab

johnny9:
Previous:
https://github.com/user-attachments/assets/ed5e64af-1f4b-494c-a6fe-cf767ab39066

Updated:
https://github.com/user-attachments/assets/8b16dac8-466b-405f-a438-7cca5a7484c8

Web prototype:
https://github.com/user-attachments/assets/af4de29d-3a72-472d-8e59-bfd819df7b70

GBKS:
Nice one. Looks like there's an issue with the web prototype in that the menu does not disappear when entering the create wallet flow. I'll make a note of that.

Something I noticed is that clicking the balance always toggle the menu to be visible. But when it is visible, and I click the balance again, the menu should disappear. Not having looked at the code at all, but knowing similar setups from building web apps, my assumption is that there is the click-outside-of-the-menu event handles is the reason for this. The mouse-down on the balance is counted as a click outside of the menu, which causes it to close. Then right after the click on the balance is triggered. Since the menu is now closed, it shows the menu again. So that mouse-down on the balance should not trigger the close (if that makes sense).

Another tweak (outside of this PR probably) is to show a "Cancel" button in the top-left of the first screen of the create wallet flow. That flow can be entered in two ways, with each one requiring different top bar options, so they make sense in context:
- First-use experience: "Skip" button on the right
- In-app wallet creation: "Cancel" button on the left

johnny9:
> Something I noticed is that clicking the balance always toggle the menu to be visible. But when it is visible, and I click the balance again, the menu should disappear. Not having looked at the code at all, but knowing similar setups from building web apps, my assumption is that there is the click-outside-of-the-menu event handles is the reason for this. The mouse-down on the balance is counted as a click outside of the menu, which causes it to close. Then right after the click on the balance is triggered. Since the menu is now closed, it shows the menu again. So that mouse-down on the balance should not trigger the close (if that makes sense).

You are spot on with this issue. I tried pretty hard to get this behavior fixed but i didn't find a good solution with the context we have. Hopefully a fresh look at it soon and I'll be able to solve it.

> Another tweak (outside of this PR probably) is to show a "Cancel" button in the top-left of the first screen of the create wallet flow. That flow can be entered in two ways, with each one requiring different top bar options, so they make sense in context:

This is a good idea and I will give myself the task of implementing this

johnny9:
Update from [2f04d12](https://github.com/bitcoin-core/gui-qml/commit/2f04d1272467b60b948e4e0f504ee759a440b999) to [482c1f0](https://github.com/bitcoin-core/gui-qml/commit/482c1f02fea3a58e29566597302ed114da9d19b8)

- rebased with main

johnny9:
Update from [482c1f0](https://github.com/bitcoin-core/gui-qml/commit/482c1f02fea3a58e29566597302ed114da9d19b8) to [be397b6](https://github.com/bitcoin-core/gui-qml/commit/be397b6ddc034c3cd29cb3c9d40e9e74b22024c2)
 - Rebased with main

johnny9:
Update from [be397b6](https://github.com/bitcoin-core/gui-qml/commit/be397b6ddc034c3cd29cb3c9d40e9e74b22024c2) to [2d2be8c](https://github.com/bitcoin-core/gui-qml/commit/2d2be8c27c3d20bb6565b2560487ec158b4ffc6a)
 - Squashed UI related commits into one

johnny9 (commented):
Tested on Ubuntu
- Click Add Wallet
- Enter wallet name and click through onboarding
- Check wallet name now appears in wallet selector and check that it can be selected.

@jarolrod

src/qml/pages/main.qml
```diff
@@ -80,13 +80,17 @@ ApplicationWindow {
...
+                main.push(createWalletWizard)
+            }
+        }
     }
 
     Component {
         id: createWalletWizard
         CreateWalletWizard {
-            onFinished: {
+           onFinished: {
```
hebasto, line 93: nit: Accidental deletion?

hebasto (approved):
ACK 2d2be8c27c3d20bb6565b2560487ec158b4ffc6a,  tested on Ubuntu 24.10.

src/qml/pages/main.qml
```diff
@@ -91,11 +91,15 @@ ApplicationWindow {
...
+        DesktopWallets {
+            onAddWallet: {
+                main.push(addWalletFlow)
+            }
+        }
     }
 
     Component {
-        id: addWallet
+        id: addWalletFlow
```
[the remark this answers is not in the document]
Rsed19901: @rsed🥰🥰

src/qml/pages/main.qml
johnny9: yes, ill fix that in the next PR

PR #421, opened 2024-09-18
UI Only AU Snapshot Loading Flow

D33r-Gee:
Introducing the UI flow to be able to import a AssumeUTXO (AU) snapshot into the Bitcoin Core App.

The user will be able to load an AU snapshot from the "Connection Settings" page as designed [here](https://bitcoincore.app/snapshot/) also AU [figma](https://www.figma.com/file/ek8w3n3upbluw5UL2lGhRx/Bitcoin-Core-App-Design?type=design&node-id=8842%3A848807&mode=design&t=URVaraZVchEvzHX0-1) 

*This is a UI only commit no actual functionality has been coded. Wiring to actually import a AU snapshot will happen in a future PR 

<details>
  
Screenshot 2024-09-18 142821
Screenshot 2024-09-18 142834
Screenshot 2024-09-26 111410
Screenshot 2024-09-18 142910
Screenshot 2024-09-18 142930

</details>


src/qml/components/ConnectionSettings.qml
```diff
@@ -8,7 +8,37 @@ import QtQuick.Layouts 1.15
...
+    function setSnapshotImported(imported) {
+        snapshotImported = imported
+    }
+
     spacing: 4
+    Setting {
+        id: gotoSnapshot
+        Layout.fillWidth: true
+        header: qsTr("Import snapshot")
+        description: qsTr("Speed up the initial download")
```
MarnixCroes, line 21: as you said correctly in the import snapshot dialog, it doesn't speed up the download itself.
So this might be misleading.

I think it's better to rephrase it, something like this might be more appropriate `Start using right away and do initial synchronization in the background` or maybe something short like `Do initial synchronization in the background`

src/qml/pages/settings/SettingsProgressBar.qml
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2024 - PresentThe Bitcoin Core developers
```
MarnixCroes, line 1:
```suggestion
// Copyright (c) 2024-present The Bitcoin Core developers
```
nit
and add other files you created it has 2023 instead of 2024

MarnixCroes (commented):
- Doesn't it make more sense to (also) have this part of the onboarding workflow?
- Also, having it at `Connection` menu would not be something where I would be looking at to find this feature. Maybe it needs an `Advanced` section?
-  A UI for an incorrect snapshot and/or failure of importing the snapshot is missing

yashrajd:
Great to see this come to PR stage... how long is the loading process expected take to take on an avg computer? like in the range of seconds or minutes or hours?
We should probably discuss the designs in one of the design community calls.

src/qml/components/ConnectionSettings.qml
D33r-Gee: thanks for catching that... updated it

src/qml/pages/settings/SettingsProgressBar.qml
D33r-Gee: Great catch updated the headers with the latest push [ed27a5a](https://github.com/bitcoin-core/gui-qml/commit/ed27a5a523c4fc6586e758df940daa80e8d0dc41)

D33r-Gee:
With [ed27a5a](https://github.com/bitcoin-core/gui-qml/commit/ed27a5a523c4fc6586e758df940daa80e8d0dc41) addressed comments and added loading snapshot simulation for better testing and feedback

D33r-Gee:
> how long is the loading process expected take to take on an avg computer?

That depends on the user's device and the snapshot (signet, testnet, and mainnet) they are importing. Averages are difficult to calculate since every device is different...
Once we test the functionality (which this PR does not address, being UI only) it will become clearer how long the import process takes.

D33r-Gee:
> We should probably discuss the designs in one of the design community calls.

That's a great idea! Will be joining future calls so we can discuss design and UX matters.

meanwhile please take the time to test it on your device using one the builds [here](https://github.com/bitcoin-core/gui-qml/actions)

D33r-Gee:
minor tweak with 5c61613 changed the "Import Snapshot" description to "Instant use with background sync"

pablomartin4btc (commented):
Concept ACK

- Load/ Loading instead of Import. Everywhere in the documentation is using this term I'd keep it consistent.
- Import label description "Speed up initial download". This is not entirely true, that would be more the [assume-valid](https://river.com/learn/terms/a/assume-valid/) feature. The point of [assume-utxo](https://bitcoinops.org/en/topics/assumeutxo/) is to allow the user to see recent transactions and spend them before having to wait until the full sync is done. 
- Once the snapshot load is completed, the last page in figma, a message says that the snapshot file will be removed, perhaps this can be optional or asked to the user before starting the process.
- Last thing, the green ticked icon indicating completion (and background validation), once the validation of the snapshot chain is done, those files related with assume-utxo will be deleted automatically (chain state directory), so there's no way whether a node was "utxo-set" sped up or not. So not sure how long this green indicator would last.. definitely won't be there next restart as explained.

Please also add the [figma](https://www.figma.com/design/ek8w3n3upbluw5UL2lGhRx/Bitcoin-Core-App-Design) link of the designers team definitions in the description, I'll add these comments also there.

cc @fjahr, @Sjors, @GBKS

D33r-Gee:
With https://github.com/bitcoin-core/gui-qml/commit/a1193fcb162eb020e4fd1de29a05bbe5f4d5aea9 Changed "Import" to "Load" based on @pablomartin4btc suggestion

Also updated the description to include figma link and a screenshot of the "Progress Bar" page

D33r-Gee:
> * Once the snapshot load is completed, the last page in figma, a message says that the snapshot file will be removed, perhaps this can be optional or asked to the user before starting the process.

Yeah maybe a button can be added so that the user has the option to delete the snapshot file. This will be especially relevant for mobile (android) users looking to save storage space on their device.

D33r-Gee:
> * Last thing, the green ticked icon indicating completion (and background validation), once the validation of the snapshot chain is done, those files related with assume-utxo will be deleted automatically (chain state directory), so there's no way whether a node was "utxo-set" sped up or not. So not sure how long this green indicator would last.. definitely won't be there next restart as explained.

Great point, maybe the option can be greyed out with a check mark to indicate the snapshot has been fully validated and the node is synced.

Thoughts @yashrajd @GBKS?

D33r-Gee:
update [f016406](https://github.com/bitcoin-core/gui-qml/commit/f016406c8c239c4495043318de20a24283593371) rebase over main

GBKS:
Thanks for working on this. I just tested locally, and there are a lot of visual details that need tweaking (icon needs to be bigger, title text bold, vertical spacing too tight, line-height too tight, etc). Do you want me to list those out? It's pretty easy to spot them comparing the application to the designs.

A larger thing is the page transitions. The idea here is that the snapshot flow is just "one page deep". So instead of this:

> Settings -> Snapshot -> Loading snapshot -> Done

It's just this:

> Settings -> Snapshot

So you stay in the same screen, but the state of the screen changes and therefore the info you see, with fade out/in transitions. You can see this in the [web prototype](https://lively-kashata-cfde7e.netlify.app/screen/snapshot?t=slide-left).

---

Regarding the point about indicating that the snapshot was validated completely. My assumption was that at that point (snapshot is validated and the user has seen the screen that states so) we replace the "Load snapshot" functionality with the "Create snapshot" functionality. Does that work?

GBKS:
Also, would it be possible to fix the actions? They are currently not building all the artifacts correctly. Thanks.

D33r-Gee:
> Also, would it be possible to fix the actions? They are currently not building all the artifacts correctly. Thanks.

from @jarolrod in discord PM: "the ci error is unrelated to code and just needs to be rerun"

D33r-Gee:
> Do you want me to list those out?

No worries, I'll make the tweaks and see if they are addressed...

pablomartin4btc (commented):
ACK f016406c8c239c4495043318de20a24283593371

Regarding the commits, the first commit adds the files that don't exist yet, I'd move the first commit to the end, becoming the 4th.

The 2nd commit imports the controls that are being updated in the 3rd, so I'd swap these 2 commits.

That's all, I'll review the code when I play around with the testing.

D33r-Gee:
> That's all, I'll review the code when I play around with the testing.

Thanks for taking a look... I'm about to update the commits incorporating Christoph's comments... So please wait before testing?

D33r-Gee:
with [e5bbabd](https://github.com/bitcoin-core/gui-qml/commit/e5bbabdc68c36e4beca7d12fdbb2f102fc259e8c) squashed the commits into one, addressed spacing issues, and got rid off pages using Stack layout in SnapshotSettings.qml

pablomartin4btc:
> with [e5bbabd](https://github.com/bitcoin-core/gui-qml/commit/e5bbabdc68c36e4beca7d12fdbb2f102fc259e8c) squashed the commits into one

Ok, it was not necessary as per my [review](https://github.com/bitcoin-core/gui-qml/pull/421#pullrequestreview-2343844862), the separation made sense, just the order, but up to you.

I think for the lint failure you'll need to exclude the `QML` resources as it's done for `QT` in `get_pathspecs_exclude_whitespace` (`test/lint/test_runner/src/main.rs`): `"src/qt/res/src/*.svg",`

D33r-Gee:
> the separation made sense

Thanks, decided to adhere to the CONTIBUTING guidelines more thoroughly ;)

For the lint it turns out a couple of svg files had trailing whitespaces, [3ca75a4](https://github.com/bitcoin-core/gui-qml/commit/3ca75a4e69a984cb58555af98114606c735e98f2) addresses that

D33r-Gee:
Closing this PR since most of the work has moved to #424 and #446

PR #424, opened 2024-10-11
QML Load Snapshot Signet (160,000 height)

D33r-Gee:
Open for review and testing only. Do not merge!

# GUI Integration for UTXO Snapshot Loading

## Overview
This PR adds initial GUI support for loading a `signet` UTXO snapshot at height 160,000, building on Bitcoin Core's `assumeutxo` infrastructure.

## What This PR Does

1. Adds a basic GUI interface for loading UTXO snapshots via the Connection Settings panel
2. Implements non-blocking snapshot loading via QT Thread
3. Provides visual confirmation of snapshot activation
4. Adds a snapshot progress notifications accessible via a handler

### Core Components Modified

1. **Node Interface** (`src/node/interfaces.cpp`)
- Implements `snapshotLoad()` based on the [`loadtxoutset`](https://github.com/bitcoin/bitcoin/blob/17834bd1976df7a2ff6c2f5f05a59ae3fd3f6875/src/rpc/blockchain.cpp#L2983) RPC
- Uses background QT thread for non-blocking operation

2. **QML Integration** (`src/qml/models/nodemodel.cpp`)
- Manages snapshot loading state and progress
- Introduces a worker class to manage snapshot metadata
- Uses Qt's thread system to prevent UI blocking
- Provides progress updates via Qt signals

3. **UI Components** (`src/qml/components/SnapshotSettings.qml`)
- File selection dialog
- Snapshot verification status
- Progress tracking and visualization 

### Key Design Decisions

1. **Non-blocking Implementation**
- Snapshot loading runs in separate QT thread
- UI remains responsive during load

2. **Extensibility**
- Interface designed to accommodate future `assumeutxo` changes
- Error handling framework in place

## Testing Instructions

1. Build
2. Launch Bitcoin Core QML GUI
3. Navigate to Connection Settings (after onboarding is done)
5. Test with sample snapshot:
```
magnet:?xt=urn:btih:9da986cb27b3980ea7fd06b21e199b148d486880&dn=utxo-signet-160000.dat
```
6. Verify (see screenshots below):
   - "Snapshot Loaded" confirmation appears
   - Block clock shows ~50% progress
   - For errors, check debug.log for [snapshot] or [loadsnapshot] messages
  

<details>

Screenshot 2024-10-11 100308
Screenshot 2024-10-11 100319
Load_UTXO_Snapshots_FileDialog
Screenshot 2024-11-13 144609
Screenshot 2024-11-13 144917
Screenshot 2024-11-13 145350

</details>

### Expected Behavior
- File selection dialog works
- Success/failure state properly displayed
- Node becomes usable while background validation continues

## Future Work

1. **Error Handling**
- Add user-facing error messages
- Implement recovery paths
- Add snapshot validation UI once background validation has completed

2. **Integration**
- Coordinate with upstream `assumeutxo` changes
- Add support for network-specific snapshots (`testnet` and `mainnet`)
- Integrate with initial setup wizard

## Notes for Reviewers

 **Snapshot Compatibility**
- Load option available after Onboarding in the Connections Settings page
- Currently only works with `signet` UTXO snapshot
- Limited to specific 160,000 height
- Validates against chain parameters [`m_assumeutxo_data`](https://github.com/bitcoin-core/gui-qml/blob/f23ffb874cc5fe583b97f8fec9caff87fcf42f67/src/kernel/chainparams.cpp#L374)
(the changes there can be discarded once synced since [`m_assumeutxo_data`](https://github.com/bitcoin/bitcoin/blob/17834bd1976df7a2ff6c2f5f05a59ae3fd3f6875/src/kernel/chainparams.cpp#L499) is already hardcoded)

feedback welcome on the approach and implementation details.


D33r-Gee:
With [cb409e3](https://github.com/bitcoin-core/gui-qml/commit/cb409e3342ad4cee71aeb0102e23a208d3978485) rebased and got rid off the extra options_mode settings which are not relevant for this evaluation.

Also deleted the extra loadsnapshot.cpp file and instead added the load snapshot function in src/node/interfaces.cpp

yashrajd:
Should I be getting a binary specific for this issue? I tried loading a signet snapshot into the one for #421 but was unable to...

D33r-Gee:
> Should I be getting a binary specific for this issue?

Yes please download the artifact for your system [here](https://github.com/bitcoin-core/gui-qml/actions/runs/11351500065)

> I tried loading a signet snapshot into the one for #421 but was unable to...

That's correct behaviour, 421 is UI only and doesn't not have backend wiring

D33r-Gee:
With [670ca05](https://github.com/bitcoin-core/gui-qml/commit/670ca05c77c13dceaeecbf3d816400e4942a86f9) minor tweaks to the src/node/interfaces.cpp snapshotload() function. 

It mimics the rpc call [loadtxoutset](https://github.com/bitcoin/bitcoin/blob/479715e9db039e881d27c6fa9bb85b81e6b375e7/src/rpc/blockchain.cpp#L2985) from bitcoin/bitcoin repository

Next will connect the loading to the progress bar

D33r-Gee:
with [c59e557](https://github.com/bitcoin-core/gui-qml/commit/c59e557df8849db0940a662bbb2f7fe27404ef47) Integrated the UI with the C++ backend, implementing a progress callback function to dynamically update the loading progress based on the number of coins processed.

Touched src/validation.h and src/validation.cpp to extract the coin loading progress.

D33r-Gee:
With [e3f8c70](https://github.com/bitcoin-core/gui-qml/commit/e3f8c705493a795d473c76dbb8814c5b79f70572) moved the load snapshot functionality from nodeModel to optionsModel since it make it easier to store the snapshot info and reduces redundant variable declarations.

jarolrod (commented):
> With [e3f8c70](https://github.com/bitcoin-core/gui-qml/commit/e3f8c705493a795d473c76dbb8814c5b79f70572) moved the load snapshot functionality from nodeModel to optionsModel since it make it easier to store the snapshot info and reduces redundant variable declarations.

[e3f8c70](https://github.com/bitcoin-core/gui-qml/commit/e3f8c705493a795d473c76dbb8814c5b79f70572) is a regression because optionsmodel is not an appropriate place for this, and we are now violating the role of settings.json to store the state of a process.

NodeModel and AppMode are more appropriate places for this information. But the entire approach here is still a concept NACK as we are performing unecessary changes to validation and chainparams.

yashrajd:
Ok I tested artefact#235 and functionality & UI seems to work. 

Didn't see block clock so couldn't verify easily. Adding some UI feedback below:

assumeutxo-424-sc1 copy



assumeutxo-424-sc2 copy



assumeutxo-424-sc4 copy



assumeutxo-424-sc3 copy

src/qml/models/options_model.cpp
```diff
@@ -201,6 +211,90 @@ void OptionsQmlModel::setDataDir(QString new_data_dir)
     }
 }
 
+void OptionsQmlModel::setSnapshotLoadCompleted(bool new_snapshot_load_completed)
```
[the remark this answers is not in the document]
D33r-Gee: Thanks for taking a look at this and for the guidance regarding what can go into the optionsmodel.

It has been address with [90dc0dd](https://github.com/bitcoin-core/gui-qml/commit/90dc0dd3bf45ab1cd7ca6f3d9143c563b65f22ca)

src/qml/models/options_model.cpp
```diff
@@ -201,6 +211,90 @@ void OptionsQmlModel::setDataDir(QString new_data_dir)
...
+{
+    if (new_snapshot_load_completed != m_snapshot_load_completed) {
+        m_snapshot_load_completed = new_snapshot_load_completed;
+        if (m_onboarded) {
+            m_node.updateRwSetting("snapshotloadcompleted", new_snapshot_load_completed);
```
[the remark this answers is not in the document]
D33r-Gee: Makes sense, although I have a question for when we would want to allow the user to load a snapshot during onboarding, could the snapshot file path be stored in the settings.json?

D33r-Gee:
with [90dc0dd](https://github.com/bitcoin-core/gui-qml/commit/90dc0dd3bf45ab1cd7ca6f3d9143c563b65f22ca) addressed @jarolrod concerns and reverted back to just a minimal integration of the loading snapshot functionality.

@yashrajd I will address the UI issues next

As a follow up will investigate how to access the snapshot verification progress

Also looking into how to make the snapshot data persist and be available for display (i.e. snapshot block height, date, and merkle hash)

D33r-Gee:
> But the entire approach here is still a concept NACK as we are performing unecessary changes to validation and chainparams.

For the chainparams the changes there are only temporary for evaluating and testing within this repo. I checked [upstream](https://github.com/bitcoin/bitcoin/blob/97b790e844abd2f92c928239a7dc786d37fad18b/src/kernel/chainparams.cpp#L499) and the signet snapshot has already been hardcoded so once our repo is synced the changes in chainparams.cpp can be discarded.

D33r-Gee:
with [85592f5](https://github.com/bitcoin-core/gui-qml/commit/85592f576547befad747614d8e63fd7d48501811) made the snapshot activation persist on node restart by making use of the interface with chain->hasAssumedValidChain 

Also changed the green-check icon size to 30 pix

Next looking into how to extract the progress from the ActivateSnapshot process

D33r-Gee:
with [955c65d](https://github.com/bitcoin-core/gui-qml/commit/955c65dc1ba04bf603ac7e6ef0985d84ffbf22c8) modified the snapshotLoad() function in src/node/interfaces.cpp and the initializeSnapshot() function in nodemodel.cpp to extract progress information using regular expressions.

This design decision aimed to keep snapshot loading mechanisms self-contained.

However, it looks like the use of `std::stod` is triggering some lint failures... will look into how to address that. 

Perhaps the way to go is to use the handler in src/node/interfaces_ui ... will explore that path then update

D33r-Gee:
with [0eba513](https://github.com/bitcoin-core/gui-qml/commit/0eba513ca092675260b39e5686f9c7c0c322b7d8) reverted back to minimal snapshot load functionality.

Snapshot loading progress is still not were I want it to be so in the meantime replaced the progress bar with placeholder text.

Also updated the snapshot info display. once successfully loaded, and I'm using the hardcoded info in `chainparams` since if the snapshot has loaded successfully it means the info there has been validated.
This is make the info persistent and not rely on the `snapshot.dat` file (in case it gets deleted)

src/qml/components/SnapshotSettings.qml
```diff
@@ -0,0 +1,278 @@
...
+    property bool snapshotVerified: onboarding ? false : chainModel.isSnapshotActive
+    property string snapshotFileName: ""
+    property var snapshotInfo: ({})
+
+    id: columnLayout
+    width: Math.min(parent.width, 450)
+    anchors.horizontalCenter: parent.horizontalCenter
+
+    // TODO: Remove this once the verification progress is available
+    Timer {
```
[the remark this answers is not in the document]
D33r-Gee: yep will do

src/qml/components/SnapshotSettings.qml
```diff
@@ -0,0 +1,278 @@
...
+                }
+            }
+        }
+    }
+
+    StackLayout {
+        id: settingsStack
+        currentIndex: onboarding ? 0 : snapshotVerified ? 2 : 0
+
+        function startLoading() {
```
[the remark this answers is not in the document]
D33r-Gee: Ah ok going to use signals and connections instead

src/qml/components/SnapshotSettings.qml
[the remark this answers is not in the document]
D33r-Gee: ok added documentation commnents for testing

src/qml/components/SnapshotSettings.qml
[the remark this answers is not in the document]
D33r-Gee: using nodeModel member variable so that snapshotLoading persist while navigating away from settings page.

D33r-Gee:
with [5d0756c](https://github.com/bitcoin-core/gui-qml/commit/5d0756c62c3de75d40235b52e4720264883de0bd) updated `SnapshotSettings.qml` with comments documenting the loading process.

with [262bb03](https://github.com/bitcoin-core/gui-qml/commit/262bb03910f12c15a31a8c78c72de38656c7e67d) updated `nodemodel` by adding the `m_snapshot_loading` bool so that the "Snapshot Loading" page persist if the user navigates away from it.

Next looking into how to finesse the snapshot loading progress...

D33r-Gee:
with [a6bede5](https://github.com/bitcoin-core/gui-qml/commit/a6bede532cb44da77874304d9d285e3497930867) removed comments from `SnapshotSettings.qml`

src/qml/components/SnapshotSettings.qml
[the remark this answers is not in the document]
D33r-Gee: deleted documentation comments and instead update the PR description

D33r-Gee:
with [799b4cd](https://github.com/bitcoin-core/gui-qml/commit/799b4cd16eaf5fdf2cfb61b2baa74b413efb023a) updated `SnapshotSettings.qml` by deleting comments and adding  more properties to `filedialog`.

Also wanted to see if the Win64 CI completed successfully, which it did.

Next will ask for feedback before un-drafting

D33r-Gee:
with [5d2ffb0](https://github.com/bitcoin-core/gui-qml/commit/5d2ffb0c48de644137a76022cf8bb3509a675340) updated snapshotLoad() in src/node/interfaces.cpp. Left the polling loop for now as I investigate how to handle loading the snapshot if Pre-syncing and Syncing headers is still happening.

Also removed JS functions and moved the logic to the nodemodel

D33r-Gee:
[8beb50a](https://github.com/bitcoin-core/gui-qml/commit/8beb50ab504c6a76daf2b164d6786c2bafa6cba3) rebased over main

src/kernel/chainparams.cpp
```diff
@@ -371,6 +371,13 @@ class SigNetParams : public CChainParams {
 
         vFixedSeeds.clear();
 
+        m_assumeutxo_data = MapAssumeutxo{
+            {
+                160000,
```
Sjors, line 376: 8beb50ab504c6a76daf2b164d6786c2bafa6cba3: the signet snapshot was merged into Bitcoin Core more than a year ago in https://github.com/bitcoin/bitcoin/commit/edbed31066e3674ba52b8c093ab235625527f383. This also seems to be using the older serialization format from before https://github.com/bitcoin/bitcoin/pull/29612.

I guess that's all fine for a proof of concept, but it would be better if the main branch of this repo was more up to date.

D33r-Gee: Hi Sjors! Thanks for taking a look and yes completely agree with:

> it would be better if the main branch of this repo was more up to date

We are currently taking steps to sync with upstream, meanwhile I wanted to get the ball rolling, although I will pull in the update versions of the assume utxo functionality so that when the sync happens this PR can be merged with minimal refactoring.

src/node/interfaces.cpp
```diff
@@ -395,9 +396,42 @@ class NodeImpl : public Node
...
+
+        if (!m_context->chainman) { return false; }
+
+        ChainstateManager& chainman = *m_context->chainman;
+        CBlockIndex* snapshot_start_block = nullptr;
+
+        // Wait for the block to appear in the block index
+        //TODO: remove this once another method is implemented
+        constexpr int max_wait_seconds = 600; // 10 minutes
+        for (int i = 0; i < max_wait_seconds; ++i) {
```
johnny9, line 423: We shouldn't have a blocking component here. If there are any issues we should just fail.

src/node/interfaces.cpp
```diff
@@ -395,9 +396,42 @@ class NodeImpl : public Node
...
+            if (snapshot_start_block) break;
+            std::this_thread::sleep_for(std::chrono::seconds(1));
+        }
+
+        return chainman.ActivateSnapshot(afile, metadata, false);
+    }
     ArgsManager& args() { return *Assert(Assert(m_context)->args); }
     ChainstateManager& chainman() { return *Assert(m_context->chainman); }
     NodeContext* m_context{nullptr};
+    std::atomic<double> m_snapshot_progress{0.0};
```
johnny9, line 434: I imagine the actual progress should be owned by the object in charge of doing the processing. ChainStateManager?

johnny9 (commented):
A handler similar to handleNotifyBlockTip will be needed to so the UI interface can be event driven instead of polling the interfaces module.

src/kernel/chainparams.cpp
Sjors: Another thing you could do is look into making a PR to the Bitcoin Core repo with your proposed interface changes. One way to go about that is to somehow use them in existing or a new RPC method. E.g. snapshot load progress is probably useful to have available.

src/qml/components/SnapshotSettings.qml
```diff
@@ -0,0 +1,200 @@
...
+                Layout.fillWidth: true
+                Layout.topMargin: 20
+                headerBold: true
+                header: qsTr("Snapshot Loaded")
+                descriptionBold: false
+                descriptionColor: Theme.color.neutral6
+                descriptionSize: 17
+                descriptionLineHeight: 1.1
+                description: snapshotInfo && snapshotInfo["date"] ?
+                    qsTr("It contains transactions up to %1. Newer transactions still need to be downloaded." +
```
Sjors, line 129: It's not really "transactions up to", it's all coins (utxo set) that exist at the snapshot height.

Maybe say something like:

> It contains a snapshot as of block %1. Once loaded, all blocks from %2 to the present are download. After that Bitcoin Core is ready to use, while all historical blocks from genesis to %2 are downloaded and validated in the background.

D33r-Gee: Great suggestion! I will update the description with the next push.

src/node/interfaces.cpp
D33r-Gee: Yes! Will set it up so that the handler notifies the snapshot progress and that ownership belongs to ChainstateManager.

src/kernel/chainparams.cpp
D33r-Gee: > Another thing you could do is look into making a PR to the Bitcoin Core repo with your proposed interface changes.

That's the current medium term plan, once we have an implementation that works and makes sense here we plan to go upstream (bitcoin/bitcoin) and open a PR there.

src/qml/components/SnapshotSettings.qml
GBKS: In the sequence, this is the fourth snapshot-related screen (state) people see:

1. Snapshot option in settings where people learn about the benefit of this feature
2. "Load snapshot" screen that gives more detail about how it works
3. "Loading snapshot" screen state with a loader that tells people they can do other stuff in the mean time
4. "Snapshot loaded" screen state that informs people what was imported, what to keep in mind and what will happen next (this is the screen that we are talking about in this thread)

I think the point that not all historical transactions are in the snapshot is worth clarifying. The first time people are exposed to what this feature does is in 2, so I'd probably start tweaking at that point. We call it a "recent transaction snapshot" there, which is more precise than just "transactions" ("unspent transaction snapshot" would be a little closer).

We don't need to be 100% technically accurate and I'm not sure users are well-served if we're completely vague about this either. They should be able to form a good enough mental model for making decisions and understanding their impact.

So how about this?

> It contains recent transactions up to %1. Newer and historical transactions will be automatically downloaded and verified.

Or this?

> It contains unspent transactions up to %1. Newer and historical transactions will be automatically downloaded and verified.

Either way, we should ensure to tweak the text across the sequence if we make changes here.

We show the block height in a "View details" option below (did not check if it's implemented in this PR already), so not sure whether we need to include that in the text as well.

As for what happens next, my thinking was that we just say that the data will be filled in and verified here (keep it short), and then have indicators contextually in the UI as to what is happening (balance, send and transaction list have respective load states). So less absorption of logic here, and more contextual info in the appropriate places.

Sorry, this has gotten quite long. Hope it's a helpful response.

Sjors: > We don't need to be 100% technically accurate and I'm not sure users are well-served if we're completely vague about this either. 

I think it's better to be incomplete than incorrect. The nice thing about being vague is that it's clear to the reader there's more to learn.

One reason the phrasing "recent / historical transactions" is potentially confusing is that people may think they can recover historical wallet transactions with it. But it has nothing to do with that. 

The concept that's useful to explain is that the snapshot allows us to start syncing at block %1, saving time by skipping historical blocks. This way the user can get started (with their wallet) earlier. Meanwhile it's safe, because we do check those historical blocks in the background later.

yashrajd: TL;DR I reviewed the onboarding text again, and it seems like the use of `transactions` terminology in the snapshot section provides context continuity with what's said before...see screenshot below.

Screenshot 2024-12-13 at 12 45 27

Unless we come up with something better, I'd be fine with retaining what we have currently though I agree it's inaccurate. Suggestions below:

----------

> I think it's better to be incomplete than incorrect. The nice thing about being vague is that it's clear to the reader there's more to learn.

It does seem like we can strike a better balance between technical accuracy & usefulness to user here, and use vagueness where required.

> We show the block height in a "View details" option below (did not check if it's implemented in this PR already), so not sure whether we need to include that in the text as well.

Agree, this was in the initial mockups I did. (Ideally, I'd like to see these BEFORE the loading process even began as a way of informing the user context about what's about to happen, but after loading is fine too.)

> The first time people are exposed to what this feature does is in 2, so I'd probably start tweaking at that point.

Agree.

> 2. "Load snapshot" screen that gives more detail about how it works.

This screen can omit mention of transactions **"The snapshot allows you to start using the application more quickly.  Once loaded, it will be automatically verified in the background. [Learn more](some_url)"** instead of _"You can start using the application more quickly by loading a recent transaction snapshot. It will be automatically verified in the background."_

> 4. "Snapshot loaded" screen state that informs people what was imported, what to keep in mind and what will happen next (this is the screen that we are talking about in this thread)

I'd suggest something like **"The loaded snapshot contains data up to <date>. The initial download will continue, while the snapshot is verified in the background."** instead of _"It contains transactions up to September 10, 2023.
Newer transactions still need to be downloaded. The data will be verified in the background."_

Frame 10036

GBKS: Should we move this conversation into a separate issue? It's just about text details, and maybe we don't want to weigh down the overall technical development of this feature? Text is easy to replace separately.

Sjors: @GBKS that makes sense.

D33r-Gee:
with [eb923ff](https://github.com/bitcoin-core/gui-qml/commit/eb923ff46f85580c52bfabc28055aa9c54a29cf9):

- Introduced a handler to keep track of the snapshot loading progress
it utilizes the notifications framework, this displays in the GUI as a progress indicator.

- Added a worker class in `nodemodel` to handle the processing of the snapshot file (i.e. metadata)

- Got rid of the blocking `for` loop that was waiting for the corresponding block header.
If the block header is not found the loading just fails and allows the user to wait a try again later.

TODO: in a follow up PR, add error handling and messaging to the user when loading fails.

D33r-Gee:
With [68efab5](https://github.com/bitcoin-core/gui-qml/commit/68efab582385135d37e0a79c87d8bd43dfccc37f) this PR is ready to transition from draft status.

The changes are separated into two distinct commits:

1. Core functionality changes intended for upstream integration into bitcoin/bitcoin (bd0f1204a2057764813f1f63621fde92f32c5c0a)

2. QML-specific implementations and adaptations (68efab582385135d37e0a79c87d8bd43dfccc37f)

D33r-Gee:
with [6c25cd7](https://github.com/bitcoin-core/gui-qml/commit/6c25cd7d2b3ad3a4e88904ab8b5b7bbd7d25650b) :

- rebased with main
- updated the qml pages to behave more like the designer feedback in #446
- Also update some of the cpp code to enable the UI changes

Will keep this as draft until we get closer to the sync with upstream

src/qml/models/options_model.cpp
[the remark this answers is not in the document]
D33r-Gee: will follow up on this in another PR

src/kernel/chainparams.cpp
D33r-Gee: Opened https://github.com/bitcoin-core/gui/pull/870 as a draft based on this exchange...

D33r-Gee: Will open a PR in `bitcoin/bitcoin` when appropriate...

src/qml/components/SnapshotSettings.qml
D33r-Gee: I believe this is resolved with [4e43df9](https://github.com/bitcoin-core/gui-qml/commit/4e43df99aea4b26c6f5a97de58eb87c0777f4766)

Is this correct @yashrajd and/or @GBKS ?

D33r-Gee:
With [bbbbf42](https://github.com/bitcoin-core/gui-qml/commit/bbbbf428eb0353ceb79a56a0a269e778cf8fa440) adds error page to handle when a invalid snapshot gets used, this is based on @GBKS feedback in #449 

Feedback welcomed!

hebasto:
> Based on #421.

This is confusing, given that #421 is closed.

D33r-Gee:
[e02f382](https://github.com/bitcoin-core/gui-qml/commit/e02f382894f967e14e5d9b7f61f23fb6a29da296) rebased over main

D33r-Gee:
friendly ping @yashrajd and @johnny9

D33r-Gee:
with [d2fccc4](https://github.com/bitcoin-core/gui-qml/commit/d2fccc49e6638dc0a6228b98f5c118239c0c6250) rebased over main

yashrajd:
ok I just tested the latest build on signet...following are observations:

tl;dr – snapshot loading worked!

(but shown hash doesn't seem to match file hash from terminal (but maybe I'm barking up the wrong tree)

assumeutxo testing

-----------


- I see the block status widget updated after snapshot loading is completed.

after loading snapshot


--------

- then I quit the app, and relaunched it....the block status widget not on the same page anymore...

relaunced app after loading


---------

- go to Connection Settings, the Load snapshot option has disappeared (iirc this is expected behaviour) so I assume it still worked.


load snapshot option has disappeared


Did I test it right? Did I miss something I should've tested...lemme know.

D33r-Gee:
> (but shown hash doesn't seem to match file hash from terminal (but maybe I'm barking up the wrong tree)

Thanks @yashrajd  for the attached screenshot showing the two different hashes associated with the loaded UTXO snapshot.

I’d like to clarify the distinction between them to help avoid any confusion:

1. **SHA-256 Hash of the Snapshot Data File**

   * This hash is generated by running a command such as `shasum -a 256 utxo-signet-160000.dat` directly on the snapshot file.
   * Example from the image:

     ```
     eeeca845385ba91e84ef58c9d38f98f246a24feadaad57fe1e5874f3f92ef8c  utxo-signet-160000.dat
     ```
   * It represents a checksum of the full file contents and is typically used to verify file integrity during download or transfer.

2. **Hash Displayed in the Application (Snapshot Metadata Hash)**

   * The hash shown in the user interface (e.g., `5225141cb62dee63ab3be95f9b03d60801f264010b1816d4bd00618b2736e7be`) is **not** the SHA-256 of the file itself.
   * Rather, it’s the hash of the **serialized snapshot metadata**, generated internally using the `hash_serialized` function after parsing the file.
   * This metadata hash is what Bitcoin Core uses to validate the snapshot’s correctness and ensure compatibility with the expected chain state. You can find relevant references in the source code, such as `AssumeutxoHash` in `chainparams.cpp` and `ExpectedAssumeutxo` in `validation.cpp`.

I hope this helps clarify the purpose and origin of each hash. Please let me know if the above makes sense?

D33r-Gee:
with [508b486](https://github.com/bitcoin-core/gui-qml/commit/508b486cbece24d386e25eba014436107eeb87e4) rebased over main

hebasto:
Closing for now.

PR #427, opened 2024-10-25
Introduce Signal Based Navigation Model to have Self-Contained Pages

jarolrod:
Here we introduce a Signal Based navigation model, so that we can have self-contained pages. Self-contained pages include everything they need where they are declared, and do not have more than one definition. This makes pages modular, easier to reason about, but also easier to maintain by eliminating dependencies on hard-coded IDs, direct calls to parent views, and interlinked properties between pages.

When we want to reason about `PageA.qml`, we only have to look in there and the child components it uses, and not have to use a search function to ensure we have the entire scope of `PageA.qml`.

To Achieve this, this PR implements the following main ideas:

1. Ensure that pages are not hardcoding and reaching into an outside view.
2. Ensure we don't alias down any components into pages, and instead have pages define all their components, but react to state events
3. Ensure we don't have interlinked properties between pages
4. Move navigation and page linking logic from the pages themselves and into the declaration of the view itself.

Additionally, while here we structure the CreateWallet flow into a Wizard.

**Follow-Ups:**

Consider moving onboarding and walletName into more appropriate backend objects.

[Build Artifacts]()

### DIFF
--- a/src/qml/pages/settings/SettingsAbout.qml
+++ b/src/qml/pages/settings/SettingsAbout.qml
@@ -32,7 +32,9 @@ Item {
             description: qsTr("Bitcoin Core is an open source project.\nIf you find it useful, please contribute.\n\n This is experimental software.")
             descriptionMargin: 20
             detailActive: true
-            detailItem: AboutOptions {}
+            detailItem: AboutOptions {
+                onNext: aboutSwipe.incrementCurrentIndex()
+            }
         }
         SettingsDeveloper {
             id: about_developer
--- a/src/qml/pages/settings/SettingsAbout.qml
+++ b/src/qml/pages/settings/SettingsAbout.qml
@@ -9,16 +9,11 @@ import "../../controls"
 import "../../components"
 
 Item {
-    property alias navLeftDetail: aboutSwipe.navLeftDetail
-    property alias navMiddleDetail: aboutSwipe.navMiddleDetail
-    property alias devMiddleDetail: aboutSwipe.devMiddleDetail
-    property alias showHeader: aboutSwipe.showHeader
+    id: root
+    signal back
+    property bool onboarding: false
     SwipeView {
         id: aboutSwipe
-        property alias navLeftDetail: about_settings.navLeftDetail
-        property alias navMiddleDetail: about_settings.navMiddleDetail
-        property alias devMiddleDetail: about_developer.navMiddleDetail
-        property alias showHeader: about_settings.showHeader
         anchors.fill: parent
         interactive: false
         orientation: Qt.Horizontal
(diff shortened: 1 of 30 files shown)


src/qml/pages/settings/SettingsConnection.qml
```diff
@@ -28,13 +24,63 @@ Item {
...
+                State {
+                    when: !root.onboarding
+                    PropertyChanges {
+                        target: connection_settings
+                        navLeftDetail: backButton
+                        navMiddleDetail: header
+                        navRightDetail: null
+                    }
+                }
+
```
MarnixCroes, line 54: remove white line?

src/qml/pages/settings/SettingsAbout.qml
```diff
@@ -27,23 +22,58 @@ Item {
...
             headerMargin: 0
             description: qsTr("Bitcoin Core is an open source project.\nIf you find it useful, please contribute.\n\n This is experimental software.")
             descriptionMargin: 20
             detailActive: true
-            detailItem: AboutOptions {}
+            detailItem: AboutOptions {
+                onNext: aboutSwipe.incrementCurrentIndex()
+            }
+
+            states : [
```
MarnixCroes, line 35: nit
```suggestion
            states: [
```

src/qml/pages/settings/SettingsConnection.qml
```diff
@@ -28,13 +24,63 @@ Item {
...
+            showHeader: root.onboarding
             headerText: qsTr("Connection settings")
             headerMargin: 0
             detailActive: true
-            detailItem: ConnectionSettings {}
+            detailItem: ConnectionSettings {
+                onNext: connectionSwipe.incrementCurrentIndex()
+            }
+
+            states : [
```
MarnixCroes, line 35:
```suggestion
            states: [
```

src/qml/pages/settings/SettingsStorage.qml
```diff
@@ -9,10 +9,57 @@ import "../../controls"
...
+    id: root
+    property bool onboarding: false 
     bannerActive: false
     bold: true
+    showHeader: root.onboarding
     headerText: qsTr("Storage settings")
     headerMargin: 0
     detailActive: true
     detailItem: StorageSettings {}
+    states : [
```
MarnixCroes, line 21: same

MarnixCroes (commented):
cACK
I went through all the pages, including onboarding and it work ok.

related lint failure should be fixable

johnny9:
Concept ACK

No qml errors when going through onboard. Looks like there are just some whitespace issues to fix to make the lint job happy.

johnny9 (approved):
ACK 3f9424042b7bcaa9a4bafc6a466b3e8d0283325c

MarnixCroes (approved):
ACK 3f9424042b7bcaa9a4bafc6a466b3e8d0283325c

pablomartin4btc (commented):
post-merge tACK 3f9424042b7bcaa9a4bafc6a466b3e8d0283325c

> we should set a style guide actually

As discussed offline, we should do that + some specific dev notes for this repo.

PR #430, opened 2024-11-01
Allow IPv6 in Proxy settings and moving validation out from the UI into the model/ interface

pablomartin4btc:
The main purposes of this PR are:
- Allow configure proxy server with IPv6 (after testing exhaustively in current `bitcoin-qt` [repo](https://github.com/bitcoin-core/gui/) and `bitcoind` - e.g. [IPv6 on Qt](https://github.com/bitcoin-core/gui/pull/836#issue-2527108426); [IP validation in Qt](https://github.com/bitcoin-core/gui/pull/813))
- More importantly: moving the validation logic out from the UI (component/ control/ widget) into the back-end (model/ interfaces). It seemed currently in [Qt](https://github.com/bitcoin-core/gui/) all this logic (validation, network classes) was coupled to the UI since the beginning (more than [12y ago](https://github.com/bitcoin-core/gui/blame/c4443c2be141e5f45bb10376056f3083e97cde50/src/qt/optionsmodel.cpp)) instead of using interfaces (introduced much later) that would be the correct thing to do.

Feedback and suggestions are very welcome and will help establish a foundation for leaving business logic out of the UI going forward.

---
<details>
<summary>In order to test it, if it's the first time you run <code>./src/qt/bitcoin-qt</code> you need to go thru the on-boarding flow (<i>start->next->next->next->next->connection settings->Proxy settings</i>) otherwise you need to go to the settings (top right gear icon) then <i>Connection->Proxy settings</i>.</summary>
 
image

</details>

Before this change, only IPv4 address were allow in the value input, now also IPv6 can be entered.

---
There are some [look and feel kind of issues](https://github.com/bitcoin-core/gui-qml/pull/430#issuecomment-2498871269) that will be handled on follow-ups (issues: #437 and #438).

### DIFF
--- a/src/interfaces/node.h
+++ b/src/interfaces/node.h
@@ -22,6 +22,9 @@
 #include <tuple>
 #include <vector>
 
+static const char DEFAULT_PROXY_HOST[] = "127.0.0.1";
+static constexpr uint16_t DEFAULT_PROXY_PORT = 9050;
+
 class BanMan;
 class CFeeRate;
 class CNodeStats;
@@ -125,6 +128,12 @@ public:
 
     //! Get proxy.
     virtual bool getProxy(Network net, Proxy& proxy_info) = 0;
+
+    //! Get default proxy address.
+    virtual std::string defaultProxyAddress() = 0;
+
+    //! Validate a proxy address.
+    virtual bool validateProxyAddress(const std::string& addr_port) = 0;
 
     //! Get number of connections.
     virtual size_t getNodeCount(ConnectionDirection flags) = 0;
--- a/src/node/interfaces.cpp
+++ b/src/node/interfaces.cpp
@@ -45,6 +45,7 @@
 #include <uint256.h>
 #include <univalue.h>
 #include <util/check.h>
+#include <util/strencodings.h>
 #include <util/translation.h>
 #include <validation.h>
 #include <validationinterface.h>
@@ -169,6 +170,27 @@ public:
     }
     void mapPort(bool use_upnp, bool use_natpmp) override { StartMapPort(use_upnp, use_natpmp); }
     bool getProxy(Network net, Proxy& proxy_info) override { return GetProxy(net, proxy_info); }
+    std::string defaultProxyAddress() override
+    {
+        return std::string(DEFAULT_PROXY_HOST) + ":" + ToString(DEFAULT_PROXY_PORT);
+    }
+    bool validateProxyAddress(const std::string& addr_port) override
+    {
+        uint16_t port{0};
+        std::string hostname;
+        // First, attempt to split the input address into hostname and port components.
+        // We call SplitHostPort to validate that a port is provided in addr_port.
+        // If either splitting fails or port is zero (not specified), return false.
+        if (!SplitHostPort(addr_port, port, hostname) || !port) return false;
+
+        // Create a service endpoint (CService) from the address and port.
+        // If port is missing in addr_port, DEFAULT_PROXY_PORT is used as the fallback.
+        CService serv(LookupNumeric(addr_port, DEFAULT_PROXY_PORT));
+
+        // Construct the Proxy with the service endpoint and return if it's valid
+        Proxy addrProxy = Proxy(serv, true);
+        return addrProxy.IsValid();
+    }
     size_t getNodeCount(ConnectionDirection flags) override
     {
         return m_context->connman ? m_context->connman->GetNodeCount(flags) : 0;
--- a/src/qml/components/ProxySettings.qml
+++ b/src/qml/components/ProxySettings.qml
@@ -7,7 +7,12 @@ import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.15
 import "../controls"
 
+import org.bitcoincore.qt 1.0
+
 ColumnLayout {
+    property string ipAndPortHeader: qsTr("IP and Port")
+    property string invalidIpError: qsTr("Invalid IP address or port format. Use '255.255.255.255:65535' or '[ffff::]:65535'")
+
     spacing: 4
     Header {
         headerBold: true
@@ -41,14 +46,17 @@ ColumnLayout {
     Setting {
         id: defaultProxy
         Layout.fillWidth: true
-        header: qsTr("IP and Port")
-        errorText: qsTr("Invalid IP address or port format. Please use the format '255.255.255.255:65535'.")
+        header: ipAndPortHeader
+        errorText: invalidIpError
         state: !defaultProxyEnable.loadedItem.checked ? "DISABLED" : "FILLED"
         showErrorText: !defaultProxy.loadedItem.validInput && defaultProxyEnable.loadedItem.checked
         actionItem: IPAddressValueInput {
             parentState: defaultProxy.state
-            description: "127.0.0.1:9050"
+            description: nodeModel.defaultProxyAddress()
             activeFocusOnTab: true
+            onTextChanged: {
+                validInput = nodeModel.validateProxyAddress(text);
+            }
         }
         onClicked: {
             loadedItem.filled = true
@@ -89,14 +97,17 @@ ColumnLayout {
     Setting {
         id: torProxy
         Layout.fillWidth: true
-        header: qsTr("IP and Port")
-        errorText: qsTr("Invalid IP address or port format. Please use the format '255.255.255.255:65535'.")
+        header: ipAndPortHeader
+        errorText: invalidIpError
         state: !torProxyEnable.loadedItem.checked ? "DISABLED" : "FILLED"
         showErrorText: !torProxy.loadedItem.validInput && torProxyEnable.loadedItem.checked
         actionItem: IPAddressValueInput {
             parentState: torProxy.state
-            description: "127.0.0.1:9050"
+            description: nodeModel.defaultProxyAddress()
             activeFocusOnTab: true
+            onTextChanged: {
+                validInput = nodeModel.validateProxyAddress(text);
+            }
         }
         onClicked: {
             loadedItem.filled = true
--- a/src/qml/controls/IPAddressValueInput.qml
+++ b/src/qml/controls/IPAddressValueInput.qml
@@ -16,9 +16,9 @@ TextInput {
     property bool validInput: true
     enabled: true
     state: root.parentState
-    validator: RegExpValidator { regExp: /[0-9.:]*/ } // Allow only digits, dots, and colons
+    validator: RegularExpressionValidator { regularExpression: /^[\][0-9a-f.:]+$/i } // Allow only IPv4/ IPv6 chars
 
-    maximumLength: 21
+    maximumLength: 47
 
     states: [
         State {
@@ -52,31 +52,5 @@ TextInput {
 
     Behavior on color {
         ColorAnimation { duration: 150 }
-    }
-
-    function isValidIPPort(input)
-    {
-        var parts = input.split(":");
-        if (parts.length !== 2) return false;
-        if (parts[1].length === 0) return false; // port part is empty
-        var ipAddress = parts[0];
-        var ipAddressParts = ipAddress.split(".");
-        if (ipAddressParts.length !== 4) return false;
-        for (var i = 0; (i < ipAddressParts.length); i++) {
-            if (ipAddressParts[i].length === 0) return false; // ip group number part is empty
-            if (parseInt(ipAddressParts[i]) > 255) return false;
-        }
-        var port = parseInt(parts[1]);
-        if (port < 1 || port > 65535) return false;
-        return true;
-    }
-
-    // Connections element to ensure validation on editing finished
-    Connections {
-        target: root
-        function onTextChanged() {
-            // Validate the input whenever editing is finished
-            validInput = isValidIPPort(root.text);
-        }
     }
 }
--- a/src/qml/models/nodemodel.cpp
+++ b/src/qml/models/nodemodel.cpp
@@ -166,3 +166,13 @@ void NodeModel::ConnectToNumConnectionsChangedSignal()
             setNumOutboundPeers(new_num_peers.outbound_full_relay + new_num_peers.block_relay);
         });
 }
+
+bool NodeModel::validateProxyAddress(QString address_port)
+{
+    return m_node.validateProxyAddress(address_port.toStdString());
+}
+
+QString NodeModel::defaultProxyAddress()
+{
+    return QString::fromStdString(m_node.defaultProxyAddress());
+}
--- a/src/qml/models/nodemodel.h
+++ b/src/qml/models/nodemodel.h
@@ -62,6 +62,9 @@ public:
     void startShutdownPolling();
     void stopShutdownPolling();
 
+    Q_INVOKABLE bool validateProxyAddress(QString addr_port);
+    Q_INVOKABLE QString defaultProxyAddress();
+
 public Q_SLOTS:
     void initializeResult(bool success, interfaces::BlockAndHeaderTipInfo tip_info);
 


src/qml/controls/IPAddressValueInput.qml
```diff
@@ -16,7 +16,7 @@ TextInput {
     property bool validInput: true
     enabled: true
     state: root.parentState
-    validator: RegExpValidator { regExp: /[0-9.:]*/ } // Allow only digits, dots, and colons
+    validator: RegExpValidator { regExp: /[0-9A-Fa-f.:\[\]]*/ } // Allow only IPv4/ IPv6 chars
```
vasild, line 19: I read https://doc.qt.io/qt-5/qml-qtquick-regexpvalidator.html but it is unclear whether the framework would prefix the regexp with `^` and suffix it with `$`. If it does not then we need to add it ourselves, because `[0-9.:]*` matches anything, e.g.:

```sh
$ echo xyz |grep -E '[0-9.:]*'
xyz
```
(this is because by default there is a match if the input _contains_ a matching pattern and `[0-9.:]*` denotes 0 or more chars, so "xyz" contains 0 of those chars followed by "xyz")

In the examples from the doc they add an explicit `$` at the end but not `^` at the front, hmm?

In a bracket expression `\` looses its special meaning. So the above closes the bracket expression at the first `]`: `[0-9A-Fa-f.:\[\]` followed by a literal `]`. Probably the regex should be something like `^[][0-9A-Fa-f.:]*$`, but is this even needed, given that there is a validation function?

src/qml/controls/IPAddressValueInput.qml
```diff
@@ -53,30 +53,4 @@ TextInput {
...
-    {
-        var parts = input.split(":");
-        if (parts.length !== 2) return false;
-        if (parts[1].length === 0) return false; // port part is empty
-        var ipAddress = parts[0];
-        var ipAddressParts = ipAddress.split(".");
-        if (ipAddressParts.length !== 4) return false;
-        for (var i = 0; (i < ipAddressParts.length); i++) {
-            if (ipAddressParts[i].length === 0) return false; // ip group number part is empty
-            if (parseInt(ipAddressParts[i]) > 255) return false;
```
vasild, line 67: Could have been negative and still considered as valid: `1.2.-3.4`

src/qml/models/nodemodel.cpp
```diff
@@ -166,3 +166,13 @@ void NodeModel::ConnectToNumConnectionsChangedSignal()
             setNumOutboundPeers(new_num_peers.outbound_full_relay + new_num_peers.block_relay);
         });
 }
+
+bool NodeModel::validateProxyAddress(QString ipAddress)
```
vasild, line 170: Variable names should use snake_case, so `ip_address`, but maybe `addr_port` or even just `proxy` or `proxy_str` is a better name, especially if `unix:/path/to/socket` is to be accepted.

src/node/interfaces.cpp
```diff
@@ -398,6 +399,21 @@ class NodeImpl : public Node
     ArgsManager& args() { return *Assert(Assert(m_context)->args); }
     ChainstateManager& chainman() { return *Assert(m_context->chainman); }
     NodeContext* m_context{nullptr};
+    bool validateProxyAddress(const std::string ipAddress) override
```
vasild, line 402: For consistency it would be best if the same rules apply to the `-proxy=` config and the text that is accepted via the GUI. For the former the validation is happening here:

https://github.com/bitcoin/bitcoin/blob/f1bcf3edc5027b501616670db33d8be1f2cb5a11/src/init.cpp#L1140-L1168

and here

https://github.com/bitcoin/bitcoin/blob/f1bcf3edc5027b501616670db33d8be1f2cb5a11/src/init.cpp#L1513-L1526

I am not sure how relevant the first check is for the proxy because the second check ends up calling `SplitHostPort()` as well. Maybe ignore the first, move the second to a new function and call this new function from the GUI and from `init.cpp`?

vasild (commented):
Concept ACK

src/interfaces/node.h
```diff
@@ -268,6 +271,8 @@ class Node
     //! accessible across processes.
     virtual node::NodeContext* context() { return nullptr; }
     virtual void setContext(node::NodeContext* context) { }
+    virtual bool validateProxyAddress(const std::string ipAddress) = 0;
```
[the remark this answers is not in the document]
pablomartin4btc: Yes, I had it that way, somehow I lost it with testing, I'll add it back.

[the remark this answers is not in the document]
pablomartin4btc: Yes, I had it that way, somehow I lost it with testing, I'll add it back.

src/interfaces/node.h
```diff
@@ -268,6 +271,8 @@ class Node
...
+    virtual std::string defaultProxyAddress() = 0;
```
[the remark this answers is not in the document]
pablomartin4btc: Ok, will do.

src/interfaces/node.h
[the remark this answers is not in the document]
pablomartin4btc: I agree, ok.

src/qml/components/ProxySettings.qml
```diff
@@ -7,7 +7,12 @@ import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.15
 import "../controls"
 
+import org.bitcoincore.qt 1.0
+
 ColumnLayout {
+    property string ipAndPortHeader: qsTr("IP and Port")
+    property string invalidIpError: qsTr("Invalid IP address or port format.")
```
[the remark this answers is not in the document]
pablomartin4btc: Ok, I thought it was too long.

src/qml/controls/IPAddressValueInput.qml
pablomartin4btc: Yeah, all that will be gone anyways.

src/qml/models/nodemodel.cpp
pablomartin4btc: True, ok.

src/node/interfaces.cpp
```diff
@@ -398,6 +399,21 @@ class NodeImpl : public Node
...
+    {
+        uint16_t port{0};
+        std::string hostname;
+        if (!SplitHostPort(ipAddress, port, hostname) || !port) return false;
+
+        CService serv(LookupNumeric(ipAddress, DEFAULT_PROXY_PORT));
```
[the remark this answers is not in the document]
pablomartin4btc: Ok, it would make sense to remove the call `SplitHostPort()` then, also because we are not returning what the cause of the invalid address is. I'll also add comments on the intention of this logic.

src/qml/controls/IPAddressValueInput.qml
pablomartin4btc: > but is this even needed, given that there is a validation function?

It's a first landing and it will avoid the user to enter invalid symbols or chars, intentionally or not.

src/node/interfaces.cpp
pablomartin4btc: Ok, let me do some testing on this. Thanks!

src/node/interfaces.cpp
```diff
@@ -398,6 +399,21 @@ class NodeImpl : public Node
...
+
+        Proxy addrProxy = Proxy(serv, true);
```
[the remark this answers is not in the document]
pablomartin4btc: As discussed offline, in the upstream repo, support to unix sockets was added (https://github.com/bitcoin/bitcoin/pull/27375) already but as this repo is not refreshed and updated yet, I've created https://github.com/bitcoin-core/gui-qml/issues/431 and will do a follow-up on this PR once we get the feature added here.

src/qml/controls/IPAddressValueInput.qml
pablomartin4btc: > In a bracket expression \ looses its special meaning... Probably the regex should be something like ^[][0-9A-Fa-f.:]*$

That's true for the terminal, but it seems regular expressions are handled a bit differently in QML.

In the terminal, `grep` interprets `[` and `]` differently and doesn’t use the same escaping rules as some programming environments (like QML). When using grep, we need to adjust the syntax to avoid issues with special characters and escaping but QML uses JavaScript-based regex syntax (similar to ECMAScript). I can use your suggestion as it works in both environments and the one I was adding was only working in QML.

Anyways, the idea of having this regex validator (in fact looking at the doc ref you provided above I also realised that class is deprecated and should use `RegularExpressionValidator` now) is intended to ensure that an input contains only characters that are valid in IPv4 and IPv6 addresses, not to validate the format (as it would do the logic added into the node interface), it's purely esthetic. I've tried other alternatives like `inputMask:` (horrible UX) and classes like `QHostAddress` or `QNetworkAddressEntry` but the regex validator seems very practical and simplest.

src/qml/models/nodemodel.cpp
pablomartin4btc: >  especially if unix:/path/to/socket is to be accepted.

As explained [above](https://github.com/bitcoin-core/gui-qml/pull/430#discussion_r1829543225), unix sockets support will be addressed in a follow-up once the feature is ready in this repo.

src/node/interfaces.cpp
pablomartin4btc: I'm planning to refactor the `init.cpp` file after adding the persistence for the proxy setting, so the logic aligns with how the value is read and parsed from the `settings.json` file.

pablomartin4btc:
<ins>_Updates_</ins>:
- Addressed @vasild's feedback; next follow-ups:
  - `init.cpp` [refactoring](https://github.com/bitcoin-core/gui-qml/pull/430#discussion_r1826541696) will be done when the setting persistence is added in a separate PR.
  - unix socket support [related changes](https://github.com/bitcoin-core/gui-qml/pull/430#discussion_r1829543225) will be added in a separate PR once the feature is refreshed from the upstream repo.

pablomartin4btc:
<ins>_Just to clarify_</ins>: 
- this PR is ready for further review as the follow-ups mentioned in the previous [updates](https://github.com/bitcoin-core/gui-qml/pull/430#issuecomment-2458722893) are going to be implemented after this one is merged.

src/qml/controls/IPAddressValueInput.qml
vasild: After playing with this a bit and reading https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Character_class#non-v-mode_character_class I think it should be:

```sh
/^[\][0-9a-f.:]*$/i
```

```js
#!/usr/bin/env node

if ("[A]:2.3b".match(/^[\][0-9a-f.:]*$/i)) {
    console.log("match");
} else {
    console.log("no match");
}
```

Do we want to accept the empty string as valid? If no, then change the `*` to `+`.

src/qml/controls/IPAddressValueInput.qml
```diff
@@ -16,9 +16,9 @@ TextInput {
     property bool validInput: true
     enabled: true
     state: root.parentState
-    validator: RegExpValidator { regExp: /[0-9.:]*/ } // Allow only digits, dots, and colons
+    validator: RegularExpressionValidator { regularExpression: /^[][0-9A-Fa-f.:]*$/ } // Allow only IPv4/ IPv6 chars
 
-    maximumLength: 21
+    maximumLength: 45
```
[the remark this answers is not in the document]
pablomartin4btc: I'll correct it. Thanks!

src/qml/controls/IPAddressValueInput.qml
pablomartin4btc: Ok, let me check that again. Thanks!

pablomartin4btc: As commented to you offline, current implementation of the regex works as expected and doesn't allow spaces, it also works on a terminal with grep.

But I tried both current and your suggestion on https://jsconsole.com/, current implementation doesn't work well there but your does so I'm deciding to use yours (which includes `i` flag that allows case-insensitive); regarding `*` and `+`, it doesn't matter as the validation is being triggered anyways but for consistency and since it works in the `jsconsole`, I'll use `+` too.

pablomartin4btc:
<ins>_Updates_</ins>:
- Addressed @vasild's latest feedback:
  - Corrected max length of the IP input as [suggested](https://github.com/bitcoin-core/gui-qml/pull/430#discussion_r1836636887).
  - Updated regex more [in accordance](https://github.com/bitcoin-core/gui-qml/pull/430#discussion_r1837230383) with JS.
- Updated top description with test instructions suggested offline by @jarolrod.

vasild (approved):
ACK ebabe49eeb48d7579d4465d0f709839f0efe9fff

I am not very familiar with the QML side of things, but this change looks ok to me.

Good to have in a followup: unify the checks for valid proxy from `init.cpp` and from the GUI: https://github.com/bitcoin-core/gui-qml/pull/430#discussion_r1826541696

pablomartin4btc:
> Good to have in a followup: unify the checks for valid proxy from `init.cpp` and from the GUI: [#430 (comment)](https://github.com/bitcoin-core/gui-qml/pull/430#discussion_r1826541696)

Thanks @vasild for taking the time to review this PR! I'm working already on the next PR which will have both the persistence logic of the proxy and the refactoring of the `init.cpp` you suggested.

D33r-Gee (commented):
tACK [ebabe49](https://github.com/bitcoin-core/gui-qml/commit/ebabe49eeb48d7579d4465d0f709839f0efe9fff) works as expected on WSL Ubuntu 22.04

src/node/interfaces.cpp
pablomartin4btc: > I am not sure how relevant the first check is for the proxy because the second check ends up calling SplitHostPort() as wel

I’ll restore the call to `SplitHostPort()` as advised offline. This is necessary because, without it, `LookupNumeric` would apply a default port if none is provided, making an address entered without a port (like `127`) appear valid. I discovered this while testing the persistence of values in `settings.json` (a separate PR, which is nearly finished and will be pushed ASAP). For example, if a user enters `127` in the UI, it’s stored in the file without a port. While adding a default port could work, it might not be transparent for the user.

Another point we could discuss, if necessary, in the "setting persistence" PR (coming soon today) is why an IP like 127 remains valid even when a port is specified. This is because, as discussed with @vasil during his review of https://github.com/bitcoin-core/gui/pull/836, the IP validation is left to the OS (thru CService. For example, `$ ping 127` results in `PING 127 (0.0.0.127) 56(84) bytes of data`, and this behavior aligns with how Qt and `bitcoind` currently handle IPs.

vasild (approved):
ACK 4dea7c91a537710f3582d57b44f8908767ebb944

MarnixCroes (commented):
4dea7c91a537710f3582d57b44f8908767ebb944

the layout breaks when you paste when the default value (127.0.0.1:9050) is still present:
Screenshot from 2024-11-25 10-24-25

this doesn't happen when first removing the default value and then pasting.

Maybe add scroll functionality to this page?
If you're not on full screen and paste and get the invalid message, the bottom of the page cannot be accessed anymore as you can't scroll down.
On desktop switching to full screen works, but it might be a bigger problem on mobile.
So simple scroll functionality is good to add to the proxy settings page.

pablomartin4btc:
Thanks @MarnixCroes.

As discussed in our call among the team, both the overlapping and when the validation occurs (reverting `onTextChanged` into `onEditingFinished` would need a "done" button at the top - _opposed to the "<back" button_ - as @GBKS proposed) will be done on follow-ups.

image

GBKS:
Check out the [updated web prototype](https://lively-kashata-cfde7e.netlify.app/screen/proxy-settings) to see for a proposal of how this logic could work out.

pablomartin4btc:
> Check out the [updated web prototype](https://lively-kashata-cfde7e.netlify.app/screen/proxy-settings) to see for a proposal of how this logic could work out.

I really like it, also the fact that the default is greyed out and it disappears when the user starts typing instead of having to delete it (I think it behaves like that currently) . Many thanks!

src/node/interfaces.cpp
```diff
@@ -169,6 +170,27 @@ class NodeImpl : public Node
     }
     void mapPort(bool use_upnp, bool use_natpmp) override { StartMapPort(use_upnp, use_natpmp); }
     bool getProxy(Network net, Proxy& proxy_info) override { return GetProxy(net, proxy_info); }
+    std::string defaultProxyAddress() override
+    {
+        return std::string(DEFAULT_PROXY_HOST) + ":" + ToString(DEFAULT_PROXY_PORT);
+    }
+    bool validateProxyAddress(const std::string& addr_port) override
```
johnny9, line 177: validateProxyAddress shouldn't be in this module. It should exist in the nodemodel or possibly in netbase.

src/node/interfaces.cpp
```diff
@@ -169,6 +170,27 @@ class NodeImpl : public Node
     }
     void mapPort(bool use_upnp, bool use_natpmp) override { StartMapPort(use_upnp, use_natpmp); }
     bool getProxy(Network net, Proxy& proxy_info) override { return GetProxy(net, proxy_info); }
+    std::string defaultProxyAddress() override
```
johnny9, line 173: I don't believe this should return a string. I think it should be an in/out Proxy&.

src/interfaces/node.h
```diff
@@ -126,6 +129,12 @@ class Node
     //! Get proxy.
     virtual bool getProxy(Network net, Proxy& proxy_info) = 0;
 
+    //! Get default proxy address.
+    virtual std::string defaultProxyAddress() = 0;
```
johnny9, line 133: Are these actually default settings values or just values presented to the user?

pablomartin4btc: These are proper default values. I would leave them here for now, but maybe they can be moved into the netbase later.

src/node/interfaces.cpp
pablomartin4btc: This is just for display purposes, it doesn't make sense to build the object which also involves `CService` and so on, just to come back with eg `ipv4.proxy.ToStringAddrPort();`, I'd leave it as it is but can be done on a follow-up if more ppl think the change is required.

src/node/interfaces.cpp
pablomartin4btc: I agree, maybe it could be moved deeper into de netbase but I didn't what to touch that at this stage, it can be done later once we update the interfaces upstream.

johnny9 (approved):
ACK 4dea7c91a537710f3582d57b44f8908767ebb944

GBKS:
Is my ACK needed on this? Looks more like a code task to me than a UI one.

PR #435, opened 2024-11-22
Initial Template for Request Payment page

johnny9:
These commits introduce the Request Payment flow. This view is available under the "Receive" tab on the Desktop Wallets main navigation view. 

The Request Payment form itself currently only includes fields currently supported by the Qt wallet (label, message, and amount). Follow up PR will add support for local notes. 

The Amount field in the Request form takes only valid Satoshi/Bitcoin amounts and supports flipping the units with the associated button toggle.  

When the Continue button is pressed the Request Confirmation page gets pushed to the stack will eventually reflect the fields inputted in the previous form backed by the output of the WalletModel result. In its current implemention it just has statically defined properties to check the visual layout. The page also contains an address field with the copy icon that can be used to move the address into your clipboard.

Ubuntu screenshots


johnny9:
The piece I'm most concerned about this form is the Amount input field and how to properly sanitize and convert the field. The current object backing it doesnt quite feel right but I am limited by the ability to extend the actual TextInput class.

src/qml/bitcoinamount.cpp
```diff
@@ -0,0 +1,127 @@
...
+    case Unit::SAT: return 0;
+    } // no default case, so the compiler can warn about missing cases
+    assert(false);
+}
+
+QString BitcoinAmount::sanitize(const QString &text)
+{
+    QString result = text;
+
+    // Remove any invalid characters
```
MarnixCroes, line 29: Is this the right thing to do?
QT gui simply refuses to paste when there are invalid characters.

MarnixCroes:
> The piece I'm most concerned about this form is the Amount input field and how to properly sanitize and convert the field. The current object backing it doesnt quite feel right but I am limited by the inability to extend the actual TextInput class.

yeah there are a couple weird things:
- decimals when having sats as unit
- unnecessary leading zeroes when switching between sats en BTC. i.e. it can display things like `000100` sats

johnny9:
From  from 1b869eb to adff8f
 - Rebased with main and used PageStack in RequestPayment.qml
 - Fixed lint issue
 - Fixed mac build issue with missing bitcoinamount.h in Makefile

GBKS:
Looking good. I know we talked about this before, that the basic concept of the receive page has changed a bit. Away from the two separate pages, and towards creating the QR code directly on the page. Just realized I already implemented this in the web prototype [here](https://lively-kashata-cfde7e.netlify.app/screen/receive) (note the differences in responsive layouts).

The amount input logic, which you bring up, is not properly implemented in the web prototype. There definitely has to be some sanitation. Something to look out for is that when I type `5000 sats` and switch to `₿`, it should then show `0.00 005 000 ₿` and not `5000 ₿`. There should be a conversion happening. If this is too messy right now, I'd just come back to it in a separate PR.

In a web prototype update I am working on, I am tying the unit toggle to the application setting. So when you switch the unit in the amount input, the wallet balance display in the top left also changes. Wonder if that is intuitive or not. What do you think? In the send page, that unit setting should at least live on a user flow level. So when you toggle, the unit in the amount input, but also the fee selector changes, as well as the unit in the review screen.

Small thing about the input and amount fields is that I can enter unlimited characters (and emoji). Should probs be restricted in some way.

johnny9:
> Looking good. I know we talked about this before, that the basic concept of the receive page has changed a bit. Away from the two separate pages, and towards creating the QR code directly on the page. Just realized I already implemented this in the web prototype [here](https://lively-kashata-cfde7e.netlify.app/screen/receive) (note the differences in responsive layouts).
> 

I will remove the confirmation page and get the request page in alignment with the latest. I won't be able to get QR codes functioning in this PR as that will have to be a follow up once the requests are working end-to-end.

> Small thing about the input and amount fields is that I can enter unlimited characters (and emoji). Should probs be restricted in some way.

I'll double check this. I've made some improvements with the amount but I havent tried emojis

GBKS:
Looks great. I know you want to exclude various interaction details and handle those in follow-ups, so I won't comment on those. Just one question about the layout. Does this already include the mobile one where the QR code is inline and the buttons are static at the bottom?

GBKS:
Does this need more work, or could it be merged after a rebase?

johnny9:
update from [622a114](https://github.com/bitcoin-core/gui-qml/commit/622a11401d781d13c33fb0f86d5fafacd22f9bf0) to [efe0986](https://github.com/bitcoin-core/gui-qml/commit/efe098656d432cb9ec0c8e12137a65891486aa4c)
 - Rebased with main

GBKS:
@johnny9 thanks of rebasing. What state is this PR in? From our conversation last week, I think you said that you'd like to get this basic page merged even if it is not 100% yet, and then have smaller PRs to iron out the kinks, right?

pablomartin4btc (commented):
ACK efe098656d432cb9ec0c8e12137a65891486aa4c

Tested the PR manually and also checked it with @GBKS's [prototype](https://lively-kashata-cfde7e.netlify.app/screen/receive).

@johnny9 please, for future reference, try to keep the description of the PR concise if possible and updated (eg confirmation page removal) as a reviewer that, could be different approaches, starts from the top it could be misleading or confusing.

- Look & feel

  @GBKS, I find the layout a bit unbalanced in terms of alignment, but mainly I'm not sure about all the requested fields as in other wallets the process is more straight forward and simplified with just a QR and as you have in the prototype on the top-right margin menu ("...") the user could add the amount (and labels) later so I'd keep it simpler in that sense. This (having the amount (and labels) first, as in your prototype or this PR) or having the QR first could be configurable as I think most times a user won't require an amount but perhaps as in a store the user will put the amount (and labels) every time(?) and we could add that flexibility.
  <details>
  <summary>Ubuntu's screenshot.</summary>
  
  Screenshot from 2025-02-03 12-10-28
  </details>

  When we go back and forth from B to Sats, do we want to complete with 0's all the time? (eg 2.3 B to Sats = 230000000 then back to B shows 2.30000000).
  
  Do we need the payment order or id (#)? Is this going to be shown in history somehow? In current QT I think we could remove these records.
  <details>
  <summary>QT's screenshot.</summary>
  
  image
  </details>

- Code wise
  - I think there's an unattended [feedback](https://github.com/bitcoin-core/gui-qml/pull/435/files#r1854571659) above.
  - For the `BitcoinUnits` conversions we'll need to have different converters and they'll have the responsibility of conversion and not the amount itself, this could be added later on a follow-up.
  - In terms of commits, I think there are too many, the added icons ones could be all into 1 instead of 3 different and similarly the ones adding, merging and removing the same page (eg confirmation) could be also consolidated. I understand the chronological order but a more streamlined commit history would make it easier to review and follow the code.

GBKS:
> Look & feel

@pablomartin4btc please see the rationale in the design docs [here](https://bitcoincore.app/milestones/1-6-receive/). Great to hear your feedback, and I'd like to start tracking feedback like this, but wait with acting on it after the MVP release and we have had more chance to actually use it and do user testing.

johnny9:
Update from [efe0986](https://github.com/bitcoin-core/gui-qml/commit/efe098656d432cb9ec0c8e12137a65891486aa4c) to [9ad34f0](https://github.com/bitcoin-core/gui-qml/commit/9ad34f0aa2595df6f72c8b3d1b088e985f425e5a)

- Rebased with main

johnny9:
Update from [cb6d96a](https://github.com/bitcoin-core/gui-qml/commit/cb6d96ae8f9081a3ff101fd0739dbd374a58430e) to [13535ca](https://github.com/bitcoin-core/gui-qml/commit/13535cab4b7534f346bd0b3d0716e882fc362490)
 - Rebased with main and squashed RequestPayment form commits
 -

src/qml/pages/wallet/RequestPayment.qml
```diff
@@ -0,0 +1,239 @@
...
+            anchors.horizontalCenter: parent.horizontalCenter
+            spacing: 30
+            ColumnLayout {
+                id: columnLayout
+                Layout.minimumWidth: 450
+                Layout.maximumWidth: 470
+
+                spacing: 5
+
+                Item {
```
davidgumberg, line 49: nit: Maybe amount input deserves it's own control?

src/qml/pages/wallet/RequestPayment.qml
```diff
@@ -0,0 +1,239 @@
...
+
+                    height: amountInput.height
+                    Layout.fillWidth: true
+                    CoreText {
+                        id: amountLabel
+                        width: 110
+                        anchors.left: parent.left
+                        anchors.verticalCenter: parent.verticalCenter
+                        horizontalAlignment: Text.AlignLeft
+                        color: Theme.color.neutral9
```
davidgumberg, line 62: nit: should be `neutral7` like the other labels

src/qml/bitcoinamount.cpp
davidgumberg: maybe a question for designers: but I agree that trying to input invalid characters resulting in nothing happening is not helpful to users, but as a scaffold, the current approach seems good enough for now

src/qml/pages/wallet/RequestPayment.qml
```diff
@@ -0,0 +1,239 @@
...
+                    id: continueButton
+                    Layout.fillWidth: true
+                    Layout.topMargin: 30
+                    text: qsTr("Create bitcoin address")
+                    onClicked: {
+                        if (!clearRequest.visible) {
+                            requestCounter = requestCounter + 1
+                            clearRequest.visible = true
+                            title.text = qsTr("Payment request #" + requestCounter)
+                            address.text = "bc1q f5xe y2tf 89k9 zy6k gnru wszy 5fsa truy 9te1 bu"
```
davidgumberg, line 201: Not likely, and inadvisable behavior, but it seems to me like a bad idea to put a placeholder address here when someone curious might build this branch and try to use it with mainnet funds, maybe insert an invalid character or something else to indicate this is a placeholder?

hebasto: cc @GBKS

src/qml/bitcoinamount.cpp
hebasto: @GBKS

hebasto (approved):
ACK 13535cab4b7534f346bd0b3d0716e882fc362490, I have reviewed the code and it looks OK.

Unaddressed nits can be resolved in follow-ups after receiving input from designers.

src/qml/pages/wallet/RequestPayment.qml
johnny9: Yes, it will be. I haven't committed to doing that just yet as I don't have a clear idea of exactly how the BitcoinAmounts and the Bitcoin Address inputs will be handled. I think the PR to do the input formatting of the amounts should take care of having this component extracted out with clear signals/properties.

src/qml/pages/wallet/RequestPayment.qml
johnny9: I'll double check this. There has been some flip flopping on the design side but it should be decided now

PR #441, opened 2025-01-21
Rename Block Clock to Block Status

johnny9:
These changes rename all instances of "Block Clock" to "Block Status" under the qml directory.


MarnixCroes (commented):
build fails 
```make[1]: *** No rule to make target 'qml/components/blockclockdial.cpp', needed by 'qml/components/qt_libbitcoinqt_a-blockclockdial.o'.  Stop.```

**Comment by MarnixCroes** [comment]
here still mentioned?

https://github.com/johnny9/gui-qml/blob/11a589a61fecd5462b7b8deea69e621f0410c80e/src/qml/components/BlockStatus.qml#L184

and in `blockstatusdial.h`

**Review by ciegovolador — APPROVED** [review]
ACK

**Review by pablomartin4btc — COMMENTED** [review]
ACK on removing "Clock".

Just thinking it loud, shouldn't be more like "Sync" status? (moreover to avoid confusion with an individual block status).

Also found this in current QT desktop app source (`src/qt/clientmodel.h`):

```
enum class SyncType {
    HEADER_PRESYNC,
    HEADER_SYNC,
    BLOCK_SYNC
};

```

Maybe we need something like that later to identify on which sync we are...

hebasto:
Needs rebase.

hebasto:
Closing for now.

PR #444, opened 2025-02-18
Implement createSingleSigWallet

davidgumberg:
Although imperfect, this PR enables completion of the wallet creation flow by implementing `createSingleSigWallet`.  Useful for testing #418.


johnny9 (approved):
tACK be612b03a60a100ceac4919020ce55a96ea49bec

tested on Ubuntu with "-signet -resetguisettings"

Thanks for fixing this. Looks like this method somehow got lost when we moved the module.

PR #446, opened 2025-03-04
[DRAFT] WIP QML Generate Snapshot Signet

D33r-Gee:
Based on #424. For evaluation purposes only!

# GUI Integration for Generating UTXO Snapshots

## Overview
This PR adds initial GUI support for generating a `signet` UTXO snapshot at height 160,000, building on Bitcoin Core's `assumeutxo` infrastructure.

## What This PR Does

1. Adds a basic GUI interface for generating UTXO snapshots via the Connection Settings panel
2. Implements non-blocking snapshot generation via QT Thread
3. Provides visual confirmation of snapshot generation
4. Adds a snapshot progress notifications accessible via a handler

### Core Components Modified

1. **Node Interface** (`src/node/interfaces.cpp`)
- Implements a notification handler for snapshot generation progress (rewind and restore)

2. **Snapshot Qml** (`src/qml/models/snapshotqml.h`)
- Introduces a worker class to manage snapshot generation

3. **QML Integration** (`src/qml/models/nodemodel.cpp`)
- Manages snapshot generation state and progress
- Uses Qt's thread system to prevent UI blocking
- Provides progress updates via Qt signals

4. **UI Components** (`src/qml/components/SnapshotGenSettings.qml`)
- File view for snapshot generation
- Snapshot generation progress and completion

### Key Design Decisions

1. **Non-blocking Implementation**
- Snapshot generation runs in separate QT thread
- UI remains responsive during generation
- Node is disconnected from the network during generation

2. **Extensibility**
- Interface designed to accommodate future `assumeutxo` changes
- Error handling framework in place

## Testing Instructions

1. Build
2. Launch Bitcoin Core QML GUI
3. Navigate to Connection Settings
4. Wait for IBD to complete
5. Click "Generate Snapshot"
6. Verify (see screenshots below):
   - "Snapshot Generated" confirmation appears
   - snapshot file is created in the `signet` data directory

<details>

Screenshot 2025-03-04 115218
Screenshot 2025-03-04 115246
Screenshot 2025-03-04 115309
Screenshot 2025-03-04 115323
Screenshot 2025-03-04 115356
Screenshot 2025-03-04 115641
Screenshot 2025-03-04 gen file view

</details>

### Expected Behavior
- File selection dialog works
- Success/failure state properly displayed
- Node is disconnected from the network during generation

## Future Work

1. **Cancel Snapshot Generation**
- Add ability to cancel snapshot generation

2. **Integration**
- Coordinate with upstream `assumeutxo` changes
- Add support for network-specific snapshots (`testnet` and `mainnet`)

## Notes for Reviewers

 **Snapshot Compatibility**
- Currently only works with `signet`
- Limited to specific 160,000 height
- Validates against chain parameters [`m_assumeutxo_data`](https://github.com/bitcoin-core/gui-qml/blob/f23ffb874cc5fe583b97f8fec9caff87fcf42f67/src/kernel/chainparams.cpp#L374)
(the changes there can be discarded once synced since [`m_assumeutxo_data`](https://github.com/bitcoin/bitcoin/blob/17834bd1976df7a2ff6c2f5f05a59ae3fd3f6875/src/kernel/chainparams.cpp#L499) is already hardcoded)

This is a work in progress - feedback welcome on the approach and implementation details.


yashrajd:
Great start! Thanks for continuing to do this man. I have not tested on my machine, responding to screenshots below.

> 
> ## What This PR Does
> 1. Adds a basic GUI interface for generating UTXO snapshots via the Connection Settings panel
> 2. Implements non-blocking snapshot generation via QT Thread
> 3. Provides visual confirmation of snapshot generation
> 4. Adds a snapshot progress notifications accessible via a handler
> 
> ...
> 
> 4. **UI Components** (`src/qml/components/SnapshotGenSettings.qml`)
> 
> * File view for snapshot generation
> * Snapshot generation progress and completion
> 
> ### Key Design Decisions
> 1. **Non-blocking Implementation**
> 
> * Snapshot generation runs in separate QT thread
> * UI remains responsive during generation
> * Node is disconnected from the network during generation
> 
> 2. **Extensibility**
> 
> * Interface designed to accommodate future `assumeutxo` changes
> * Error handling framework in place
> 

Top-tier PR write-up, kudos!


> Screenshot 2025-03-04 115309

My main gripe still is that 'load/generate snapshot' makes no sense in Connection Settings. Otherwise looks good.

Screenshot 2025-03-04 115323

> "A snapshot captures the current state of bitcoin transactions on the network. It can be imported into other bitcoin nodes to speed up the initial setup. You can generate a snapshot of the current chain state.

Isn't the snapshot generated a pre-decided block height only? not the current block height?

- if so, we should mention the date/block height in the text.
- we should also update the text so as to avoid giving wrong impression
- long-standing issue with reference to transactions

Suggested copy: **"A snapshot captures the state of the bitcoin network up to a certain date in the recent past. It can be imported into other bitcoin nodes to speed up their initial sync.
Snapshot date: <date> (block <block number>)**

Screenshot 2025-03-04 115356


Perhaps the "Rewinding" text is placeholder. Should be something like:
"This might take a while.

You can continue using the application while the process continues in the background."

Good to have would be to show some a % progress number

Screenshot 2025-03-04 115641


The View File button is awesome! Consider adding a Done button too that will send the user back to home screen/block clock...

D33r-Gee:
> My main gripe still is that 'load/generate snapshot' makes no sense in Connection Settings. Otherwise looks good.

in the [figma design](https://www.figma.com/design/ek8w3n3upbluw5UL2lGhRx/Bitcoin-Core-App-Design?node-id=8845-13590&t=bnYR7y34YtIkO72T-0) where would you idealy prefer these options would be placed under?

D33r-Gee:
> Isn't the snapshot generated a pre-decided block height only? not the current block height?
> 
> * if so, we should mention the date/block height in the text.
> * we should also update the text so as to avoid giving wrong impression
> * long-standing issue with reference to transactions

Great suggestions will update with the next push

D33r-Gee:
> Good to have would be to show some a % progress number

Could you share some design ideas for this?

D33r-Gee:
> Consider adding a Done button too that will send the user back to home screen/block clock...

Could you update the [figma](https://www.figma.com/design/ek8w3n3upbluw5UL2lGhRx/Bitcoin-Core-App-Design?node-id=8845-13590&t=bnYR7y34YtIkO72T-0) to reflect that?

D33r-Gee:
@GBKS could you take a look at the current iteration of the design and user flow, and le me know your thoughts?

GBKS:
Nice progress here. I just tried to test and just see the message that IBD is not complete:

image

However, the block status tells me it's all synced, as well as the CLI:

image

The signet explorer also showed this block as the latest one. Do you maybe know what the problem might be?

---

On a different note, my assumption was that we show either snapshot import (during IBD) **or** snapshot creation (post-IBD). We only need one at a time, right?

D33r-Gee:
> However, the block status tells me it's all synced, as well as the CLI:

Thanks for testing. Ok going to take a look I think I know what the issue is

> On a different note, my assumption was that we show either snapshot import (during IBD) or snapshot creation (post-IBD). We only need one at a time, right?

Yep that's correct. That said, in terms of display if the user has loaded a snapshot my assumption is that we still want to display that, is that correct? (see screenshot below)

<details>
  <summary>Connection Settings Snapshot Loaded</summary>

Screenshot 2025-03-26 093610

</details>

yashrajd:
> Yep that's correct. That said, in terms of display if the user has loaded a snapshot my assumption is that we still want to display that, is that correct? (see screenshot below)

Makes sense.

On the flip side, if they have not loaded a snapshot, the Load Snapshot text has to be greyed out. The green check icon should not be present. The description part could read "Cannot load a snapshot because sync is already completed"

or

of-course we can always just not show the Load Snapshot list item, that but I personally prefer having it.

D33r-Gee:
with [fd78f7e](https://github.com/bitcoin-core/gui-qml/commit/fd78f7e2592a10f0716ebf54d96646a17c8d566d) rebased and addressed @GBKS feedback (please see screenshots below)

<details>
  <summary>Generate Button appears after IBD completion</summary>
  
Screenshot 2025-03-27 104639


</details>

Next: will address the use case that @yashrajd mentioned...

> On the flip side, if they *did not load a snapshot during IBD, the green check icon should not be present. (the Load Snapshot text has could be greyed out too). The description part could read "Cannot load a snapshot because sync is already completed".

GBKS:
Let's clarify the situation around having an imported snapshot. Is there any user benefit to showing that a snapshot was imported, after the import is done? Of course, during the import, the user wants to see progress and at the end they want to see a confirmation that it was completed successfully. But after that, is there any reason to keep this note around?

D33r-Gee:
> Let's clarify the situation around having an imported snapshot. Is there any user benefit to showing that a snapshot was imported, after the import is done? Of course, during the import, the user wants to see progress and at the end they want to see a confirmation that it was completed successfully. But after that, is there any reason to keep this note around?

Great questions, going to let @yashrajd respond first... then will share my thoughts

yashrajd:
> Let's clarify the situation around having an imported snapshot. Is there any user benefit to showing that a snapshot was imported, after the import is done? Of course, during the import, the user wants to see progress and at the end they want to see a confirmation that it was completed successfully. But after that, is there any reason to keep this note around?

I don't have a strong preference as I said earlier, but here's why I think it should be there:

- Christoph's description above reminds me of in-app CTAs. Interface elements can take up valuable space (say CTAs on Home Screen), so it's natural to get rid of them once they're not relevant. But this is a Setting item deep in the navigation, and I'm not aware of a pattern where we remove settings that are not applicable.

- One pattern we do see is Settings items that also serve a communication or marketing purpose but are not actionable (Proton VPN and Signal do this IIRC) are still displayed as a way of informing the user that the app has this feature.

- Though this scenario is a bit different, if Load snapshot has been used then it serves to communicate that fact. If it hasn't, the item communicates that the feature exists. After all, the generated snapshot is intended to be used somewhere else, and the presence of this item communicates that and indicates to the user where to load generated snapshot on their other machine.

D33r-Gee:
Thanks @GBKS and @yashrajd for the feedback.

<details>
  <summary>So got a compromise in the form of 3 potential use cases:</summary>

1.  **The user's node has just started, therefore the snapshot load option is available (visible)**

 
Screenshot 2025-03-31 093125

2. **The user has loaded a snapshot:**

Screenshot 2025-03-31 093210

3. **the user's node has reached full sync (IBD complete) without a snapshot:**

Screenshot 2025-03-31 093245
 
</details>

Please let me know your thoughts?

GBKS:
> Christoph's description above reminds me of in-app CTAs. Interface elements can take up valuable space (say CTAs on Home Screen), so it's natural to get rid of them once they're not relevant. But this is a Setting item deep in the navigation, and I'm not aware of a pattern where we remove settings that are not applicable.
> 
> One pattern we do see is Settings items that also serve a communication or marketing purpose but are not actionable (Proton VPN and Signal do this IIRC) are still displayed as a way of informing the user that the app has this feature.
> 
> Though this scenario is a bit different, if Load snapshot has been used then it serves to communicate that fact. If it hasn't, the item communicates that the feature exists. After all, the generated snapshot is intended to be used somewhere else, and the presence of this item communicates that and indicates to the user where to load generated snapshot on their other machine.

@yashrajd we're not removing anything and it's not a marketing thing. There is a single item for this feature that is always visible and shows the appropriate state. If you can generate a snapshot, it shows that. If you can't, it shows that you can import a snapshot. It's one or the other, or am I missing something?

yashrajd:
TL;DR: IMO it's just a couple of list items that don't take up prominent or important space. They can serve multiple functions (as actionable elements, show state: enabled/disabled/progress) and we always display both, it even serves as marketing/education/communication. Happy to mockup different states if @D33r-Gee  needs it.

> Let's clarify the situation around having an imported snapshot. Is there any user benefit to showing that a snapshot was imported, after the import is done? Of course, during the import, the user wants to see progress and at the end they want to see a confirmation that it was completed successfully. But after that, is there any reason to keep this note around?

my impression is that you are saying: the Load Snapshot list item should be removed after snapshot has been imported. sorry if I misunderstood you & don't have to read the stuff below
.
.
.
----------

> @yashrajd we're not removing anything and it's not a marketing thing. There is a single item for this feature that is always visible and shows the appropriate state. If you can generate a snapshot, it shows that. If you can't, it shows that you can import a snapshot. It's one or the other, or am I missing something?

actually it's not just one or the other...it's never BOTH afaik but there's cases where it's NEITHER: if you have a pruned node,  you cannot generate a snapshot. Do we not show the Generate Snapshot list item in that case? Users can un-prune their nodes btw. Users can also delete their bitcoin folders in which you have to do IBD again & want to load a snapshot.

With your proposed approach (if I understood you correctly), content/list-items will appear and disappear as situation changes... I don't think the user wants/expects Settings items to do that.

D33r-Gee:
> Happy to mockup different states if @D33r-Gee needs it.

Yes that would be helpful to visualize the different use cases/states.

D33r-Gee:
with [d7523cd](https://github.com/bitcoin-core/gui-qml/commit/d7523cde16a6805cbe2689b9b7f3991539116c63) updated both the snapshot loading and snapshot generation pages to reflect @GBKS video (thanks for providing that) and thanks @yashraj for meeting and providing clarification.

Please let me know your thoughts...

D33r-Gee:
[1eaec25](https://github.com/bitcoin-core/gui-qml/commit/1eaec25840dea41de041637f38f20c5edf94f663) rebased over main

hebasto:
Closing for now.

PR #447, opened 2025-03-10
Fix handling of load wallet in controller

johnny9:
Don't load wallet if we already have them loaded.

There are 3 changes to the original code. In setSelectedWallet there is a proper check to see if there's already an instance of the wallet loaded in our m_wallets list and sets selected_wallet to that if true, It only loads the wallet into the list once (handleLoadWallet gets called when doing loadWallet), and it fixes the initialize to not call handlLoadWallet and emit selectedWalletChanged a bunch of times.

### DIFF
--- a/src/qml/walletqmlcontroller.cpp
+++ b/src/qml/walletqmlcontroller.cpp
@@ -4,6 +4,8 @@
 
 #include <qml/walletqmlcontroller.h>
 
+#include <qml/models/walletqmlmodel.h>
+
 #include <interfaces/node.h>
 #include <support/allocators/secure.h>
 #include <wallet/walletutil.h>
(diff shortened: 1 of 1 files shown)


johnny9:
> > Don't load wallet if we already have them loaded.
> 
> What scenarios are you referring to?

This is the QList member of the walletqmlcontroller. Previously it wasn't properly checking if we had the wallet in the list already before calling loadWallet and storing the unique_ptr in this list. You'll see the check at the top of setSelectedWallet now. The other change is that we don't use the unique_ptr after loadWallet, instead the handleLoadWallet method is used to store the resulting wallet. handleLoadWallet gets called anytime a wallet is loaded either from calling that interface method, rpc load, or at startup.

johnny9 (commented):
Tested on Ubuntu desktop. 
1. With multiple wallets created and all of them unloaded, selected each wallet to test that it doesn't crash.
2. With the wallets previously loaded, start the application and select each wallet in the select to test that it doesn't crash
3. Started the application with a datadir that contains no wallets to validate initialization handles the empty case properly

johnny9:
@jarolrod

PR #448, opened 2025-03-25
Introduce Coin Selection page

johnny9:
This PR depends on #445 

This is the first iteration of the Coin Selection menu

The option to open the Coin Selection menu is a QSetting that is toggled on/off in the ellipsis menu on the Send page. When enabled, the user can click on the Select inputs button and the Coin Selection menu will be push on the stack from below. The Coin Selection menu will list possible inputs and a checkbox is used to select the coins.

https://github.com/user-attachments/assets/f148d231-d9e9-4923-91f3-8e422049bbf5

PR #450, opened 2025-05-08
Add Multiple Recipients option to the Send form

johnny9:
This change introduces the multiple recipients controls to the Send form. It is enabled in the ellipses option menu by toggling on "Enable Multiple Recipients". This is stored as a QSetting for the user.

This PR depends on #448 which contains the

Screenshot from 2025-05-08 11-51-34
 first implementation of the Send options menu.
Screenshot from 2025-05-08 11-51-26


MarnixCroes (commented):
b4e421e19dbbf564431a57319e88630edffb820b

- the amount is not tied to each recipient. If I enter x at 1, all have amount x
- small nit, could this be centered
 
Screenshot from 2025-05-30 13-26-54

johnny9:
> [b4e421e](https://github.com/bitcoin-core/gui-qml/commit/b4e421e19dbbf564431a57319e88630edffb820b)
> 
>     * the amount is not tied to each recipient. If I enter x at 1, all have amount x
> 
>     * small nit, could this be centered
> 
> 
> Screenshot from 2025-05-30 13-26-54

I believe I have both of these issues resolved now with the latest commits.

johnny9:
Discovered an issue with the calculation of the total amount in the review ui and coin selection ui. Will resolve that today.

GBKS:
Looks great.

1. There seems to be a small issue with the removal logic.

- For recipients 1, 2, and 3, I respectively enter `a`, `b` and `c` into the `Send to` field
- With recipient 1 selected, I click `-` to remove the current recipient (as expected)
- A recipient is removed, I can tell from the header changing from `Recipient 1 of 3` to `Recipient 1 of 2`
- However, I still see `a` in the `Send to` field. I expected the current recipient to be removed and that I now would see `b`.
- I click `>` and now see `c` in the `Send to` field
- I click `<` and now see `b` in the `Send to` field

Not sure if that outline is easy to follow, but by guess is that the current recipient is correctly being removed, but the UI is not being refreshed.

2. There are some visual tweaks to be made, but those can be done separately.

3. Is there a limit on how many recipients can be added? Should we restrict it to something like 25 for now? Michael and I were wondering about that.

4. Also, should the `Note to self` be per recipient or for the whole transaction? In the prototype I have it per recipient (essentially an address label). But in this implementation it's the same for all recipients.

johnny9:
Update from [e5cd1a7](https://github.com/bitcoin-core/gui-qml/commit/e5cd1a7c23cabd66db18e7ce99bf4dfeb6ee593e) to [8a6b593](https://github.com/bitcoin-core/gui-qml/commit/8a6b593e54673c59c84c43fbdf642da81ce91e64)
 - rebased with main

johnny9:
> Looks great.
> 
>     1. There seems to be a small issue with the removal logic.
>     3. Is there a limit on how many recipients can be added? Should we restrict it to something like 25 for now? Michael and I were wondering about that.
> 

These two issues are now resolved

MarnixCroes (commented):
8a6b593e54673c59c84c43fbdf642da81ce91e64
_I found multiple things, probably some are to be fixed and some can be follow-up._

- Have 2 recipients, delete recipient 2. It stays at recipient 2 and displays `Recipient 2 of 1`:

Screenshot from 2025-06-04 09-30-31

- I noticed a delay in the coin selection `Remaining to select` / `Over required amount`, this does not happen on master

[Screencast from 2025-06-04 09-12-47.webm](https://github.com/user-attachments/assets/c5918670-e52f-47a0-9286-e63361b68b74)

- Have multiple recipients, one in sats and the other in BTC -> the Transaction Details screen is confusing as it doesn't have units:

Screenshot from 2025-06-04 09-14-57


**Some issues when having 2 or more recipients and disabling `Multiple Recipients` as the other disabled recipients are still taken into account:**
- After disabling multiple recipients, the Send view will show the current selected single recipient, but the amount in `Transaction Details` is the sum of the multiple recipients (even though disabled)

[Screencast from 2025-06-04 09-07-22.webm](https://github.com/user-attachments/assets/b4552376-7f8a-4b7d-b854-18228bb98e18)

- Recipient 2 is empty, disable `Multiple Recipients`. Send view is at recipient 1 (with address and amount) but `Review` doesn't work.

rabbitholiness:
I also noticed a few additional things to the ones mentioned above:

-  If I switch to Sats as a unit for the amount and then add a new recipient, the unit for that newly added recipient is changed to ₿. I would have expected it to also use Sats.
- Since the address is on one line, it gets cut off. I guess the[ multi-line view](https://lively-kashata-cfde7e.netlify.app/screen/send) just isn't implemented yet? 
Screenshot 2025-06-05 at 9 08 43

- If I use the same address for different recipients, then we should probably highlight that somehow. Based on how difficult it is to implement, there is a more generic way to just say "You're reusing this address for multiple recipients." But ideally we would say "This address is being reused for Recipient x and Recipient y." The names of the recipients could be a link to the corresponding form. Here's how that could look like. 
address reuse multiple recipients
- I can't get to the review screen, for some reason.
- I think the note to self should be per recipient. Especially since it should be used to identify the payments on the review screen (and on the transaction details page), eventually.

johnny9:
> [8a6b593](https://github.com/bitcoin-core/gui-qml/commit/8a6b593e54673c59c84c43fbdf642da81ce91e64) _I found multiple things, probably some are to be fixed and some can be follow-up._
> 
>     * Have 2 recipients, delete recipient 2. It stays at recipient 2 and displays `Recipient 2 of 1`:

Found the off by one error. It should work correctly now.

> 
> Screenshot from 2025-06-04 09-30-31
> 
>     * I noticed a delay in the coin selection `Remaining to select` / `Over required amount`, this does not happen on master

This is a similar fix to the validation. I've push the fix to this PR as well.

> 
> Screencast.from.2025-06-04.09-12-47.webm
> 
>     * Have multiple recipients, one in sats and the other in BTC -> the Transaction Details screen is confusing as it doesn't have units:

The transaction details I will create a new task/issue as a follow up as I would like to do a bunch of amount labels at once.

> 
> Screenshot from 2025-06-04 09-14-57
> 
> **Some issues when having 2 or more recipients and disabling `Multiple Recipients` as the other disabled recipients are still taken into account:**
> 
>     * After disabling multiple recipients, the Send view will show the current selected single recipient, but the amount in `Transaction Details` is the sum of the multiple recipients (even though disabled)
> 
> 
> Screencast.from.2025-06-04.09-07-22.webm
> 
>     * Recipient 2 is empty, disable `Multiple Recipients`. Send view is at recipient 1 (with address and amount) but `Review` doesn't work.

I now have it clearing out the list up to the first recipient when the setting is disabled and the amounts should be correct now.

GBKS:
Awesome progress. Not sure if it's covered elsewhere, here are some things I am seeing based on the latest code:

- I can enter infinite characters in the address, amount and note fields. Super long amounts end up overlapping the unit toggle and going outside of the form boundaries.
- I can enter letter (abc...) in the amount input, should be just numbers and comma/period.
- When enabling multiple recipients, it should automatically add a second one.
- In the prototype, I set up an application-wide setting for the unit format. So if you toggle it in the form for one recipient, it also toggles it for all other recipients, the balance in the top left, any other screen you visit, etc. As a mental model, that seems easier track than having individual components have their own setting.
- The disabled state for the pagination/add/remove buttons is super dark, it should be Neutral 4 instead of Neutral 2.
- When a transaction was sent, the `Multiple recipients` setting should also reset to disabled. Weird to be on the screen and having it read `Recipient 1 of 1`.

johnny9:
> Awesome progress. Not sure if it's covered elsewhere, here are some things I am seeing based on the latest code:
> 
>     * I can enter infinite characters in the address, amount and note fields. Super long amounts end up overlapping the unit toggle and going outside of the form boundaries.
> 
>     * I can enter letter (abc...) in the amount input, should be just numbers and comma/period.

These two are addressed in #462 

>     * When enabling multiple recipients, it should automatically add a second one.

This makes sense, will add
>     * In the prototype, I set up an application-wide setting for the unit format. So if you toggle it in the form for one recipient, it also toggles it for all other recipients, the balance in the top left, any other screen you visit, etc. As a mental model, that seems easier track than having individual components have their own setting.

I'll make an issue to track this for a follow up. I need to refactor some things to get the setting to be global.

>     * The disabled state for the pagination/add/remove buttons is super dark, it should be Neutral 4 instead of Neutral 2.

 Will fix.

>     * When a transaction was sent, the `Multiple recipients` setting should also reset to disabled. Weird to be on the screen and having it read `Recipient 1 of 1`.

Ill try to fix this as well.

src/qml/pages/wallet/MultipleSendReview.qml
```diff
@@ -0,0 +1,155 @@
...
+                }
+            }
+
+            Separator {
+                Layout.fillWidth: true
+                color: Theme.color.neutral3
+            }
+
+            ContinueButton {
+                id: confimationButton
```
MarnixCroes, line 144:
```suggestion
                id: confirmationButton
```

MarnixCroes (approved):
tested ACK 564231917b43bc0ab121efd90e7a70ec8761faef
LGTM for initial feature introduction

PR #454, opened 2025-05-20
Add QRImageProvider

goqusan:
The `QRImageProvider` can be used to display QR codes in QML `Image`.

The new control `QRImage` wraps the new provider in a convenient way.

The `RequestPayment` page is updated to showcase `QRImage` usage.

### DIFF
--- a/src/qml/qrimageprovider.cpp
+++ b/src/qml/qrimageprovider.cpp
@@ -1,0 +1,56 @@
+// Copyright (c) 2025 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include <qml/qrimageprovider.h>
+
+#if defined(HAVE_CONFIG_H)
+#include <config/bitcoin-config.h> /* for USE_QRCODE */
+#endif
+
+#ifdef USE_QRCODE
+#include <qrencode.h>
+#endif
+
+#include <QImage>
+#include <QQuickImageProvider>
+#include <QSize>
+#include <QString>
+#include <QUrl>
+#include <QUrlQuery>
+
+QRImageProvider::QRImageProvider()
+    : QQuickImageProvider{QQuickImageProvider::Image}
+{
+}
+
+QImage QRImageProvider::requestImage(const QString& id, QSize* size, const QSize& requested_size)
+{
+#ifdef USE_QRCODE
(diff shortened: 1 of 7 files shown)


src/qml/qrimageprovider.cpp
```diff
@@ -0,0 +1,56 @@
...
+    if (code) {
+      QImage image{code->width, code->width, QImage::Format_ARGB32};
+      unsigned char* p = code->data;
+      for (int y = 0; y < code->width; ++y) {
+          for (int x = 0; x < code->width; ++x) {
+              image.setPixelColor(x, y, (*p & 1) ? fg : bg);
+              ++p;
+          }
+      }
+      QRcode_free(code);
```
johnny9, line 47: Is it ok to free before using it for *size?

johnny9 (changes requested):
Concept ACK

Just some whitespace fix before approval

MarnixCroes (commented):
f445e6598b4afe6f1730950a97090f56ce2e8444

The QR-code is displayed and I'm able to scan it ✅
However, if I scan it with a mobile Bitcoin wallet it cannot process it?

johnny9:
> The QR-code is displayed and I'm able to scan it ✅ However, if I scan it with a mobile Bitcoin wallet it cannot process it?

Which wallet? I scanned with Zeus and it recognized it as an address to send to. I also checked sparrow and it also displays just a string containing the address in the QR code on the Receive tab.

johnny9 (approved):
tACK f445e6598b4afe6f1730950a97090f56ce2e8444

Tested on Ubuntu. Hard coded address  QR was recognized using phone camera.

GBKS:
Does the QR code contain a BIP21 URI? If so, the default OS camera should be able to open the wallet as needed (if the wallet has registered for the `bitcoin:` URI scheme).

MarnixCroes:
> > The QR-code is displayed and I'm able to scan it ✅ However, if I scan it with a mobile Bitcoin wallet it cannot process it?
> 
> Which wallet? I scanned with Zeus and it recognized it as an address to send to. I also checked sparrow and it also displays just a string containing the address in the QR code on the Receive tab.

I tried with Phoenix, Breez and Blue Wallet. Neither work.
Just tried with Zeus and that is the only one that works for me, indeed.

hebasto:
> Does the QR code contain a BIP21 URI?

No, it does not. Should be amended in a follow up, I presume.

PR #455, opened 2025-05-21
Add Skeleton loading animation to the Wallet selector

johnny9:
This adds the first skeleton loading section to the desktop wallet. Specifically the WalletBadge at the top left corner that is used for selecting a wallet.

https://github.com/user-attachments/assets/968c8da2-61af-459b-ad62-6b05c1ed7bb8

### DIFF
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -440,8 +440,9 @@ QML_RES_QML = \
   qml/controls/PageStack.qml \
   qml/controls/ProgressIndicator.qml \
   qml/controls/qmldir \
-  qml/controls/Setting.qml \
   qml/controls/SendOptionsPopup.qml \
+  qml/controls/Setting.qml \
+  qml/controls/Skeleton.qml \
   qml/controls/TextButton.qml \
   qml/controls/Theme.qml \
   qml/controls/ToggleButton.qml \
--- a/src/qml/bitcoin_qml.qrc
+++ b/src/qml/bitcoin_qml.qrc
@@ -50,6 +50,7 @@
         <file>controls/qmldir</file>
         <file>controls/SendOptionsPopup.qml</file>
         <file>controls/Setting.qml</file>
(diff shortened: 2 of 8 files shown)


johnny9:
https://github.com/user-attachments/assets/02294e33-c84b-428a-b420-6f86fc2234aa

Here is an attempt at the gradient shimmer

johnny9:
Update from [1f65d17](https://github.com/bitcoin-core/gui-qml/commit/1f65d17ec8d530eb2e0e08c2e5fc96a0eebcfba2) to [9a53889](https://github.com/bitcoin-core/gui-qml/commit/9a53889d63df5eb94119b279d66ed3dabfd59141)
 - Replaced the Animation in Skeleton with the gradient "shimmer" animation.

GBKS:
Looks great (from the video). Just built the code and tried to check it out in the app, but things load so quickly for me that I don't see it. Do I need to load it up with lots of transactions, or does it only show during IBD or so?

Small thing I noticed is that the content of the wallet display wiggles a bit horizontally when the name is shorter/longer than the balance. Could we please align things to the left?

johnny9:
Update from [9a53889](https://github.com/bitcoin-core/gui-qml/commit/9a53889d63df5eb94119b279d66ed3dabfd59141) to [99ca9fd](https://github.com/bitcoin-core/gui-qml/commit/99ca9fde8032177fb54c662cb929bc2ade790d94)
 - Removed semicolons in the qml
 - Removed useless anchor in Item

johnny9:
> Looks great (from the video). Just built the code and tried to check it out in the app, but things load so quickly for me that I don't see it. Do I need to load it up with lots of transactions, or does it only show during IBD or so?
> 
> Small thing I noticed is that the content of the wallet display wiggles a bit horizontally when the name is shorter/longer than the balance. Could we please align things to the left?
> 
> 

Created an issue to track this. It should be possible to fix the alignment.

PR #462, opened 2025-06-01
Add validation to the Send form (address and amount)

johnny9:
This is built on top of https://github.com/bitcoin-core/gui-qml/pull/450 so when reviewing, its best to select just the new commits in the "Files changed" tab.

These changes add properties for the error strings used in the QML to SendRecipient and add validation methods that are checked when the relevant properties are changed by the user.

Screenshot from 2025-06-01 12-42-20


GBKS:
Nice addition. A few notes:

1. Can we restrict the character input altogether, so I can only enter numbers (and periods/commas) in the amount field and only allowed characters in the address field?
2. We should also add a max length. I think 62 is max [for addresses](https://bitcoin.design/guide/glossary/address/), and 17 for bitcoin amounts (21000000.00000000). Not 100% on those numbers. Either way, right now I can type in infinite characters, so some restriction would be good.
3. I was not able to get any address to validate (grabbed a few different address types from [here](https://iancoleman.io/bip39/#english)).
5. I see the error "Amount exceeds maximum limit". Would be more useful if it states the limit, like "The maximum is 21,000,000." or so - sneak in the supply cap.
6. When I enter a huge number and deselect the field, the number changes. Not sure why, and I could not find a pattern how it changes.
7. For an empty wallet, there is the question whether the Send form should be enabled at all.

Always a bit finicky these input fields.

MarnixCroes (commented):
1e81bb677c35aa599547507e44a3a9f7d82eee54

- the UX is better if the verification is done without having to change focus from the field:

[Screencast from 2025-06-02 13-12-13.webm](https://github.com/user-attachments/assets/43543599-af36-4718-b385-490ceefa92af)

- the amount converter doesn't seem to be working as it should:

[Screencast from 2025-06-02 13-14-41.webm](https://github.com/user-attachments/assets/f563d368-d95b-4535-aa7c-5d521f6c2805)

src/qml/models/sendrecipient.cpp
```diff
@@ -77,3 +108,36 @@ void SendRecipient::clear()
...
+void SendRecipient::validateAmount()
+{
+    setAmountError("");
+
+    if (m_amount->isSet()) {
+        if (m_amount->satoshi() <= 0) {
+            setAmountError(tr("Amount must be greater than zero"));
+        } else if (m_amount->satoshi() > MAX_MONEY) {
+            setAmountError(tr("Amount exceeds maximum limit"));
+        } else if (m_amount->satoshi() > m_wallet->balanceSatoshi()) {
```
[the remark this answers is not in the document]
johncarl-te: Yeah, maybe just test this during transaction creation.

johncarl-te:
> [1e81bb6](https://github.com/bitcoin-core/gui-qml/commit/1e81bb677c35aa599547507e44a3a9f7d82eee54)
> 
>     * the UX is better if the verification is done without having to change focus from the field:
Good catch, I will split up the validation and formatting helpers so that validation can happen while editing and it will reformat once editing is finished.
> 
> Screencast.from.2025-06-02.13-12-13.webm
> 
>     * the amount converter doesn't seem to be working as it should:
> 
> 
> Screencast.from.2025-06-02.13-14-41.webm

bummer, must have introduced a bug with conversion. I will double check the sats to btc conversion.

johnny9:
> Nice addition. A few notes:
> 
> 1. Can we restrict the character input altogether, so I can only enter numbers (and periods/commas) in the amount field and only allowed characters in the address field?
Yes, I can add validators to restrict inputs into this form
> 2. We should also add a max length. I think 62 is max [for addresses](https://bitcoin.design/guide/glossary/address/), and 17 for bitcoin amounts (21000000.00000000). Not 100% on those numbers. Either way, right now I can type in infinite 
characters, so some restriction would be good.

I will look at what options there are to restrict the character limit

> 3. I was not able to get any address to validate (grabbed a few different address types from [here](https://iancoleman.io/bip39/#english)).

Thats interesting. I will double check. Could it be regtest/testnet/mainnet differences in format.

> 4. I see the error "Amount exceeds maximum limit". Would be more useful if it states the limit, like "The maximum is 21,000,000." or so - sneak in the supply cap.

I can update the error string to include this

> 5. When I enter a huge number and deselect the field, the number changes. Not sure why, and I could not find a pattern how it changes.

This is a precursor to what will end up being the rich formatting of the amount. If you are inputing BTC amounts it will add the satoshi trailing 0's

> 6. For an empty wallet, there is the question whether the Send form should be enabled at all.
> 
> Always a bit finicky these input fields.

I will consider this.

src/qml/models/sendrecipient.cpp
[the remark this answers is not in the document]
johnny9: I think im going to leave the circular dependency in for now.and take another look when doing validation with multiple recipients

MarnixCroes (commented):
tested LGTM 3630bc28744b3a2b9f4521af6756c2c9254d15a9

Note that when entering full balance, `Review` is highlighted but not clickable.

As the GUI won't allow creating tx's with 0 fees, maybe `Review` should only be highlighted when amount < balance

GBKS:
Looking good.

**Address input**

1. I just fixed some of the input logic in the [prototype](https://lively-kashata-cfde7e.netlify.app/screen/send), so it properly breaks into multiple lines and applies for formatting. Not sure if you're referencing that right now or not. 

**Amount input**

1. Do we need to force the 8 decimals for bitcoin amounts? Seems easier to just have `0.1` rather than `0.10000000`.
2. The unit toggle needs hover states
3. As I mentioned before, when the amount exceeds the balance, we discussed showing `Balance: 0,00348200 ₿. Send max.`, where `Send max` is underlined and a button that inserts the full balance into the input field. I have this partially working in the [prototype](https://lively-kashata-cfde7e.netlify.app/screen/send), but need to fix some logic. Gets a little tricky when we also have the `Include fee in amount` toggle and multiple recipients. Maybe something for a separate PR?

johnny9:
> tested LGTM [3630bc2](https://github.com/bitcoin-core/gui-qml/commit/3630bc28744b3a2b9f4521af6756c2c9254d15a9)
> 
> Note that when entering full balance, `Review` is highlighted but not clickable.
> 
> As the GUI won't allow creating tx's with 0 fees, maybe `Review` should only be highlighted when amount < balance

This is a good one. I will add an issue to include this in my validation when I add proper fee estimation.

johnny9:
Update from [3630bc2](https://github.com/bitcoin-core/gui-qml/commit/3630bc28744b3a2b9f4521af6756c2c9254d15a9) to [29e568c](https://github.com/bitcoin-core/gui-qml/commit/29e568cc1dbf0fada047481b7b650090acc392ba)
- Rebased with main

src/qml/pages/wallet/Send.qml
```diff
@@ -170,78 +169,134 @@ PageStack {
...
+                        id: address
+                        Layout.fillWidth: true
+                        labelText: qsTr("Send to")
+                        placeholderText: qsTr("Enter address...")
+                        text: root.recipient.address
+                        onTextEdited: root.recipient.address = address.text
+                        validator: RegExpValidator {
+                            regExp: /^[1-9A-HJ-NP-Za-km-zac-hj-np-z02-9]+$/
+                        }
+                        maximumLength: 62
```
MarnixCroes, line 185: I understand the intention, but I'm not sure if this hard coded value is a good idea.

src/qml/models/sendrecipient.cpp
```diff
@@ -77,3 +108,42 @@ void SendRecipient::clear()
...
+    }
+
+    Q_EMIT isValidChanged();
+}
+
+void SendRecipient::validateAmount()
+{
+    if (m_amount->isSet()) {
+        if (m_amount->satoshi() <= 0) {
+            setAmountError(tr("Amount must be greater than zero"));
```
MarnixCroes, line 133: I'm able to enter zero without the error.
The error was displayed the first time I entered 0, after that the error wasn't there anymore

image

src/qml/pages/wallet/Send.qml
```diff
@@ -170,78 +169,134 @@ PageStack {
...
+                            anchors.left: amountLabel.right
+                            anchors.verticalCenter: parent.verticalCenter
+                            leftPadding: 0
+                            font.family: "Inter"
+                            font.styleName: "Regular"
+                            font.pixelSize: 18
+                            color: Theme.color.neutral9
+                            placeholderTextColor: enabled ? Theme.color.neutral7 : Theme.color.neutral4
+                            background: Item {}
+                            placeholderText: "0.00000000"
```
MarnixCroes, line 241: the placeholderText should probably also be dependent on the unit.
i.e. makes no sense to display 0.0000000 when using sats.

maybe for follow up.

MarnixCroes (commented):
29e568cc1dbf0fada047481b7b650090acc392ba
LGTM,

a few minor comments:

- it's possible to have the `Amount exceeds balance` while the entry is 0.
This happens when you delete the full entry by selecting and deleting in full. Using backspace to get to zero works as it should.
Screenshot from 2025-07-18 13-35-11

hebasto:
Closing for now.

PR #463, opened 2025-06-01
Fix layouts 1/n

goqusan:
- avoid anchors, use layouts
- override contentItem
- avoid nested stack view

### DIFF
--- a/src/qml/pages/wallet/Activity.qml
+++ b/src/qml/pages/wallet/Activity.qml
@@ -20,146 +20,136 @@ PageStack {
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/qml/pages/wallet/Activity.qml
```diff
@@ -20,146 +20,136 @@ PageStack {
         }
     }
 
-    initialItem: Page {
-        id: root
-        background: null
+    initialItem: RowLayout {
```
johnny9, line 23: should Page be the initial item?

PR #465, opened 2025-06-05
Show "Select Wallet" or "Add Wallet" in the Wallet selector if a wallet isn't loaded

johnny9:
In addition to changing the component in the top left corner, the navigation tabs will be hidden if no wallet is loaded until a wallet is created/selected.

Screenshot from 2025-06-05 00-05-47
Screenshot from 2025-06-05 00-08-48
Screenshot from 2025-06-05 00-06-08


GBKS:
ACK 03b568d

Thanks for adding that. I'd like to tweak some minor visual details on these, but I think this is good to merge and we can open a separate PR.

Was just looking through the code a bit and could not find the spot where it defines the transition direction for pages to appear from and disappear to. Based on context, we'll also need these to differ. Can you maybe point me to where that is defined?

PR #475, opened 2025-07-11
Add cmake, qt6, and bitcoin core submodule

pinheadmz:
This PR rebases the QML GUI code on upstream Bitcoin Core master, updates dependencies and build system, and defines a new organizational model for the project which may eventually lead to a separation of the GUI completely from bitcoin/bitcoin.

# Code changes in `qml/`

Each commit has a link to Qt docs or to a Bitcoin Core PR to explain it. QML code had to be updated to respect API changes and other conflicts:
- Upstream master bitcoin/bitcoin
- Qt6

# Git submodule

Bitcoin core is now included as a git submodule. This is a cleaner model than forking the bitcoin core repo:
- Prevent GUI contributors from creating new conflicts with upstream by changing bitcoin core in the forked repo
- Garuntee that bitcoin core code is not modified, meaning we don't have to run upstream tests
- Simplifies the build instructions and the repo itself to make it easier for front end contributors

# Cmake

Adds a cmake build system on top of the new submodule+`qml/` pair. We currently still need to compile upstream `qt/` and there is no flag to separate its library from its binary, so the result is that both GUIs will be built, but no command-line daemon, fuzz, tests, or utility binaries:
```
build/bin
├── bitcoin-core-app
└── bitcoin-qt
```

**Note:** I'm going with the title "Bitcoin Core App" from the design community and to make even more clear the separation from `bitcoin-qt`

# CI

I added a very minimal CI to build binaries for macos and ubuntu, and upload the [artifacts](https://github.com/bitcoin-core/gui-qml/actions/runs/16242179284). This is mainly a "does it build?" test but I have downloaded and run both artifacts locally and they do work, given that the right linked libraries are available. Future work can make those artifacts more easily usable by testers.

# Future work

Obviously there is still a lot to do, but I consider these to be the next major steps:
- GUI tests
- Tests builds for all supported platforms
- Extract `qt/` from upstream and integrate in this repo
    -  Much of that code can probably be abandoned now
    - The build system can be cleaned up and the legacy binary removed from the output
- Extract Qt depends build from upstream and integrate in this repo
    - After this, the upstream repo will be set up to clear out all GUI code
- Figure out guix builds, guix attestations, project release process, and overall degree of separation from Bitcoin Core


johnny9 (approved):
tACK

This is a great start. The application appears mostly working. I tested against ubuntu provided Qt 6.4.2 as well as the latest Qt 6.9.1

There are some QML things to fix but I think its best to work on those as follow up PRs

hebasto:
On Ubuntu 24.04, the additional `qml6-module-qt-labs-settings` package was required for runtime.

PR #478, opened 2025-07-27
Fix too many arguments warning in BlockClock

johnny9:
Fixes the following warning

```
expression for onCompleted@qrc:/qml/components/BlockClock.qml:120
Too many arguments, ignoring 1
```


hebasto (approved):
ACK 0392f2c78c6eb98fbdcbbee8e5bc80e668c645f0, tested on Ubuntu 24.04.

PR #479, opened 2025-07-27
Add `.python-version` file

hebasto:
Required when using [`pyenv`](https://github.com/pyenv/pyenv).


hebasto:
cc @johnny9 @pinheadmz

johnny9 (approved):
ACK

PR #480, opened 2025-07-27
cmake: Add missed `Boost::headers`

hebasto:
This PR fixes building on macOS with Boost installed via Homebrew.


johnny9 (approved):
tACK

tested on MacOs 15.5

PR #487, opened 2025-07-31
Use handlers with explicit parameters in main.qml and DesktopWallets.qml

### LINKED ISSUE #481: qt6: Parameter "event" is not declared
A few moments after starting:
```
GUI: qrc:/qml/pages/main.qml:50:9 Parameter "event" is not declared. Injection of parameters into signal handlers is deprecated. Use JavaScript functions with formal parameters instead.
```

johnny9:
Signals with parameters should have handlers that explicitly define the arguments. Without this, Qt will throw a warning in the log stating that injection or parameters is deprecated.

fixes #481

### DIFF
--- a/qml/pages/main.qml
+++ b/qml/pages/main.qml
@@ -47,7 +47,7 @@ ApplicationWindow {
         }
         anchors.fill: parent
         focus: true
-        Keys.onReleased: {
+        Keys.onReleased: (event) => {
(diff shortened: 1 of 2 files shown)


hebasto (approved):
ACK 99d5da865466924257daf4d8b59db468992c69e6, tested on Ubuntu 24.04.

PR #497, opened 2025-08-28
Add first unittests

johnny9:
This adds an initial unittest against the bitcoinamount module. This is the simplest c++ module to test so its an easy first target. I think the Qt test framework is the best fit for the project and eventually I imagine GMock will be needed to mock out portions of bitcoin interfaces as well as more complex modules get tested. These changes update the CMake definitions and add a BUILD_APP_TESTS option to toggle building of the unittests. 

A new ci workflow is added as well to run the unittests (and eventually other testing that we have).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,3 +127,10 @@ target_link_libraries(bitcoin-core-app
     bitcoinqml
     bitcoinqt
 )
+
+option(BUILD_APP_TESTS "Build unit tests for the app" ON)
+if(BUILD_APP_TESTS)
+  include(CTest)
+  enable_testing()
+  add_subdirectory(test)
+endif()
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -38,7 +38,7 @@ jobs:
             build-essential ccache cmake pkgconf \
             libevent-dev libboost-dev libsqlite3-dev libgl-dev libqrencode-dev \
             qt6-base-dev qt6-tools-dev qt6-l10n-tools qt6-tools-dev-tools \
-            qt6-declarative-dev qml6-module-qtquick qml6-module-qtqml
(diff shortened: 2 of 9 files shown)


jarolrod:
ACK 16e86dddbdc069d126a5f2131f181cee05c80171

@hebasto can you give your opinion on the CMAKE toggle here

johnny9:
Squashed the ci fixes

johnny9:
I introduced a small qml test as well. We will use Qtest to test the c++ modules as well as the qml

D33r-Gee:
tACK [236141e](https://github.com/bitcoin-core/gui-qml/commit/236141ea2db490114cd25976cc442be733467f06) on WSL Unbuntu 24.04.3

The binaries build without issues and the test returned "Passed"

Also the additions of the qml modules dependencies is helpful

```
ctest --test-dir build --output-on-failure

 Start 1: bitcoinqml_unit_tests
1/2 Test #1: bitcoinqml_unit_tests ............   Passed    0.04 sec
    Start 2: bitcoinqml_qmltests
2/2 Test #2: bitcoinqml_qmltests ..............   Passed    0.04 sec
```

CMakeLists.txt
```diff
@@ -129,4 +126,11 @@ target_link_libraries(bitcoin-core-app
...
     bitcoinqml
     bitcoinqt
-)
\ No newline at end of file
+)
+
+option(BUILD_APP_TESTS "Build unit tests for the app" ON)
+if(BUILD_APP_TESTS)
+  include(CTest)
+  enable_testing()
```
hebasto, line 134: From `enable_testing` [docs](https://cmake.org/cmake/help/latest/command/enable_testing.html):
> This command is also automatically invoked when the [CTest](https://cmake.org/cmake/help/latest/module/CTest.html#module:CTest) module is included, except if the [BUILD_TESTING](https://cmake.org/cmake/help/latest/variable/BUILD_TESTING.html#variable:BUILD_TESTING) option is turned off.

```suggestion
  include(CTest)
```

PR #501, opened 2026-02-17
Wire RequestPayment page to PaymentRequest model

johnny9:
- add qml/models/paymentrequest.{h,cpp}
- expose WalletQmlModel::currentPaymentRequest
- add WalletQmlModel::commitPaymentRequest()
- define recent request entry struct for serialization
- persist receive requests via setAddressReceiveRequest()
- register PaymentRequest in qml/bitcoin.cpp
- update RequestPayment.qml to bind amount/label/message/address to model
- keep create/copy/clear actions model-driven

Easiest way to currently test is to create the payment request and then load up the wallet in the Qt gui

### DIFF
--- a/qml/bitcoin.cpp
+++ b/qml/bitcoin.cpp
@@ -38,6 +38,7 @@
 #include <qml/models/networktraffictower.h>
 #include <qml/models/nodemodel.h>
 #include <qml/models/options_model.h>
+#include <qml/models/paymentrequest.h>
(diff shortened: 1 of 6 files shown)


johnny9:
https://github.com/user-attachments/assets/57063499-a371-453f-b1d3-2434c085d1d8

hebasto (approved):
ACK 68acd2f3830dbd31292696139a7e0ed94eae648b, I've skimmed the code, it looks OK.

Issue #505, opened 2026-02-28
Feature Parity 01: Remove `bitcoin/src/qt` GUI Dependencies from gui-qml


johnny9:
## Summary
Create a clean boundary where gui-qml does not depend on Qt Widgets-era classes under `bitcoin/src/qt`.

## Complexity
- `XL`

## Dependencies
- `None`

## Global Constraints
- All implementation lives under `gui-qml/`.
- No GUI runtime dependency on `bitcoin/src/qt` components. Reused logic must be moved/ported into `gui-qml/`.
- Do not change `interfaces::Wallet` or `interfaces::Node` in this version.
- Issue completion must satisfy `DefinitionOfDone.md`.

## Scope
- Audit and remove includes in `qml/*` that point to `qt/*` headers.
- Port or re-implement required helpers/models in `gui-qml/qml` C++ modules (peer table/sort, formatting, URI helpers, network style wrappers as needed).
- Adjust CMake targets/includes so gui-qml compiles without qt GUI-model reuse.
- Keep behavior parity for already-working features while moving code.

## Acceptance Criteria (DoD-Aligned)
1. `rg -n "#include <qt/|#include <bitcoin/src/qt/" qml` returns no results for GUI runtime code paths.
2. App builds and boots in node-only and wallet-enabled modes.
3. Existing Peers/Activity/Send/Receive surfaces continue to function after migration.
4. No new main-thread blocking behavior introduced during migration.

## Required Tests
- C++ unit tests: Unit tests for each newly ported C++ helper/model in `qml/`.
- QML QTest: QTests for any QML surface touched by migrated models.

## Implementation Notes
- Start here; this is a hard dependency for the full parity program.
- Preserve API shapes where possible to minimize QML churn.

## Tasks
- [ ] Port `qt/bitcoinunits.h` usage to a local `qml/` formatter and update `CoinsListModel` + `WalletQmlModel` formatting call sites.
- [ ] Add unit coverage for the local amount formatter and wire it into `test/CMakeLists.txt`.
- [ ] Replace `qt/networkstyle.h` usage in `qml/bitcoin.cpp` + `qml/imageprovider.cpp` with local equivalents.
- [ ] Replace `qt/guiutil.h` usage in `qml/bitcoin.cpp` (font loading) with a local helper or Qt-native equivalent.
- [ ] Remove `qt/initexecutor.h` dependency (port/replace InitExecutor as needed).
[...]

PR #527, opened 2026-03-01
Decouple Peers Models from bitcoin/src/qt

johnny9:
Refactors peer table from src/qt into a local model for gui-qml.

GMock framework is added to the CMakeLists.txt along with a defined MockNode in the test folder

Unittests are added for the peerlistmodel and peerstatsutil. Peerlistmodel unittests use the MockNode.

Related to #505


qml/models/peerlistmodel.h
```diff
@@ -0,0 +1,73 @@
...
+        Age,
+        Address,
+        Direction,
+        ConnectionType,
+        Network,
+        Ping,
+        Sent,
+        Received,
+        Subversion,
+        StatsRole = Qt::UserRole,
```
epicleafies: StatsRole should appear first in the enum list so the ordering is clear

test/test_peerlistmodel.cpp
```diff
@@ -0,0 +1,182 @@
...
+    PeerListModel model{node, nullptr};
+    const int calls_after_ctor = get_nodes_stats_calls;
+    QCOMPARE(calls_after_ctor, 1);
+
+    model.startAutoRefresh();
+    QTRY_VERIFY_WITH_TIMEOUT(get_nodes_stats_calls > calls_after_ctor, 1200);
+
+    model.stopAutoRefresh();
+    const int calls_after_stop = get_nodes_stats_calls;
+    QTest::qWait(450);
```
epicleafies: Should the 450 be a named constant?

qml/models/peerlistsortproxy.cpp
```diff
@@ -60,9 +52,41 @@ QString PeerListSortProxy::sortBy() const
...
+    if (m_sort_by == roleName) return;
+
+    m_sort_by = roleName;
+    m_sort_role = RoleNameToRole(roleName);
+    setSortRole(m_sort_role);
+    sort(0);
+    Q_EMIT sortByChanged(roleName);
+}
+
+bool PeerListSortProxy::lessThan(const QModelIndex& left_index, const QModelIndex& right_index) const
```
epicleafies, line 64: No testing for any of the cases.

qml/models/peerlistmodel.cpp
```diff
@@ -0,0 +1,139 @@
...
+Qt::ItemFlags PeerListModel::flags(const QModelIndex& index) const
+{
+    if (!index.isValid()) return Qt::NoItemFlags;
+    return Qt::ItemIsSelectable | Qt::ItemIsEnabled;
+}
+
+void PeerListModel::refresh()
+{
+    interfaces::Node::NodesStats nodes_stats;
+    m_node.getNodesStats(nodes_stats);
```
epicleafies: Return value (bool) is ignored.

qml/models/peerlistmodel.cpp
```diff
@@ -0,0 +1,139 @@
...
+    decltype(m_peers_data) new_peers_data;
+    new_peers_data.reserve(nodes_stats.size());
+    for (const auto& node_stats : nodes_stats) {
+        const CNodeCombinedStats stats{std::get<0>(node_stats), std::get<2>(node_stats), std::get<1>(node_stats)};
+        new_peers_data.append(stats);
+    }
+
+    for (int i = 0; i < m_peers_data.size();) {
+        if (i < new_peers_data.size() && m_peers_data.at(i).nodeStats.nodeid == new_peers_data.at(i).nodeStats.nodeid) {
```
epicleafies: This comparison is fragile, it assumes the peers will be returned in the same order across calls. If not, valid peers are discarded.

qml/models/peerlistmodel.cpp
```diff
@@ -0,0 +1,139 @@
...
+            ++i;
+            continue;
+        }
+        beginRemoveRows(QModelIndex(), i, i);
+        m_peers_data.erase(m_peers_data.begin() + i);
+        endRemoveRows();
+    }
+
+    if (m_peers_data.size() < new_peers_data.size()) {
+        beginInsertRows(QModelIndex(), m_peers_data.size(), new_peers_data.size() - 1);
```
epicleafies: beginInsertRows only inserts at tail. If a new new peer appears earlier in the list it removes then reinserts all peers at the end, this causes the peers list to temporary show a different order.

qml/models/peerlistmodel.cpp
```diff
@@ -0,0 +1,139 @@
...
+        m_peers_data.swap(new_peers_data);
+        endInsertRows();
+    } else {
+        m_peers_data.swap(new_peers_data);
+    }
+
+    if (rowCount() > 0) {
+        const auto top_left = index(0, 0);
+        const auto bottom_right = index(rowCount() - 1, 0);
+        Q_EMIT dataChanged(top_left, bottom_right);
```
epicleafies, line 140: Double fetches the newly added rows, should only get the already existing rows.

qml/models/peerlistmodel.cpp
```diff
@@ -0,0 +1,139 @@
...
+            continue;
+        }
+        beginRemoveRows(QModelIndex(), i, i);
+        m_peers_data.erase(m_peers_data.begin() + i);
+        endRemoveRows();
+    }
+
+    if (m_peers_data.size() < new_peers_data.size()) {
+        beginInsertRows(QModelIndex(), m_peers_data.size(), new_peers_data.size() - 1);
+        m_peers_data.swap(new_peers_data);
```
epicleafies: Using std::move would be clearer and more efficient.

qml/models/peerlistmodel.cpp
```diff
@@ -0,0 +1,139 @@
...
+int PeerListModel::rowCount(const QModelIndex& parent) const
+{
+    if (parent.isValid()) return 0;
+    return m_peers_data.size();
+}
+
+QVariant PeerListModel::data(const QModelIndex& index, int role) const
+{
+    if (!index.isValid() || index.row() < 0 || index.row() >= m_peers_data.size()) return {};
+    CNodeCombinedStats* rec = const_cast<CNodeCombinedStats*>(&m_peers_data.at(index.row()));
```
epicleafies: data() is const but strips it to pass StatsRole as amutable pointer. It should either take a const CNodeCOmbinedStats* or it should expose a seperate method for raw stats access.

test/test_peerlistmodel.cpp
```diff
@@ -0,0 +1,182 @@
...
+            out_stats = stats;
+            return true;
+        }));
+
+    PeerListModel model{node, nullptr};
+    const int calls_after_ctor = get_nodes_stats_calls;
+    QCOMPARE(calls_after_ctor, 1);
+
+    model.startAutoRefresh();
+    QTRY_VERIFY_WITH_TIMEOUT(get_nodes_stats_calls > calls_after_ctor, 1200);
```
epicleafies: Should the 1200 be a named constant? Less than 5x slower than timer period may need larger safety margin.

epicleafies (commented):
I was able to build it successfully and all tests passed. Most issues around the refresh function in peerlistmodel.cpp, there are also a few other small things. Over all looks good.

qml/models/peerlistmodel.h
johnny9: Applied in a4103133a: moved StatsRole to the top of the enum so role ordering is explicit.

test/test_peerlistmodel.cpp
johnny9: Addressed in 8e71060f2: extracted the stop wait into a named constant.

qml/models/peerlistsortproxy.cpp
johnny9: Addressed in 8e71060f2: added sort proxy coverage for each role comparator in test_peerlistmodel.cpp.

qml/models/peerlistmodel.cpp
johnny9: Fixed in a4103133a: refresh() now checks getNodesStats() and returns early on failure.

qml/models/peerlistmodel.cpp
johnny9: Fixed in a4103133a: refresh() now detects structural/id-order changes explicitly and handles them without relying on positional equality assumptions.

qml/models/peerlistmodel.cpp
johnny9: Addressed in a4103133a: structural changes now use model reset rather than tail-only insertion, so peers can appear at the correct position immediately.

qml/models/peerlistmodel.cpp
johnny9: Fixed in a4103133a: dataChanged() is now only emitted for stable row sets; structural changes take the reset path instead of touching newly inserted rows.

qml/models/peerlistmodel.cpp
johnny9: Updated in a4103133a: switched from swap() to std::move assignment for clearer ownership transfer.

qml/models/peerlistmodel.cpp
johnny9: Fixed in a4103133a: StatsRole now returns a const CNodeCombinedStats* and constness is preserved through model/sort/details usage.

test/test_peerlistmodel.cpp
johnny9: Addressed in 8e71060f2: extracted the timeout to a named constant and increased it to 2000ms for additional safety margin.

epicleafies:
tACK
610e65db6

epicleafies:
tACK f2e675a7f

One thing not directly code related, since BUILD_APP_TESTS now defaults to on may want to add libgtest-dev and libgmock-dev as dependencies in the README.

PR #528, opened 2026-03-02
Move amount formatting off qt/bitcoinunits

johnny9:
Add a local QmlBitcoinUnits formatter in gui-qml and switch wallet/coins QML  models to use it instead of qt/bitcoinunits.h.

Related to #505


hebasto (approved):
ACK 579024b8885686cb0e9f4cc5bda3f6212341330e, I have skimmed the changes and they look reasonable.

Issue #507, opened 2026-02-28
Feature Parity 03: Peer Management: Disconnect, Ban/Unban, and Bans Diagnostics


johnny9:
## Summary
Deliver full peer control parity for disconnect + ban workflows and banned-peers diagnostics surface.

## Complexity
- `L`

## Dependencies
-  #505 
-  #506 

## Global Constraints
- All implementation lives under `gui-qml/`.
- No GUI runtime dependency on `bitcoin/src/qt` components. Reused logic must be moved/ported into `gui-qml/`.
- Do not change `interfaces::Wallet` or `interfaces::Node` in this version.
- Issue completion must satisfy `DefinitionOfDone.md`.

## Scope
- Add QML actions for disconnect by peer id/address from peer details and peer list contexts.
- Add ban/unban actions and duration handling.
- Implement bans list model/view with refresh and unban controls.
- Wire to `interfaces::Node` methods only (`disconnectById`, `disconnectByAddress`, `ban`, `unban`, `getBanned`).

## Acceptance Criteria (DoD-Aligned)
1. User can disconnect a connected peer from UI and sees immediate state update.
2. User can ban and unban peers; bans list reflects changes without app restart.
3. Error states (invalid address, RPC failure, race where peer disconnects first) are surfaced clearly.
4. No polling loops or blocking calls on UI thread.

## Required Tests
- C++ unit tests: Unit tests for bans/peer action model wrappers around node interface.
- QML QTest: QTests for peer action button enable/disable logic and status rendering.
- Python E2E (`qml/test/testbridge.h`): E2E flow: open peers -> ban peer -> verify bans list -> unban -> verify restored behavior.

## Implementation Notes
- Likely files: `qml/models/nodemodel.*`, new `qml/models/banlistmodel.*`, `qml/pages/node/Peers.qml`, `qml/pages/node/PeerDetails.qml`.

johnny9:
https://github.com/bitcoin-core/gui-qml/pull/529

PR #529, opened 2026-03-02
Functional tests for peer disconnect, ban and unban

epicleafies:
Tests disconnecting peers (with one and more than one peers), banning peers (all durations, persistence between sessions and with more than one peer) and unbanning peers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,7 @@ option(ENABLE_TEST_AUTOMATION "Enable test automation bridge for QML UI testing"
 set(BUILD_BENCH OFF)
 set(BUILD_BITCOIN_BIN OFF)
 set(BUILD_CLI OFF)
-set(BUILD_DAEMON OFF)
+option(BUILD_DAEMON "Build bitcoind daemon alongside the app (required for functional peer tests)" OFF)
 set(BUILD_FOR_FUZZING OFF)
 set(BUILD_FUZZ_BINARY OFF)
 set(BUILD_GUI_TESTS OFF)
--- a/test/functional/qml_test_peers.py
+++ b/test/functional/qml_test_peers.py
@@ -60,49 +60,18 @@ BAN_DURATIONS = [
 def find_bitcoind():
     """Locate the bitcoind binary.
 
-    Search order:
-      1. BITCOIND environment variable
-      2. Same directory as bitcoin-core-app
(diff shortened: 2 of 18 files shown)


test/functional/qml_test_peers.py
```diff
@@ -0,0 +1,675 @@
...
+
+def find_bitcoind():
+    """Locate the bitcoind binary.
+
+    Search order:
+      1. BITCOIND environment variable
+      2. Same directory as bitcoin-core-app
+      3. build/bin/bitcoind relative to repo root
+      4. Sibling bitcoin repository: <repo_root>/../bitcoin/build/bin/bitcoind
+      5. Fall back to bitcoin-core-app itself (it is a full node)
```
johnny9: Lets not bother with this complex failover. We should trust that our environment is stable. Pick bitcoind in build/bin/bitcoind to test against.

test/functional/qml_test_peers.py
```diff
@@ -0,0 +1,675 @@
...
+
+    _wait_for_peer_gone(harness, target_id)
+
+    peers = harness.rpc_call("getpeerinfo")
+    remaining_ids = [p["id"] for p in peers]
+    assert target_id not in remaining_ids, \
+        f"Peer {target_id} should be disconnected"
+    assert other_id in remaining_ids, \
+        f"Peer {other_id} should still be connected"
+    print(f"  PASSED: peer {target_id} gone; peer {other_id} still present")
```
johnny9, line 570: We should also assert the number of peers is correct in case the original peer reconnected with a new id

johnny9:
Related to #507

CMakeLists.txt
johnny9: Lets patch this so it is configurable so that we can have bitcoind built for this test to use. You'll need to update the github workflow files to add this env variable.

johnny9 (changes requested):

johnny9 (commented):
Everything looks good to me. Just need to add the new test to the github workflow file now.

johnny9 (approved):
tACK 684c7f203dd9b71690f3b149cd9dafc0254d5c83

hebasto:
Please rebase.

johnny9 (approved):
tACK 3fd99d00058ecc21b9ad46f58cb117de36dbea2a

ran test on macos

Issue #510, opened 2026-02-28
Open Debug Log Action


johnny9:
## Summary
Add user action to open `debug.log` from current datadir in desktop environments.

## Complexity
- `S`

## Dependencies
- `01`
- `02`

## Global Constraints
- All implementation lives under `gui-qml/`.
- No GUI runtime dependency on `bitcoin/src/qt` components. Reused logic must be moved/ported into `gui-qml/`.
- Do not change `interfaces::Wallet` or `interfaces::Node` in this version.
- Issue completion must satisfy `DefinitionOfDone.md`.

## Scope
- Expose resolved log path from gui-qml backend.
- Add UI entry in settings/about/developer area.
- Handle missing file and platform open failures with explicit error messages.

## Acceptance Criteria (DoD-Aligned)
1. Action opens the correct debug log file for active chain/datadir.
2. If file does not exist or cannot be opened, user gets actionable error text.

## Required Tests
- C++ unit tests: Unit tests for path resolution and error mapping.
- QML QTest: QTest for action visibility and error banner behavior.
- Python E2E (`qml/test/testbridge.h`): E2E step asserting action triggers expected backend call and success/error UI state.

## Implementation Notes
- Use Qt platform APIs directly from gui-qml code; do not call Qt Widgets GUI helpers.

johnny9:
https://github.com/bitcoin-core/gui-qml/pull/532

PR #532, opened 2026-03-10
Debug log page and tests

epicleafies:
Created the debug log page #510. 
Created functional and C++ unit tests for the debug log page. 
The debug log is viewable within the page and can be viewed in an external text editor. 
Debug log can be updated by the user and also auto updates.
Debug log automatically loads some of the logs and user can load more.
Screenshot from 2026-03-11 21-35-00
Screenshot from 2026-03-11 21-35-22


johnny9 (commented):
I think 26e8ce708eaf31b4b90f7296194c3e49067f3286 should be pulled out into a different PR as it didn't fully address the gui test flakiness. Take a look at #533 I think it resolved the flakiness. At least i am able to run 50+ times in a row with those changes. I suspect waiting on PageStack transitions had the biggest impact.

hebasto:
Please rebase.

qml/models/peerdetailsmodel.cpp
```diff
@@ -23,37 +23,50 @@ PeerDetailsModel::PeerDetailsModel(const CNodeCombinedStats* nodeStats, PeerList
     connect(parent, &PeerListModel::modelReset, this, &PeerDetailsModel::onModelReset);
 }
 
-void PeerDetailsModel::onModelRowsRemoved(const QModelIndex&  parent, int first, int last)
+void PeerDetailsModel::onModelRowsRemoved(const QModelIndex& parent, int first, int last)
```
johnny9: Drop the commit 35b9c49f9d6fe939b2ed8e46754b929ecca9ba37 from this PR and open a new PR if this is still needed.

test/functional/qml_test_peers.py
```diff
@@ -31,7 +31,7 @@
     dump_qml_tree,
     find_gui_binary,
 )
-from qml_driver import QmlDriver
+from qml_driver import QmlDriver, QmlDriverError
```
johnny9: Drop commit 7c67c2a5e125ba81e1bb17da623c6f84706c42fb as i don't think it is necessary anymore

johnny9 (commented):
Drop some of the unneeded commits that were created while we trying to resolve issues with the peer end to end test

epicleafies:
All peer test fix attempts should be removed.

qml/models/nodemodel.h
```diff
@@ -65,6 +80,24 @@ class NodeModel : public QObject
     void startShutdownPolling();
     void stopShutdownPolling();
 
+    QString debugLogPath() const;
+    Q_INVOKABLE bool openDebugLogFile();
+    QString debugLogOpenError() const { return m_debug_log_open_error; }
+    Q_INVOKABLE QVariantList debugLogLines(int max_lines = 10000);
+
+    QString formattedDebugLog() const { return m_formatted_debug_log; }
```
johnny9: All of this is quite complicated. Enough so that I think it would be better to create a model specifically for parsing the debug log. DebugLogModel

johnny9 (commented):
https://github.com/user-attachments/assets/88900b7b-bebe-423d-a529-2315c00d1ec0

Really cool set of features here. Unfortunately, it seems the html rich text rendering isnt working well on Ubuntu or Fedora. You can see the text disappearing after scrolling. It might be the case that this needs to be converted to a ListView of line items instead of an html table.

Also the spinner needs adjustment

johnny9:
> Really cool set of features here. Unfortunately, it seems the html rich text rendering isnt working well on Ubuntu or Fedora. You can see the text disappearing after scrolling. It might be the case that this needs to be converted to a ListView of line items instead of an html table.

After testing I found that it was related to trying to render the html table layout

epicleafies:
The spinner works for me locally, I'll look into it. Does the rpc console display work for you? If so I'll probably update the debug log page so that the display is consistent.

johnny9:
> The spinner works for me locally, I'll look into it. Does the rpc console display work for you? If so I'll probably update the debug log page so that the display is consistent.

the RPC console listview renders correctly but the scrolling doesnt work quite right. I left a comment on the PR on how to solve it. Switching to a listview to column layout for the debug log unfortunately means that the standard mouse selection highlight wont work. It would render exactly like we want but we'd have to think about the select+copy ux a bit. Anyway, it doesnt need to be perfect but lets at least solve this rendering problem

qml/components/MonospaceOutputView.qml
```diff
@@ -0,0 +1,223 @@
+// Copyright (c) 2026 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import org.bitcoincore.qt 1.0
```
johnny9, line 8:
```suggestion
import org.bitcoincore.qt 1.0
import "../controls"
```

johnny9 (changes requested):
Missing an import in MonospaceOutputView

Everything else looks good enough for an ACK

qml/components/MonospaceOutputView.qml
johnny9: this is needed to resolve "GUI: qrc:/qml/components/MonospaceOutputView.qml:45: ReferenceError: Theme is not defined"

qml/pages/settings/SettingsDebugLog.qml
```diff
@@ -0,0 +1,317 @@
...
+
+                Rectangle {
+                    anchors.fill: parent
+                    radius: 5
+                    color: refreshArea.containsMouse ? Theme.color.neutral2
+                                                     : Theme.color.background
+                    Behavior on color { ColorAnimation { duration: 150 } }
+                }
+
+                Icon {
```
MarnixCroes, line 83: The icon should have tooltip on hover, what it does

qml/pages/settings/SettingsDebugLog.qml
```diff
@@ -0,0 +1,317 @@
...
+
+                Rectangle {
+                    anchors.fill: parent
+                    radius: 5
+                    color: exportArea.containsMouse ? Theme.color.neutral2
+                                                    : Theme.color.background
+                    Behavior on color { ColorAnimation { duration: 150 } }
+                }
+
+                Icon {
```
MarnixCroes, line 126: same here, should have a tooltip

MarnixCroes (commented):
8ecc7e5efb7b6ddadf7e6936640261df6c1c5804
 
- The latest entry should be at the bottom, not the top. Same as it is in the actual `debug.log` file
- The time should be the actual time stamps instead of `x minutes ago`, like in the debug log. As it is dev debug log for developers
- I've had a few times that when searching the app becomes really laggy, almost freezes.

MarnixCroes:
> Really cool set of features here. Unfortunately, it seems the html rich text rendering isnt working well on Ubuntu or Fedora. You can see the text disappearing after scrolling. It might be the case that this needs to be converted to a ListView of line items instead of an html table.

> Also the spinner needs adjustment

Can repro, needs to be fixed before merge imo

johnny9:
ACK [bfc2389](https://github.com/bitcoin-core/gui-qml/pull/532/commits/bfc2389a4a26d5138c7464f80744c0c05b17899a)

missing Theme import is fixed.

For follow up: move loading and updates to worker thread,test using ListView for optimization.

johnny9 (approved):

PR #533, opened 2026-03-11
Address functional/qml_test_peers flakiness

johnny9:
Improve `qml_test_peers.py` reliability in the multi-peer flows and around nested page stack transitions.

The peer test had two sources of flakiness:
- the multi-peer disconnect path could act on the wrong peer process or leave behind an extra peer
- some page transitions could be triggered while the nested `StackView` was still animating, causing pushes to be ignored

- map GUI peers back to local peer processes using `session_id` instead of relying on peer ordering
- stop and clean up the second peer process consistently so later test phases start from a clean state
- wait for the exact expected remaining peer after disconnecting one of two peers
- expose the nested `NodeSettings` `PageStack` to test automation
- wait for the nested stack to become idle before and after peer-management page transitions:

## Testing

- `python3 test/functional/qml_test_peers.py`
- repeated locally 30 times with no failures after the transition-wait changes

### DIFF
--- a/qml/pages/node/NodeSettings.qml
+++ b/qml/pages/node/NodeSettings.qml
@@ -15,6 +15,7 @@ PageStack {
     property alias showDoneButton: doneButton.visible
 
     id: root
+    objectName: "nodeSettingsStack"
 
     initialItem: Page {
         background: null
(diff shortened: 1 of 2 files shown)


johnny9:
A follow up on this could be to look at the management of multiple nodes and see if that can be turned into a common utility library. This can wait until we have other multi-node tests.

Another follow up would be to look into disabling animations for test automation. This will speed up the tests and increase reliability. I've done this in another qml project so will look into the scheme that was used for that.

epicleafies:
tACK 33daa4a

Issue #511, opened 2026-02-28
Feature Parity 07: Complete Default/Tor Proxy Settings


johnny9:
## Summary
Finish proxy settings so default and Tor proxy values persist and apply (currently partial).

## Complexity
- `M`

## Dependencies
- `01`
- `02`

## Global Constraints
- All implementation lives under `gui-qml/`.
- No GUI runtime dependency on `bitcoin/src/qt` components. Reused logic must be moved/ported into `gui-qml/`.
- Do not change `interfaces::Wallet` or `interfaces::Node` in this version.
- Issue completion must satisfy `DefinitionOfDone.md`.

## Scope
- Extend options model with proxy enabled/value properties for default and onion.
- Persist via `Node::updateRwSetting` with proper key mapping and validation.
- Update `ProxySettings.qml` to load/edit/save real values.
- Surface restart-required warnings where applicable.

## Acceptance Criteria (DoD-Aligned)
1. Default and Tor proxy settings survive restart and reflect persisted values.
2. Invalid input is blocked and shows inline error before save.
3. Disabling a proxy path correctly clears or deactivates its setting.

## Required Tests
- C++ unit tests: Unit tests for proxy parse/validation and settings serialization.
- QML QTest: QTests for enable toggle, validation text, and load/save behavior.
- Python E2E (`qml/test/testbridge.h`): E2E flow: set proxy values -> restart app -> verify values retained and behavior updated.

## Implementation Notes
- Do not rely on `qt/optionsmodel.*`; keep logic in `qml/models/options_model.*`.

jarolrod:
I’m working on changes that address this issue in a settings/onboarding QML branch. I plan to submit the PR soon and will link it here once it’s open.

PR #534, opened 2026-03-12
Default/Tor proxy settings persistence and tests

epicleafies:
Wired the existing proxy settings UI to the options model so that proxy and Tor proxy configuration is saved to disk and survives restart (#511). 
Added functional and qml tests for proxy configuration. 
Added restart needed alert to the proxy page.

Screenshot from 2026-03-26 17-29-17


qml/models/options_model.cpp
```diff
@@ -45,6 +45,14 @@ OptionsQmlModel::OptionsQmlModel(interfaces::Node& node, bool is_onboarded)
     m_server = SettingToBool(m_node.getPersistentSetting("server"), false);
 
     m_dataDir = getDefaultDataDirString();
+
+    QString proxy_setting = QString::fromStdString(SettingToString(m_node.getPersistentSetting("proxy"), ""));
```
johnny9, line 59: QString proxy_setting = QString::fromStdString(
SettingToString(m_node.getPersistentSetting("proxy"), "")
);
if (proxy_setting == "0") proxy_setting.clear();

m_proxy_enabled = !proxy_setting.isEmpty();
m_proxy_address = proxy_setting;

double check but I think core allows 0 which should be treated the same as ""

test/CMakeLists.txt
```diff
@@ -50,3 +51,25 @@ target_link_libraries(bitcoinqml_qmltests
 )
 
 add_test(NAME bitcoinqml_qmltests COMMAND bitcoinqml_qmltests -platform offscreen)
+
+# C++ model tests that require GoogleMock.
```
johnny9: There should not be a new executable target for the model tests. It should be included in the unit_tests target.

johnny9 (changes requested):
Don't add a new test target. Keep the unit tests bundled in the unit_test executable and the qml tests in the qmltests executable.

test/CMakeLists.txt
```diff
@@ -4,13 +4,27 @@ cmake_minimum_required(VERSION 3.22)
...
+find_package(GTest QUIET)
+
+set(UNIT_TEST_SOURCES
   test_unit_tests_main.cpp
   test_bitcoinamount.cpp
   test_qmlbitcoinunits.cpp
-  ${CMAKE_CURRENT_SOURCE_DIR}/../qml/bitcoinamount.cpp
-  ${CMAKE_CURRENT_SOURCE_DIR}/../qml/bitcoinunits.cpp
 )
+if(TARGET GTest::gmock)
```
johnny9: I think we should just fail if gmock isnt present

johnny9 (commented):
This looks good. We'll need to get the earlier PR in to get gmock added to the project then this PR will need to be rebased. I think everything else here looks pretty good.

hebasto:
Please rebase.

test/CMakeLists.txt
johnny9: @epicleafies

johnny9 (commented):
Tested the PR and it looks good however the CMakeList.txt changes still need to be addressed. Gmock shouldnt be optional, the build/test should fail if it is missing since we use it for validation

epicleafies:
GMock is now a requirement

johnny9 (approved):
tACK ccf5e7b18f4a127e1bfbd1f97d4f9666ad730cf5

tested on ubuntu, set proxy, set tor, and validated settings.json includes the config

Issue #512, opened 2026-02-28
Language and Display Unit Settings


johnny9:
## Summary
Add display settings for UI language and BTC/sat display unit parity.

## Complexity
- `L`

## Dependencies
- `01`
- `02`
- `07`

## Global Constraints
- All implementation lives under `gui-qml/`.
- No GUI runtime dependency on `bitcoin/src/qt` components. Reused logic must be moved/ported into `gui-qml/`.
- Do not change `interfaces::Wallet` or `interfaces::Node` in this version.
- Issue completion must satisfy `DefinitionOfDone.md`.

## Scope
- Expose language option and persistence (`-lang` compatible behavior).
- Expose display-unit setting and use it across send, receive, activity, and request history views.
- Remove ad-hoc per-widget unit toggling conflicts by defining clear precedence (global vs local temporary toggle).
- Ensure restart messaging where required.

## Acceptance Criteria (DoD-Aligned)
1. User can set language and display unit in settings UI.
2. Display unit updates all relevant screens consistently, including fees/totals/history.
3. Language selection behavior is deterministic across restart.
4. No regressions in amount parsing/formatting edge cases.

## Required Tests
- C++ unit tests: Unit tests for amount formatting/parsing service and settings persistence.
- QML QTest: QTests covering global unit changes across bound controls.
- Python E2E (`qml/test/testbridge.h`): E2E flow: change unit/language -> navigate major screens -> restart -> verify persisted behavior.

## Implementation Notes
- Likely requires replacing `qt/bitcoinunits` usage with gui-qml-local formatting utilities.

PR #536, opened 2026-03-16
Lang units settings

epicleafies:
Added the option to change the display unit under display settings, can choose between BTC or sats. #512 
Added the option to change language under display settings, currently only English and Spanish are translated.
All instances of bitcoin values are now tagged so the change with display unit settings (s for sats).
All text now tagged so that it will change with language (with the exception of wallet names).
Added unit, qml and functional tests for display units and language.

Screenshot from 2026-03-25 14-55-43

Screenshot from 2026-03-25 14-56-50

Screenshot from 2026-03-25 14-57-26

### DIFF
--- a/qml/components/BitcoinAmountInputField.qml
+++ b/qml/components/BitcoinAmountInputField.qml
@@ -92,13 +92,29 @@ ColumnLayout {
                 }
             }
 
-            CoreText {
+            Loader {
                 id: unitLabel
                 anchors.right: flipIcon.left
                 anchors.verticalCenter: parent.verticalCenter
-                text: root.amount ? root.amount.unitLabel : ""
-                font.pixelSize: 18
-                color: enabled ? Theme.color.neutral7 : Theme.color.neutral4
+                sourceComponent: root.amount && root.amount.unit === BitcoinAmount.SAT
+                    ? satoshiIconComponent
+                    : btcTextComponent
+            }
+            Component {
+                id: btcTextComponent
+                CoreText {
+                    text: root.amount ? root.amount.unitLabel : ""
+                    font.pixelSize: 18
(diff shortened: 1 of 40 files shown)


hebasto:
Please rebase.

qml/components/BitcoinAmountInputField.qml
```diff
@@ -92,13 +92,29 @@ ColumnLayout {
                 }
             }
 
-            CoreText {
+            Loader {
```
johnny9: I think loaders aren't really needed.

qml/components/BitcoinAmountInputField.qml
```diff
@@ -92,13 +92,29 @@ ColumnLayout {
...
+                CoreText {
+                    text: root.amount ? root.amount.unitLabel : ""
+                    font.pixelSize: 18
+                    color: root.enabled ? Theme.color.neutral7 : Theme.color.neutral4
+                }
+            }
+            Component {
+                id: satoshiIconComponent
+                CoreText {
+                    text: "s"
```
johnny9: I think the label for sats is going to be sat/sats

johnny9 (approved):
ACK 2dd902459e0038b228008294ff4af117aec4f009

I tested this out and it works well. For follow ups, we should get the system defaults to be the actual language of the system its running on and not just english. Should also tie BitcoinAmount's units with the option.

johnny9 (commented):
ACK bfe33593be336ddf4568beaf2ec9006dc1e8aea5

johnny9 (approved):

hebasto:
Please rebase.

johnny9 (approved):
ACK d2f2c9c50664fe3fad89e71b77c84913eb3729b8

Tested on Ubuntu by switching to Spanish and going between sats and BTC.

Issue #513, opened 2026-02-28
Desktop Tray and Window Behavior Settings


johnny9:
## Summary
Add desktop settings for tray icon behavior and minimize/close window handling parity.

## Complexity
- `L`

## Dependencies
- `01`
- `02`

## Global Constraints
- All implementation lives under `gui-qml/`.
- No GUI runtime dependency on `bitcoin/src/qt` components. Reused logic must be moved/ported into `gui-qml/`.
- Do not change `interfaces::Wallet` or `interfaces::Node` in this version.
- Issue completion must satisfy `DefinitionOfDone.md`.

## Scope
- Implement settings: show tray icon, minimize-to-tray, and close-to-tray/window behavior.
- Integrate behavior with application lifecycle/shutdown semantics.
- Platform guard non-desktop targets.

## Acceptance Criteria (DoD-Aligned)
1. Tray and minimize behavior matches configured settings on desktop.
2. Close action does not accidentally shutdown node when close-to-tray is enabled.
3. Behavior is persisted and restored on restart.

## Required Tests
- C++ unit tests: Unit tests for tray/window behavior state machine.
- QML QTest: QTests for settings UI and state-dependent labels/warnings.
- Python E2E (`qml/test/testbridge.h`): Desktop E2E flow covering minimize, close, reopen via tray, and shutdown correctness.

## Implementation Notes
- Implement in gui-qml backend wrappers; do not use `bitcoingui` widgets dependencies.

GBKS:
Are there any specific icons that need to be created for this functionality? I'd assume that different operating systems require different images sizes/formats/treatments.

johnny9:
I think we can reuse the current bitcoin icon for this and the menu can remain text only for now

PR #538, opened 2026-03-18
Desktop tray functionality and controls

epicleafies:
Added window behavior as a new option in settings. #513
Added window behavior page, allows the user to configure if the tray is visible, behavior on minimization and behavior on closing.
Tray icon matches dark and light mode in the app.
Tray icon gives option to show or quit when clicked on.
Added unit, qml and functional tests.
 
Screenshot from 2026-03-25 15-35-28

Screenshot from 2026-03-25 15-35-39

### DIFF
--- a/qml/models/desktoptrayiconcontroller.cpp
+++ b/qml/models/desktoptrayiconcontroller.cpp
@@ -1,0 +1,256 @@
+// Copyright (c) 2021-2026 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include <qml/models/desktoptrayiconcontroller.h>
+
+#include <QDebug>
+#include <QDir>
+#include <QFile>
+#include <QIcon>
+#include <QImage>
+#include <QMenu>
+#include <QStandardPaths>
+#include <QStringList>
+#include <QTextStream>
+#include <QTimer>
+
+DesktopTrayIconController::DesktopTrayIconController(QObject* parent)
+    : QObject(parent)
+    , m_tray_icon(new QSystemTrayIcon(this))
+{
+    // Tooltip is required by the StatusNotifierItem/AppIndicator protocol —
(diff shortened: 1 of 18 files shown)


epicleafies:
Needs testing on Windows and Mac. Some of the behavior is OS dependent.

hebasto:
Please rebase.

johnny9 (commented):
I like the settings options but lets try to use the QML SystemTrayIcon instead of implementing the platform-specific DBus notification apis. These will only work on linux platforms.

One problem with these tray icons is that Qt doesn't support the native versions on all platforms and this means we'll need to link Widgets in order to implement this on platforms that don't have native support which I think includes Windows

qml/models/desktoptrayiconcontroller.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <QMenu>
+
+DesktopTrayIconController::DesktopTrayIconController(QObject* parent)
+    : QObject(parent)
+    , m_tray_icon(new QSystemTrayIcon(this))
+{
+    m_tray_icon->setToolTip(QStringLiteral("Bitcoin Core"));
+
+    auto* menu = new QMenu();
+    auto* showAction = menu->addAction(tr("Show"));
```
MarnixCroes, line 17: Usually `Show` changes to `Minimize` when app is shown.
Maybe you can add this?

MarnixCroes (commented):
9bae920c6a02b039c146ad73fa0f3161d5c97073

When I put the app to light mode, the tray Icon is not visible. Only on hover it becomes slightly visible

Image

Image

johnny9 (approved):
ACK 9bae920c6a02b039c146ad73fa0f3161d5c97073

Tested on Ubuntu. Menu options and tray icon can be adjusted with follow up issues.

Edit:
issue added here https://github.com/bitcoin-core/gui-qml/issues/595 to track follow up

epicleafies:
Superseded by #746

Issue #508, opened 2026-02-28
RPC Command Console QML Surface


johnny9:
## Summary
Provide a QML-native RPC console using `interfaces::Node::executeRpc` and `listRpcCommands`.

## Complexity
- `L`

## Dependencies
- `01`
- `02`

## Global Constraints
- All implementation lives under `gui-qml/`.
- No GUI runtime dependency on `bitcoin/src/qt` components. Reused logic must be moved/ported into `gui-qml/`.
- Do not change `interfaces::Wallet` or `interfaces::Node` in this version.
- Issue completion must satisfy `DefinitionOfDone.md`.

## Scope
- Create RPC console page with input, history, output, help hinting, and clear behavior.
- Support command parsing with params and safe error display.
- Include anti-scam warning messaging equivalent to Qt behavior.
- Do not reuse `qt/rpcconsole.*`.

## Acceptance Criteria (DoD-Aligned)
1. Known safe RPC calls (e.g., `getnetworkinfo`) execute and render JSON output.
2. Invalid commands and malformed params show actionable errors.
3. Command history works via keyboard navigation and does not leak across app modes unexpectedly.

## Required Tests
- C++ unit tests: Unit tests for command parsing/execution adapter.
- QML QTest: QTests for history navigation, focus behavior, and rendering states.
- Python E2E (`qml/test/testbridge.h`): E2E flow: open console -> run valid command -> run invalid command -> verify output/error surfaces.

## Implementation Notes
- Long-running RPC calls must run off UI thread and marshal results back safely.

GBKS:
Initial designs are [here](https://bitcoincore.app/console/). Very sure they need to be iterated and finessed, for which it would be great to get some feedback from people who use the console frequently.

hebasto:
It seems prudent to keep [this](https://github.com/bitcoin-core/gui/issues/932) in mind during work on the implementation.

GBKS:
> It seems prudent to keep [this](https://github.com/bitcoin-core/gui/issues/932) in mind during work on the implementation.

Interesting one. QML might not have this problem, but we may also need to ask if it is actually useful to spit out ~4MB of raw data on the screen. That does not seem very useful or usable. An alternative could be to show a truncated version with the option to download the full data (as JSON or otherwise). What do users typically do with the raw transaction data?

hebasto:
> > It seems prudent to keep [this](https://github.com/bitcoin-core/gui/issues/932) in mind during work on the implementation.
> 
[...]

PR #540, opened 2026-03-25
RPC command console

epicleafies:
Created page for rpc command console. #508 
Created option under settings for console.
Preserved parsing logic currently in use.
Added unit and functional tests.

Screenshot from 2026-03-24 16-38-26

Screenshot from 2026-03-24 16-39-07


qml/pages/node/CommandConsole.qml
```diff
@@ -0,0 +1,415 @@
...
+        // the call until after the layout pass so positionViewAtEnd() sees the
+        // updated content height.
+        onCountChanged: Qt.callLater(positionViewAtEnd)
+    }
+
+    // Unfocus inputField on any click in the output area. Placed outside the
+    // ListView so Flickable's childMouseEventFilter does not interfere.
+    // z:1 puts it above the ListView (z:0); mouse.accepted = false passes the
+    // press through to the ListView so scrolling still works.
+    MouseArea {
```
johnny9: This mouse area causes issues with the scollbar. It seems also redundant as there is a TapHandler doing the same thing

qml/pages/node/CommandConsole.qml
```diff
@@ -0,0 +1,415 @@
...
+    TextMetrics {
+        id: timestampMetrics
+        font.family: "monospace"
+        font.pixelSize: 12
+        text: "[00:00:00]"
+    }
+
+    // Output area — ListView virtualizes delegates so only visible rows are
+    // instantiated, preventing layout churn during long console sessions.
+    ListView {
```
johnny9: The list view has problems with very long command responses. In my testing, the best fix is to convert this to a Flickable+Column+Repeater. With that we just need to make sure the command history buffer has a sensible limit.

johnny9 (commented):
Noticed issues with scrolling when command responses are large. Switch from a List View to columns fixes it due to the simpler layout mechanism of columns

qml/pages/node/CommandConsole.qml
```diff
@@ -0,0 +1,415 @@
...
+                    if (!activeFocus) {
+                        autocompletePopup.close()
+                    }
+                }
+            }
+
+            ContinueButton {
+                id: submitButton
+                objectName: "consoleSubmitButton"
+                text: qsTr("Run")
```
MarnixCroes: Isn't an icon better, than text?

Maybe an arrow? (with better color and style)

Image

MarnixCroes (commented):
10a6d50e278cda634e1ab80bd6d575a7d1a4d463

A few observations from first quick test:

- The entry box should automatically be selected when opening the console window.
- Selecting the entry box does not always work that well. I have to click it twice.
- Up and down keystroke should navigate suggested commands (when there are, for example type only `get`), instead of the history
- When selecting a suggested option using mouse arrow and clicking it, the entry box gets deselected. Which it shouldn't, so the user can select the command and then press enter. Without having to select the entry box again first and then press enter.

MarnixCroes:
@epicleafies
What is the status of the PR? I see some previous issues are addressed while some others are still present.

johnny9:
Related to https://github.com/bitcoin-core/gui-qml/issues/508

epicleafies:
Superseded by #746

Issue #543, opened 2026-03-31
Missing qml6-module-qtqml-workerscript on debian


pinheadmz:
Does it need to be added to the README for installation or am I missing something ...?

johnny9:
Did you need it to build locally? I actually think it needs to be removed from the github workflows because we don't use WorkerScripts in the qml

pinheadmz:
Yeah local build, got `“QtQml.WorkerScript is not installed”` following the current directions

johnny9:
It looks like it just is default needed with QtQuick even if it isn't used and it is packaged separately for some reason. Just need to add to the readme.

PR #544, opened 2026-03-31
ci: drop unused qtqml-workerscript package

johnny9:
This removes qml6-module-qtqml-workerscript from CI dependency install lists.

I checked qt6 and found no WorkerScript / QtQml.WorkerScript usage in QML, tests, or CMake, so this package appears unused.

Related to #543


johnny9:
QtQuick.WorkerScript is always auto imported with QtQuick so it is always needed despite being packaged separately

Issue #520, opened 2026-02-28
Feature Parity 16: Send Advanced Fee Controls: Custom Fee and Subtract Fee


johnny9:
## Summary
Add manual/custom fee entry and include/subtract-fee behavior parity.

## Complexity
- `XL`

## Dependencies
- `02`

## Global Constraints
- All implementation lives under `gui-qml/`.
- No GUI runtime dependency on `bitcoin/src/qt` components. Reused logic must be moved/ported into `gui-qml/`.
- Do not change `interfaces::Wallet` or `interfaces::Node` in this version.
- Issue completion must satisfy `DefinitionOfDone.md`.

## Scope
- Replace hardcoded fee rate in `WalletQmlModel::prepareTransaction` with model-driven settings.
- Add custom fee entry UI with validation and presets coexistence.
- Expose subtract-fee-from-amount in single and multi-recipient flows.
- Update review screen amounts (recipient totals, fee, effective send) correctly.

## Acceptance Criteria (DoD-Aligned)
1. User can switch between preset fee targets and custom fee input.
2. Subtract fee from amount toggles correctly and matches transaction result.
3. Validation handles insufficient funds, dust, and invalid fee input cleanly.

## Required Tests
- C++ unit tests: Unit tests for fee/subtract calculation and coin control mapping.
- QML QTest: QTests for fee UI states and subtract toggle behavior.
- Python E2E (`qml/test/testbridge.h`): E2E flow: send with custom fee + send with subtract-fee; validate review and broadcast outcomes.

## Implementation Notes
- Likely touches `qml/models/sendrecipient.*`, `qml/models/walletqmlmodel.*`, `qml/components/FeeSelection.qml`, send review pages.

https://bitcoincore.app/send/send-form-input-fields/

johnny9:
<img width="1600" height="732" alt="Image" src="https://github.com/user-attachments/assets/1d959b15-05d8-4818-a512-134d56f945a6" />

johnny9:
<img width="1600" height="1436" alt="Image" src="https://github.com/user-attachments/assets/0b7c94ae-02b1-4ec1-88ac-0d17dd13f969" />

johnny9:
standard fee spec

<img width="1600" height="1013" alt="Image" src="https://github.com/user-attachments/assets/5a0d1900-2252-47d8-a95e-95b7271dc52e" />

GBKS:
[...]

PR #546, opened 2026-04-02
Fee selection

johnny9:
image
image
image

Wires up fee selection in the Send page. Overrides are in place for Regtest where CoinControl fee estimations aren't supported.

### DIFF
--- a/qml/models/walletqmlmodel.h
+++ b/qml/models/walletqmlmodel.h
@@ -37,6 +37,9 @@ class WalletQmlModel : public QObject
     Q_PROPERTY(WalletQmlModelTransaction* currentTransaction READ currentTransaction NOTIFY currentTransactionChanged)
     Q_PROPERTY(unsigned int targetBlocks READ feeTargetBlocks WRITE setFeeTargetBlocks NOTIFY feeTargetBlocksChanged)
     Q_PROPERTY(QString estimatedFee READ estimatedFee NOTIFY estimatedFeeChanged)
+    Q_PROPERTY(bool customFeeEnabled READ customFeeEnabled WRITE setCustomFeeEnabled NOTIFY customFeeEnabledChanged)
+    Q_PROPERTY(QString customFeeRate READ customFeeRate WRITE setCustomFeeRate NOTIFY customFeeRateChanged)
+    Q_PROPERTY(bool customFeeRateValid READ customFeeRateValid NOTIFY customFeeRateValidChanged)
     Q_PROPERTY(bool feeEstimatePending READ feeEstimatePending NOTIFY feeEstimatePendingChanged)
     Q_PROPERTY(int feeEstimateRevision READ feeEstimateRevision NOTIFY feeEstimateRevisionChanged)
     Q_PROPERTY(bool isWalletLoaded READ isWalletLoaded NOTIFY walletIsLoadedChanged)
@@ -57,6 +60,9 @@ public:
     PaymentRequest* currentPaymentRequest() const { return m_current_payment_request; }
     WalletQmlModelTransaction* currentTransaction() const { return m_current_transaction; }
     QString estimatedFee() const;
+    bool customFeeEnabled() const { return m_custom_fee_enabled; }
+    QString customFeeRate() const { return m_custom_fee_rate; }
+    bool customFeeRateValid() const;
     bool feeEstimatePending() const { return m_fee_estimate_pending; }
     int feeEstimateRevision() const { return m_fee_estimate_revision; }
     Q_INVOKABLE bool prepareTransaction();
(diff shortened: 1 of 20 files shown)


johnny9:
Related to #520

MarnixCroes (commented):
build fails 5c7cd5152d86f25825ef68583509598a304a54be

```
In file included from /usr/include/x86_64-linux-gnu/qt6/QtTest/qtesteventloop.h:8,
                 from /usr/include/x86_64-linux-gnu/qt6/QtTest/qsignalspy.h:11,
                 from /usr/include/x86_64-linux-gnu/qt6/QtTest/QtTest:10,
                 from /home/marnix/projects/gui-qml/test/test_walletqmlmodel.cpp:5:
/home/marnix/projects/gui-qml/test/test_walletqmlmodel.cpp: In lambda function:
/home/marnix/projects/gui-qml/test/test_walletqmlmodel.cpp:409:9: error: return-statement with no value, in function returning ‘util::Result<std::shared_ptr<const CTransaction> >’ [-fpermissive]
  409 |         QCOMPARE(recipients.size(), 1U);
      |         ^~~~~~~~
gmake[2]: *** [test/CMakeFiles/bitcoinqml_unit_tests.dir/build.make:216: test/CMakeFiles/bitcoinqml_unit_tests.dir/test_walletqmlmodel.cpp.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:2075: test/CMakeFiles/bitcoinqml_unit_tests.dir/all] Error 2
gmake: *** [Makefile:146: all] Error 2
```

johnny9:
fix the compile issue in the unittest

MarnixCroes (commented):
a4de80bd506fc72317f3ac8f6d9a80e3b2bfde32

some comments from first test (some maybe for follow up):

- Fee amount is not displayed in sats in case this is active
- The Fee Rate entry is not limited when no decimal is used (i.e. it accepts an unlimited amount of characters as long as no comma is used)
- When entering full balance, and not `Include fee in amount` the _Review_ button doesn't work. Which makes sense, but there is no clear user feedback on why.

pseudoramdom:
[a4de80b](https://github.com/bitcoin-core/gui-qml/commit/a4de80bd506fc72317f3ac8f6d9a80e3b2bfde32)

Works well. 
Would it be good to include fee rate in each of the option - high, default, low? (doesn't have to be in this PR, we can track this separately)

Screenshot 2026-04-22 at 11 40 26 AM

johnny9:
> [a4de80b](https://github.com/bitcoin-core/gui-qml/commit/a4de80bd506fc72317f3ac8f6d9a80e3b2bfde32)
> 
> Works well. Would it be good to include fee rate in each of the option - high, default, low? (doesn't have to be in this PR, we can track this separately)
> Screenshot 2026-04-22 at 11 40 26 AM

The fees are based on number of blocks. I think the fee rate might clutter up the information in a dropdown a bit too much but there might be an argument for the fee rate being more important than the fee itself but that migth just be for people that know bitcoin too well.

qml/models/walletqmlmodel.h
```diff
@@ -97,6 +130,17 @@ class WalletQmlModel : public QObject
...
+    QThread* m_fee_estimation_thread{nullptr};
+    QTimer* m_fee_estimation_timer{nullptr};
+    QHash<unsigned int, QString> m_fee_estimates;
+    QString m_custom_fee_estimate;
+    QString m_custom_fee_rate;
+    QString m_preview_change_address;
+    quint64 m_fee_estimate_request_id{0};
+    int m_fee_estimate_revision{0};
+    bool m_custom_fee_enabled{false};
+    bool m_fee_estimate_pending{false};
```
pseudoramdom, line 143: Do you think it'll be better to extract the fee estimation related code to its own model, say `feeselectionmodel`. 
This way both send and RBF flows can use it

johnny9: Yes I think that makes sense. Probably some association per transaction with the current_transaction being the one in the send dialog and maybe another for a chosen RBF transaction? I'm not totally sure where it would land but just at a glance it looks like it might be a needed refactor.

qml/pages/wallet/Send.qml
```diff
@@ -308,18 +349,57 @@ PageStack {
...
+                    Icon {
+                        source: "image://images/check"
+                        size: 18
+                        color: Theme.color.green
+                    }
+
+                    CoreText {
+                        objectName: "sendFeeIncludedNoteText"
+                        Layout.fillWidth: true
+                        text: qsTr("Fees are included in the amount")
```
MarnixCroes: In the dropdown it is singular `Fee`, where here it is plural.
Best if they match. All singular is best imo

MarnixCroes (approved):
ACK f3ca7a0f28cdf4ee9221dac7fdb835317a277706

---
minor issues can be in follow-up
For example, below dust threshold handling

Image

PR #552, opened 2026-04-22
QtTest now see GMock Failures

epicleafies:
Fixed bug where GMock failures where not being seen by QtTest because there were no QtTest assertions leading to test cases passing even with GMock errors. Added QVERIFY at the end of tests so that QtTest is aware of the GMock output. Added GMock initialization to the test's main(), GMock partially works without it but this makes sure GMock's global state is working correctly. 

Fixed bug in OptionsModelTests::onboardWritesProxy where updateRwSetting being called with something other than proxy first would cause a GMock error (previously hidden because of above bug). Added catch-all so other updateRwSetting calls (ex. listen) wouldn't lead to a failing test case, as the test case should only fail in absence of proxy but not in the presence of other calls.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -11,6 +11,7 @@ endif()
 
 add_executable(bitcoinqml_unit_tests
   test_unit_tests_main.cpp
+  gmocktestfixture.h
   test_bitcoinamount.cpp
   test_peerlistmodel.cpp
   test_peerstatsutil.cpp
--- a/test/gmocktestfixture.h
+++ b/test/gmocktestfixture.h
@@ -1,0 +1,99 @@
+// Copyright (c) 2026 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef BITCOIN_QML_TEST_GMOCKTESTFIXTURE_H
+#define BITCOIN_QML_TEST_GMOCKTESTFIXTURE_H
+
+#include <QObject>
+#include <QtTest/QtTest>
+
+#ifdef Assert
+#pragma push_macro("Assert")
+#undef Assert
+#define BITCOIN_QML_RESTORE_ASSERT_MACRO
+#endif
+
+#include <gmock/gmock.h>
+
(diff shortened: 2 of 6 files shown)


johnny9 (commented):
Concept ACK

This is a good change but we should do another step or two further. I found that we need to add an additional check to make sure StrictMock failures are caught properly as well. 

"QVERIFY(Mock::VerifyAndClearExpectations(...)) is not a complete bridge between gmock and QTest: it catches unmet expectations on that specific mock, but it does not reliably convert all gmock failures, such as StrictMock uninteresting calls, into QtTest failures. The cleaner fix is to keep QTest, initialize gmock once in the shared test runner, and add a small harness-level bridge using a GoogleTest event listener plus a shared Qt base fixture so gmock failures are collected centrally and failed automatically in each test’s cleanup(), without repeating manual verify calls in every test."

This produces a nice reusable harness that appears to catch all of the GMock errors. Then we just need to modify each test to use that harness/fixture.

johnny9 (commented):
ACK 684b13ddeadaa43a65a4dfccca49e15ec62cef79.


Verified locally and this fixes QtTest passing even when GMock expectations fail. It correctly causes Mocks to fail if an expect isn't met.

```
-- a/qml/models/options_model.cpp
+++ b/qml/models/options_model.cpp
@@ -156,9 +156,9 @@ void OptionsQmlModel::setProxyEnabled(bool enabled)
         m_proxy_enabled = enabled;
         if (m_onboarded) {
             if (enabled && !m_proxy_address.isEmpty()) {
-                m_node.updateRwSetting("proxy", m_proxy_address.toStdString());
+                //m_node.updateRwSetting("proxy", m_proxy_address.toStdString());
             } else {
-                m_node.updateRwSetting("proxy", common::SettingsValue{});
+                //m_node.updateRwSetting("proxy", common::SettingsValue{});
             }
         }
         if (proxySettingsDirty() != was_dirty) {
```
This patch fails the tests when it previously wouldn't due to the mocks not triggering a fail in QTest.

```
cmake --build build --target bitcoinqml_unit_tests -j2
./bin/bitcoinqml_unit_tests -platform offscreen
```

johnny9 (approved):

Issue #518, opened 2026-02-28
Generation, History, QR, and BIP21 Sharing


johnny9:
## Summary
Replace placeholder receive flow with real wallet-backed payment request generation and sharing.

## Complexity
- `L`

## Dependencies
- `01`
- `10`
- `02`

## Global Constraints
- All implementation lives under `gui-qml/`.
- No GUI runtime dependency on `bitcoin/src/qt` components. Reused logic must be moved/ported into `gui-qml/`.
- Do not change `interfaces::Wallet` or `interfaces::Node` in this version.
- Issue completion must satisfy `DefinitionOfDone.md`.

## Scope
- Generate real receive addresses via wallet interface.
- Persist receive request entries using current wallet API (`getAddressReceiveRequests` / `setAddressReceiveRequest`).
- Create recent requests/history model in gui-qml (no dependency on `qt/recentrequeststablemodel`).
- Support copy/share for address, QR payload, and BIP21 URI.

## Acceptance Criteria (DoD-Aligned)
1. Receive form creates a real address and stores request metadata.
2. History survives app restart and supports selection/copy/remove.
3. QR and BIP21 payloads match entered amount/label/message.
4. Malformed persisted entries do not crash startup and are safely skipped with diagnostics.

## Required Tests
- C++ unit tests: Unit tests for request serialization/deserialization and corruption handling in gui-qml model.
- QML QTest: QTests for receive form, history list, and share actions.
- Python E2E (`qml/test/testbridge.h`): E2E flow: create request -> verify history -> restart -> verify persistence -> copy QR/URI.

## Implementation Notes
- No wallet interface changes in this version; keep compatibility with existing raw storage API.

PR #553, opened 2026-04-25
Receive requests page 

epicleafies:
Redesigned the receive page and added the receive details page.
Moved QR code and receive address to the receive details page.
Added BIP21 URI as dominant receive request format.
Added receive requests to the activity log page.
Added unit, qml and functional tests for the receive functionality.
Note: the sharing functionality was decided as out of scope for this PR, the button exists but is disabled.
#518 

image

image

image

image

image


johnny9:
Hold off on redesigning the Receive page. The Payment request page is good and showing the requests in the Activity list is key. 

I will test/review to understand your "View previous payment requests" label and how the relationship between the Activity list and the Payment request page

johnny9 (commented):
Ok, I think we want to keep the Receive just a single page so lets drop the redesign of that from the PR. I think the link the view previous payment requests is good and should be added to the Receive page.

Then, in the Activity list, when you click on a Pending payment request, pull up the Payment request details page that you created instead of the regular transaction list. Even better is if you can join the list of received transactions and the payment requests so that you can click on a received transaction and then have a list of the payment requests that might be related.

johnny9:
Pulled into #708

PR #557, opened 2026-04-26
Add PSBT import review flow

johnny9:
Screenshot from 2026-04-26 00-21-06
Screenshot from 2026-04-26 00-22-06

Import a PSBT that we can fully sign and move to the SendReivew page. PSBT requiring a signature we can't provide is not supported right now. I think that should be designed around the future multisig flows.

Also, since I was in the send popup, I added the clear form row and qmltests to validate.

### DIFF
--- a/qml/controls/EllipsisMenuActionItem.qml
+++ b/qml/controls/EllipsisMenuActionItem.qml
@@ -1,0 +1,82 @@
+// Copyright (c) 2025 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import org.bitcoincore.qt 1.0
+
+AbstractButton {
+    id: root
+
+    property url leftIconSource
+    property color bgDefaultColor: "transparent"
+    property color bgHoverColor: Theme.color.neutral2
+    property color textColor: Theme.color.neutral7
+    property color textHoverColor: Theme.color.neutral9
+
+    hoverEnabled: AppMode.isDesktop
+    padding: 0
+
+    implicitWidth: 280
+    implicitHeight: 33
+
+    MouseArea {
+        anchors.fill: parent
+        enabled: false
+        hoverEnabled: true
+        cursorShape: Qt.PointingHandCursor
+    }
+
+    contentItem: RowLayout {
+        anchors.fill: parent
+        anchors.leftMargin: 10
+        anchors.rightMargin: 5
+        spacing: 5
+
+        RowLayout {
+            Layout.alignment: Qt.AlignVCenter
+            Layout.fillWidth: true
+            spacing: 7
+
+            Item {
+                Layout.alignment: Qt.AlignVCenter
+                Layout.preferredWidth: 18
+                Layout.preferredHeight: 18
+
+                Icon {
+                    anchors.centerIn: parent
+                    source: root.leftIconSource
+                    color: root.hovered ? root.textHoverColor : root.textColor
+                    size: 18
+                }
+            }
+
+            CoreText {
+                id: buttonText
+                Layout.fillWidth: true
+                Layout.alignment: Qt.AlignVCenter
+                text: root.text
+                horizontalAlignment: Text.AlignLeft
+                font.pixelSize: 15
+                color: root.hovered ? root.textHoverColor : root.textColor
+            }
+        }
+
+        Item {
+            Layout.alignment: Qt.AlignVCenter
+            Layout.preferredWidth: 18
+            Layout.preferredHeight: 18
+        }
+    }
+
+    background: Rectangle {
+        color: root.hovered ? root.bgHoverColor : root.bgDefaultColor
+        radius: 0
+
+        Behavior on color {
+            ColorAnimation { duration: 150 }
+        }
+    }
+}
--- a/qml/controls/EllipsisMenuToggleItem.qml
+++ b/qml/controls/EllipsisMenuToggleItem.qml
@@ -55,11 +55,7 @@ Button {
             Layout.preferredWidth: 40
             Layout.preferredHeight: 24
             checked: root.checked
-
-            MouseArea {
-                anchors.fill: parent
-                onClicked: root.toggle()
-            }
+            enabled: false
         }
     }
 
--- a/qml/controls/SendOptionsPopup.qml
+++ b/qml/controls/SendOptionsPopup.qml
@@ -18,6 +18,7 @@ OptionPopup {
     property alias multipleRecipientsEnabled: multipleRecipientsToggle.checked
 
     signal importPsbtFromFileRequested()
+    signal clearFormRequested()
 
     implicitWidth: 305
     implicitHeight: columnLayout.implicitHeight + 10
@@ -36,6 +37,12 @@ OptionPopup {
             id: coinControlToggle
             objectName: "sendOptionsCoinControlToggle"
             Layout.fillWidth: true
+            Layout.preferredHeight: 33
+            Layout.minimumHeight: 33
+            Layout.maximumHeight: 33
+            bgRadius: 0
+            contentTopPadding: 4
+            contentBottomPadding: 5
             text: qsTr("Enable Coin control")
         }
 
@@ -43,6 +50,12 @@ OptionPopup {
             id: multipleRecipientsToggle
             objectName: "sendOptionsMultipleRecipientsToggle"
             Layout.fillWidth: true
+            Layout.preferredHeight: 33
+            Layout.minimumHeight: 33
+            Layout.maximumHeight: 33
+            bgRadius: 0
+            contentTopPadding: 4
+            contentBottomPadding: 5
             text: qsTr("Multiple Recipients")
         }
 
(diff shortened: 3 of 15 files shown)


test/functional/qml_test_psbt.py
```diff
@@ -0,0 +1,240 @@
...
+        fixture.write(base64.b64decode(response["psbt"]))
+    return fixture_path
+
+
+def navigate_to_send(gui):
+    gui.click("sendTabButton")
+    gui.wait_for_property("sendOptionsButton", "visible", True, timeout_ms=10000)
+
+
+def import_psbt(gui, psbt_path):
```
epicleafies: Suggestion: good end-to-end coverage. Worth adding: corrupt/undecodable PSBT, the base64-armored fallback path (walletqmlmodel.cpp:1582), file-open failure, the >25 recipient cap, and the own-inputs-but-can't-sign rejection (walletqmlmodel.cpp:1644) — I verified that last one reachable via a watch-only wallet.

qml/models/walletqmlmodel.h
```diff
@@ -71,9 +77,20 @@ class WalletQmlModel : public QObject
     Q_PROPERTY(QString transactionError READ transactionError NOTIFY transactionErrorChanged)
     Q_PROPERTY(bool transactionNeedsUnlock READ transactionNeedsUnlock NOTIFY transactionNeedsUnlockChanged)
     Q_PROPERTY(QString settingsError READ settingsError NOTIFY settingsErrorChanged)
+    Q_PROPERTY(bool importedPsbtLoaded READ importedPsbtLoaded NOTIFY importedPsbtChanged)
```
epicleafies, line 80: Suggestion: of these 10 new properties, only `importedPsbtError` (L83) is consumed by QML — the other nine are unused (they're for the #663 operations screen). `importedPsbtLoaded()` keys off `m_imported_psbt_mode == "draft"`, which is never produced, so it's always false. Either reference #663 so this reads as intentional scaffolding, or land them with their consumer.

qml/models/walletqmlmodel.h
```diff
@@ -123,6 +140,12 @@ class WalletQmlModel : public QObject
     Q_INVOKABLE void clearSettingsError();
     Q_INVOKABLE void setDefaultReceiveAddressType(const QString& address_type);
     Q_INVOKABLE QString receiveAddressTypeLabel(const QString& address_type) const;
+    Q_INVOKABLE QString importPsbtFromFile(const QString& path);
+    Q_INVOKABLE void clearImportedPsbt();
+    Q_INVOKABLE void signImportedPsbt();
```
epicleafies, line 145: Suggestion: `signImportedPsbt` / `broadcastImportedPsbt` / `copyImportedPsbtToClipboard` / `saveImportedPsbtToFile` have no QML caller yet — they're the backend for #663. Only `importPsbtFromFile` (L143) and `clearImportedPsbt` (L144) are used in this PR.

qml/models/walletqmlmodel.cpp
```diff
@@ -1439,6 +1534,346 @@ bool WalletQmlModel::sendTransactionInternal()
     return true;
 }
 
+void WalletQmlModel::setImportedPsbtError(const QString& error)
+{
+    m_imported_psbt.reset();
```
epicleafies, line 1539: Suggestion (root cause): `m_imported_psbt` is only ever `.reset()` (here and L1554) and never assigned a value, so every method that dereferences it (sign/broadcast/copy/save/refresh/summary) early-returns and is unreachable in this PR. This is the foundation for #663 (PSBT operations screen). Suggest referencing #663/#664 in the PR + a commit, or landing this backend together with its consumer so it's exercised.

qml/models/walletqmlmodel.cpp
```diff
@@ -1439,6 +1534,346 @@ bool WalletQmlModel::sendTransactionInternal()
...
+    if (complete) {
+        refreshImportedPsbtState(tr("PSBT signed. Transaction is ready for broadcast."));
+    } else if (signed_inputs > 0) {
+        refreshImportedPsbtState(tr("Signed %1 input(s). More signatures are still required.").arg(static_cast<int>(signed_inputs)));
+    } else {
+        refreshImportedPsbtState(tr("This wallet could not add any signatures to the PSBT."));
+    }
+}
+
+void WalletQmlModel::broadcastImportedPsbt()
```
epicleafies, line 1819: Suggestion (unreachable, no consumer yet): #663's broadcast action. Adds untested broadcast surface until wired up.

qml/models/walletqmlmodel.cpp
```diff
@@ -1439,6 +1534,346 @@ bool WalletQmlModel::sendTransactionInternal()
...
+void WalletQmlModel::copyImportedPsbtToClipboard()
+{
+    if (!m_imported_psbt) {
+        return;
+    }
+    QGuiApplication::clipboard()->setText(SerializePsbtBase64(*m_imported_psbt));
+    refreshImportedPsbtState(tr("PSBT copied to clipboard."));
+}
+
+void WalletQmlModel::saveImportedPsbtToFile(const QString& path)
```
epicleafies, line 1859: Suggestion (unreachable, no consumer yet): #663's save action. Adds untested arbitrary-path file-write surface until wired up.

qml/models/walletqmlmodel.cpp
```diff
@@ -1439,6 +1534,346 @@ bool WalletQmlModel::sendTransactionInternal()
...
+            reason = tr("Only PSBTs that spend this wallet's own inputs are supported right now.");
+            return false;
+        }
+    }
+
+    PartiallySignedTransaction signed_psbt{psbt};
+    bool complete{false};
+    size_t signed_inputs{0};
+    const std::optional<common::PSBTError> fill_error{
+        m_wallet->fillPSBT(std::nullopt, /*sign=*/true, /*bip32derivs=*/true, &signed_inputs, signed_psbt, complete)};
```
epicleafies, line 1634: Suggestion: `fillPSBT(sign=true)` runs before the cheaper rejections — multisig (L1638), standard-output (L1658), recipient cap (L1673). A multisig/non-standard PSBT is fully signed into a throwaway copy and only then rejected. Consider doing the static checks before the signing fill.

qml/models/walletqmlmodel.cpp
```diff
@@ -1439,6 +1534,346 @@ bool WalletQmlModel::sendTransactionInternal()
...
+        const QString address{QString::fromStdString(EncodeDestination(destination))};
+        draft_recipients.push_back({address, getAddressLabel(address), output.nValue});
+        recipient_total += output.nValue;
+    }
+
+    if (draft_recipients.empty()) {
+        reason = tr("The PSBT does not have any recipient outputs to review.");
+        return false;
+    }
+    if (draft_recipients.size() > 25) {
```
epicleafies, line 1673: Suggestion: magic number `> 25`. If this mirrors the `SendRecipientsList` limit, reference that constant; otherwise name it (e.g. `MAX_REVIEWABLE_RECIPIENTS`) so it can't drift from the real UI cap. #664 plans broader recipient sets, so naming it now makes that change a one-liner.

qml/models/walletqmlmodel.cpp
```diff
@@ -1439,6 +1534,346 @@ bool WalletQmlModel::sendTransactionInternal()
...
+QString WalletQmlModel::importPsbtFromFile(const QString& path)
+{
+    const QString file_path{LocalFilePath(path)};
+    QFile file(file_path);
+    if (!file.open(QIODevice::ReadOnly)) {
+        setImportedPsbtError(tr("Could not open PSBT file: %1").arg(file.errorString()));
+        return QStringLiteral("unsupported");
+    }
+
+    const QByteArray bytes{file.readAll()};
```
epicleafies, line 1576: Suggestion: `file.readAll()` is unbounded. A sanity size cap before reading avoids pulling an arbitrarily large file into memory.

qml/models/walletqmlmodel.cpp
```diff
@@ -278,11 +296,88 @@ QString OutputTypeDescription(OutputType type)
...
+    const PSBTInput& input{psbt.inputs[index]};
+    if (IsMultisigScript(input.redeem_script) || IsMultisigScript(input.witness_script)) {
+        return true;
+    }
+
+    CTxOut utxo;
+    return psbt.GetInputUTXO(utxo, index) && IsMultisigScript(utxo.scriptPubKey);
+}
+
+bool DecodePsbtFromBytes(const QByteArray& bytes, PartiallySignedTransaction& psbt, std::string& error)
```
epicleafies, line 350: Nit: this copies the `QByteArray` into a `std::vector<std::byte>` byte-by-byte. You can span it directly (`reinterpret_cast<const std::byte*>(bytes.constData())` / `std::as_bytes`) and drop the loop.

qml/models/walletqmlmodel.cpp
```diff
@@ -278,11 +296,88 @@ QString OutputTypeDescription(OutputType type)
...
+QString LocalFilePath(const QString& path)
+{
+    const QUrl url(path);
+    if (url.isLocalFile()) {
+        return url.toLocalFile();
+    }
+    return path;
+}
+
+QString FormatBtc(CAmount amount)
```
epicleafies, line 309: Nit: `FormatBtc` hardcodes the BTC unit, ignoring the user's display setting. It's currently only reached by unreachable code (`buildPsbtSummary`), so two paths: (1) remove it together with that code; or (2) if it's kept as #663 scaffolding or later restored, fix `FormatBtc` to honor the active display unit (e.g. format via `QmlBitcoinUnits` with the user's selected unit) instead of hardcoding BTC.

qml/pages/wallet/Send.qml
```diff
@@ -87,8 +88,91 @@ PageStack {
...
+            title: qsTr("Import PSBT")
+            fileMode: FileDialog.OpenFile
+            nameFilters: [qsTr("Partially Signed Bitcoin Transactions (*.psbt)"), qsTr("All files (*)")]
+            onAccepted: sendPage.handlePsbtImportResult(root.wallet.importPsbtFromFile(selectedFile.toString()))
+        }
+
+        // Kept hidden so functional tests can inject a PSBT path until the
+        // native file dialog is automatable through the QML test bridge.
+        TextField {
+            id: psbtAutomationPathField
```
epicleafies, line 119: Suggestion: this hidden `psbtAutomationPathField` ships in the release binary purely so functional tests can inject a path (consumed at L247-255). A hidden field that triggers an arbitrary-path file import in shipping QML is a smell — consider gating it behind `ENABLE_TEST_AUTOMATION` (the bridge is already compiled conditionally) so it doesn't exist in release builds.

qml/pages/wallet/Send.qml
```diff
@@ -97,7 +181,7 @@ PageStack {
             onMultipleRecipientsEnabledChanged: {
                 if (!multipleRecipientsEnabled) {
                     root.wallet.recipients.clearToFront()
-                } else {
+                } else if (root.wallet.recipients.count === 1) {
```
epicleafies, line 184: Question: the `else` → `else if (recipients.count === 1)` change is needed so import doesn't add an extra recipient on top of populated ones — good, but the rationale isn't obvious; a one-line comment would help. Is `count === 1` the right guard if the list could ever be at 0?

qml/controls/EllipsisMenuToggleItem.qml
```diff
@@ -55,11 +55,7 @@ Button {
             Layout.preferredWidth: 40
             Layout.preferredHeight: 24
             checked: root.checked
-
-            MouseArea {
-                anchors.fill: parent
-                onClicked: root.toggle()
-            }
+            enabled: false
```
epicleafies, line 58: Nit: switching the inner switch to `enabled: false` correctly removes the double-toggle and lets the outer Button drive `checked`. Just confirm it doesn't render the switch as greyed/disabled in a way that reads as an unavailable option rather than a decorative indicator.

qml/controls/EllipsisMenuActionItem.qml
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) 2025 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import org.bitcoincore.qt 1.0
+
+AbstractButton {
```
epicleafies, line 10: Nit (a11y): this interactive row sets no `Accessible.name` / `Accessible.role`. Given the mobile/assistive targets, add `Accessible.role: Accessible.Button` and `Accessible.name: text`.

qml/controls/EllipsisMenuActionItem.qml
```diff
@@ -0,0 +1,82 @@
...
+                Layout.fillWidth: true
+                Layout.alignment: Qt.AlignVCenter
+                text: root.text
+                horizontalAlignment: Text.AlignLeft
+                font.pixelSize: 15
+                color: root.hovered ? root.textHoverColor : root.textColor
+            }
+        }
+
+        Item {
```
epicleafies, line 67: Nit: the trailing empty 18×18 `Item` has no comment — a brief note ("reserved for right-side affordance/icon") clarifies why it's here.

qml/controls/SendOptionsPopup.qml
```diff
@@ -16,36 +17,94 @@ OptionPopup {
     property alias coinControlEnabled: coinControlToggle.checked
     property alias multipleRecipientsEnabled: multipleRecipientsToggle.checked
 
-    implicitWidth: 300
-    implicitHeight: columnLayout.implicitHeight + 20
+    signal importPsbtFromFileRequested()
+    signal clearFormRequested()
+
+    implicitWidth: 305
```
epicleafies, line 23: Nit: `implicitWidth: 305` / `implicitHeight: + 10` plus the per-row `Layout.preferredHeight: 33` literals are scattered magic numbers; factoring the row height into a property would ease future tweaks.

test/functional/qml_test_psbt.py
epicleafies, line 105: Suggestion: good end-to-end coverage. Worth adding: corrupt/undecodable PSBT, the base64-armored fallback path (walletqmlmodel.cpp:1582), file-open failure, the >25 recipient cap, and the own-inputs-but-can't-sign rejection (walletqmlmodel.cpp:1644) — I verified that last one reachable via a watch-only wallet.

epicleafies (commented):
tested ACK c144a33

Unit tests pass and the functional suite passes (only `qml_test_wallet_migration.py` is skipped as it needs a legacy bitcoind, unrelated to this PR)
Also manually tested on regtest (Pop!_OS/COSMIC): fully-signable PSBTs route to single/multiple SendReview and broadcast + confirm at the recipient, while foreign-input and own-inputs-but-can't-sign (watch-only wallet) PSBTs both show the "unsupported" modal.

Main issue is the scaffolding code which is currently unreachable, is for #663, but its not referenced in this PR. There are also some smaller suggestions and nits. 

Pre-existing issues, not introduced here (noted while testing): the import flow lands in two existing rough edges in untouched files — "View transaction" is a no-op pop-to-root (`main.qml:166`), and multi-recipient sends render as N activity rows with the fee on one (`activitylistmodel.cpp:177`).

pseudoramdom:
We can close this in favor of https://github.com/bitcoin-core/gui-qml/pull/750

pseudoramdom:
> We can close this in favor of #750

@epicleafies Can you move the review to #750 ? Sorry for the trouble.

PR #559, opened 2026-04-27
Activity search filter

johnny9:
Based on #553 to include RequestHistoryModel.

Adds a toggle to open up text search and drop downs for time and type filters.

The toggle state is saved as a QSetting.

Added an export activity to csv as well.

image
image
image
image
image


MarnixCroes (commented):
94153d48a6b999e756e32ce38a71a9f7b5dbec33

- There is an issue when `Create payment request` does not work at first try. You have to change dialog, then it works

https://github.com/user-attachments/assets/3a052ea1-c249-41d7-ba25-9b081a2578d0

- `View previous requests` should be grayed out when there are none.
- I think the default save location of the CSV should not be `build/bin`
- I guess the message should be displayed here as well, when there is one?

Image

- For created payment requests, I think it does not make sense to display 0 confirmation count? It's a request, not a pending incoming tx

johnny9:
#708

PR #561, opened 2026-04-29
Implement Addresses list and Sign/Verify in Wallet Settings

### LINKED ISSUE #519: Feature Parity 15: Address Book and Contacts Management
## Summary
Deliver address book and contacts CRUD with send/receive integration.

## Complexity
- `XL`

## Dependencies
- `10`
- `02`

## Global Constraints

## Scope
- Create address/contacts list model using wallet address APIs.
- Add add/edit/delete contact flows with purpose tags where relevant.
- Integrate contact selection into send and receive flows.
- Support search/filter and duplicate validation.

1. User can create, edit, delete contacts and see updates reflected in send/receive views.
2. Invalid addresses and duplicate entries are blocked with clear messages.
3. Contact labels flow through activity/history where applicable.

## Required Tests
- C++ unit tests: Unit tests for contact model CRUD and validation.
- QML QTest: QTests for contact form and list interactions.
- Python E2E (`qml/test/testbridge.h`): E2E flow: add contact -> send using contact -> verify label propagation.

## Implementation Notes
- Keep model and UI fully inside gui-qml.

### LINKED ISSUE #524: Feature Parity 20: Sign and Verify Message Flow
## Summary
Provide message signing and verification screens with parity behavior and warnings.

## Complexity
- `M`

## Dependencies
- `10`
- `15`
- `02`

## Global Constraints

## Scope
- Add sign message flow using wallet sign API for eligible addresses.
- Add verify message flow using local verification utilities.
- Include phishing/safety warning copy and reset/copy helpers.
- Integrate address selection from contacts/address book when available.

1. User can sign message for valid wallet-owned signing address and copy signature.
2. User can verify signature/address/message combinations with correct success/failure states.
3. Unsupported address types are clearly explained.

## Required Tests
- C++ unit tests: Unit tests for sign/verify adapter and validation logic.
- QML QTest: QTests for input validation, success/error states, and reset actions.
- Python E2E (`qml/test/testbridge.h`): E2E flow for valid sign/verify and invalid signature failure case.

## Implementation Notes
- No reliance on `qt/signverifymessagedialog.*`.

pseudoramdom:
Add support to view receive and change addresses. (Fixes https://github.com/bitcoin-core/gui-qml/issues/519)

- Add `addresslistmodel` 
  - exposes wallet-owned receive and change addresses to QML including labels, usage state, balances, formatter amounts, script type etc.
  - This will be owned by `WalletQmlModel` which provides helper methods related to addresses.
- Add a reusable control `SegmentedPicker.qml` for segmented picker with multiple options.

Implement Sign/verify message page in Wallet Settings. Fixes https://github.com/bitcoin-core/gui-qml/issues/524

Add SignVerifyMessageModel to coordinate signing/verifying and will be owned
by WalletQmlModel.
This change was dependent on generating a legacy P2PKH addresses which
was missing in the receive flow. Implemented a picker to choose address types.
Implement functional, unit tests.

#### Tests
- Updated functional wallet harness so app-only GUI tests do not require a built bitcoind.

### Screenshots

<details>
<summary>Address list</summary>

Single use address

Change address

Screenshot 2026-05-18 at 5 17 45 PM

</details> 

<details>
<summary> Sign / Verify message </summary>

Screenshot 2026-05-18 at 4 54 42 PM

Screenshot 2026-05-18 at 4 55 07 PM

Screenshot 2026-05-18 at 4 55 03 PM

</details>

### DIFF
--- a/qml/models/signverifymessagemodel.cpp
+++ b/qml/models/signverifymessagemodel.cpp
@@ -1,0 +1,171 @@
+// Copyright (c) 2026 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include <qml/models/signverifymessagemodel.h>
+
+#include <common/signmessage.h>
+#include <key_io.h>
+#include <qml/models/walletunlock.h>
+#include <qml/util.h>
+#include <support/allocators/secure.h>
+
+namespace {
+std::optional<PKHash> LegacyP2PKHFromAddress(const QString& address)
+{
+    const CTxDestination destination{DecodeDestination(address.trimmed().toStdString())};
+    if (const auto* pkhash{std::get_if<PKHash>(&destination)}) {
+        return *pkhash;
+    }
+    return std::nullopt;
+}
+} // namespace
+
+SignVerifyMessageModel::SignVerifyMessageModel(interfaces::Wallet* wallet, QObject* parent)
+    : QObject(parent), m_wallet(wallet)
+{
+}
+
+void SignVerifyMessageModel::setWallet(interfaces::Wallet* wallet)
+{
+    if (m_wallet == wallet) {
+        return;
+    }
+    m_wallet = wallet;
+    clear();
+}
+
+void SignVerifyMessageModel::setSecurityStateChangedFn(SecurityStateChangedFn fn)
+{
+    m_security_state_changed = std::move(fn);
+}
+
+bool SignVerifyMessageModel::isLegacyP2PKHAddress(const QString& address) const
+{
+    return LegacyP2PKHFromAddress(address).has_value();
+}
+
+bool SignVerifyMessageModel::signMessage(const QString& address, const QString& message)
+{
+    return signMessageInternal(address, message, std::nullopt);
+}
+
+bool SignVerifyMessageModel::signMessageWithPassphrase(const QString& address, const QString& message, const QString& passphrase)
+{
+    return signMessageInternal(address, message, passphrase);
+}
+
+bool SignVerifyMessageModel::signMessageInternal(const QString& address, const QString& message, const std::optional<QString>& passphrase)
+{
+    clearSigningStatus();
+    setSignature(QString());
+
+    if (!m_wallet) {
+        setSigningStatus(tr("No wallet is selected."));
+        return false;
+    }
+    const auto pkhash{LegacyP2PKHFromAddress(address)};
+    if (!pkhash) {
+        setSigningStatus(tr("Enter a legacy P2PKH bitcoin address."));
+        return false;
+    }
+
+    bool relock{false};
+    if (!unlockForSigning(passphrase, relock)) {
+        return false;
+    }
+
+    WalletRelockGuard relock_guard{*m_wallet, [this] { notifySecurityStateChanged(); }, relock};
+
+    std::string signature;
+    const SigningResult result{m_wallet->signMessage(message.toStdString(), *pkhash, signature)};
+
+    if (result != SigningResult::OK) {
+        setSigningStatus(QString::fromStdString(SigningResultString(result)));
+        return false;
+    }
+
+    setSignature(QString::fromStdString(signature));
+    return true;
+}
+
+bool SignVerifyMessageModel::verifyMessage(const QString& address, const QString& message, const QString& signature) const
+{
+    if (!LegacyP2PKHFromAddress(address)) {
+        return false;
+    }
+    if (signature.trimmed().isEmpty()) {
+        return false;
+    }
+    return MessageVerify(
+        address.trimmed().toStdString(),
+        signature.trimmed().toStdString(),
+        message.toStdString()) == MessageVerificationResult::OK;
+}
+
+void SignVerifyMessageModel::clear()
+{
+    clearSigningStatus();
+    setSignature(QString());
+}
+
+void SignVerifyMessageModel::clearSigningStatus()
+{
+    setSigningStatus(QString());
+}
+
+bool SignVerifyMessageModel::unlockForSigning(const std::optional<QString>& passphrase, bool& relock)
+{
+    relock = false;
+    if (!m_wallet) {
+        return true;
+    }
+    if (!m_wallet->isCrypted() || !m_wallet->isLocked()) {
+        return true;
+    }
+    if (!passphrase.has_value()) {
+        setSigningStatus(tr("Enter your wallet password to sign this message."), true);
+        return false;
+    }
+
+    SecureString secure_passphrase{QmlUtil::SecureStringFromQString(*passphrase)};
+    switch (TryUnlockWithPassphrase(*m_wallet, secure_passphrase)) {
+    case WalletUnlockResult::IncorrectPassphrase:
+        setSigningStatus(tr("The wallet password you entered was incorrect."));
+        return false;
+    case WalletUnlockResult::AlreadyUnlocked:
+        return true;
+    case WalletUnlockResult::UnlockedNowRelockRequired:
+        relock = true;
+        notifySecurityStateChanged();
+        return true;
+    }
+    return false;
+}
+
+void SignVerifyMessageModel::setSigningStatus(const QString& error, bool needs_unlock)
+{
+    if (m_signing_error != error) {
+        m_signing_error = error;
+        Q_EMIT signingErrorChanged();
+    }
+    if (m_signing_needs_unlock != needs_unlock) {
+        m_signing_needs_unlock = needs_unlock;
+        Q_EMIT signingNeedsUnlockChanged();
+    }
+}
+
+void SignVerifyMessageModel::setSignature(const QString& signature)
+{
+    if (m_signature != signature) {
+        m_signature = signature;
+        Q_EMIT signatureChanged();
+    }
+}
+
+void SignVerifyMessageModel::notifySecurityStateChanged()
+{
+    if (m_security_state_changed) {
+        m_security_state_changed();
+    }
+}
--- a/qml/models/signverifymessagemodel.h
+++ b/qml/models/signverifymessagemodel.h
@@ -1,0 +1,61 @@
+// Copyright (c) 2026 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef BITCOIN_QML_MODELS_SIGNVERIFYMESSAGEMODEL_H
+#define BITCOIN_QML_MODELS_SIGNVERIFYMESSAGEMODEL_H
+
+#include <interfaces/wallet.h>
+
+#include <functional>
+#include <optional>
+
+#include <QObject>
+#include <QString>
+
+class SignVerifyMessageModel : public QObject
+{
+    Q_OBJECT
+    Q_PROPERTY(QString signingError READ signingError NOTIFY signingErrorChanged)
+    Q_PROPERTY(bool signingNeedsUnlock READ signingNeedsUnlock NOTIFY signingNeedsUnlockChanged)
+    Q_PROPERTY(QString signature READ signature NOTIFY signatureChanged)
+
+public:
+    using SecurityStateChangedFn = std::function<void()>;
+
+    explicit SignVerifyMessageModel(interfaces::Wallet* wallet = nullptr, QObject* parent = nullptr);
+
+    QString signingError() const { return m_signing_error; }
+    bool signingNeedsUnlock() const { return m_signing_needs_unlock; }
+    QString signature() const { return m_signature; }
+
+    void setWallet(interfaces::Wallet* wallet);
+    void setSecurityStateChangedFn(SecurityStateChangedFn fn);
+
+    Q_INVOKABLE bool isLegacyP2PKHAddress(const QString& address) const;
+    Q_INVOKABLE bool signMessage(const QString& address, const QString& message);
+    Q_INVOKABLE bool signMessageWithPassphrase(const QString& address, const QString& message, const QString& passphrase);
+    Q_INVOKABLE bool verifyMessage(const QString& address, const QString& message, const QString& signature) const;
+    Q_INVOKABLE void clear();
+    Q_INVOKABLE void clearSigningStatus();
+
+Q_SIGNALS:
+    void signingErrorChanged();
+    void signingNeedsUnlockChanged();
+    void signatureChanged();
+
+private:
+    bool signMessageInternal(const QString& address, const QString& message, const std::optional<QString>& passphrase);
+    bool unlockForSigning(const std::optional<QString>& passphrase, bool& relock);
+    void setSigningStatus(const QString& error, bool needs_unlock = false);
+    void setSignature(const QString& signature);
+    void notifySecurityStateChanged();
+
+    interfaces::Wallet* m_wallet{nullptr};
+    SecurityStateChangedFn m_security_state_changed;
+    QString m_signing_error;
+    bool m_signing_needs_unlock{false};
+    QString m_signature;
+};
+
+#endif // BITCOIN_QML_MODELS_SIGNVERIFYMESSAGEMODEL_H
--- a/qml/models/walletqmlmodel.cpp
+++ b/qml/models/walletqmlmodel.cpp
@@ -8,10 +8,13 @@
 #include <common/messages.h>
 #include <qml/bitcoinamount.h>
 #include <qml/models/activitylistmodel.h>
+#include <qml/models/addresslistmodel.h>
 #include <qml/models/paymentrequest.h>
 #include <qml/models/sendrecipient.h>
 #include <qml/models/sendrecipientslistmodel.h>
+#include <qml/models/walletunlock.h>
 #include <qml/models/walletqmlmodeltransaction.h>
+#include <qml/util.h>
 
 #include <chainparams.h>
 #include <consensus/amount.h>
@@ -30,10 +33,12 @@
 #include <wallet/coincontrol.h>
 #include <wallet/wallet.h>
 
-#include <QByteArray>
 #include <QDateTime>
 #include <QMetaObject>
 #include <QRegularExpression>
+#include <QSettings>
+#include <QVariantList>
+#include <QVariantMap>
 
 #include <array>
 #include <functional>
@@ -257,56 +262,43 @@ QString LocalizedString(const bilingual_str& value)
     return QString::fromStdString(value.translated.empty() ? value.original : value.translated);
 }
 
-class WalletRelockGuard
+QString OutputTypeId(OutputType type)
 {
-public:
-    WalletRelockGuard(interfaces::Wallet& wallet, std::function<void()> refresh_security_state, bool active)
-        : m_wallet{wallet}, m_refresh_security_state{std::move(refresh_security_state)}, m_active{active}
-    {
-    }
-
-    ~WalletRelockGuard()
-    {
-        relock();
-    }
-
-    WalletRelockGuard(const WalletRelockGuard&) = delete;
-    WalletRelockGuard& operator=(const WalletRelockGuard&) = delete;
-
-    void relock()
-    {
-        if (!m_active) {
-            return;
-        }
-        m_wallet.lock();
-        m_refresh_security_state();
-        m_active = false;
-    }
-
-private:
-    interfaces::Wallet& m_wallet;
-    std::function<void()> m_refresh_security_state;
-    bool m_active;
-};
-
-SecureString SecureStringFromQString(const QString& value)
-{
-    QByteArray bytes{value.toUtf8()};
-    SecureString secure;
-    secure.assign(bytes.constData(), bytes.constData() + bytes.size());
-    if (!bytes.isEmpty()) {
-        memory_cleanse(bytes.data(), static_cast<size_t>(bytes.size()));
-    }
-    return secure;
+    return QString::fromStdString(FormatOutputType(type));
 }
 
-void ClearSecureString(SecureString& value)
+QString OutputTypeLabel(OutputType type)
 {
-    if (value.empty()) {
-        return;
+    switch (type) {
+    case OutputType::BECH32M:
+        return QObject::tr("Bech32m (Taproot)");
+    case OutputType::BECH32:
+        return QObject::tr("Bech32 (SegWit)");
+    case OutputType::P2SH_SEGWIT:
+        return QObject::tr("Base58 (P2SH-SegWit)");
+    case OutputType::LEGACY:
+        return QObject::tr("Base58 (Legacy)");
+    case OutputType::UNKNOWN:
+        return {};
     }
-    memory_cleanse(value.data(), value.size());
-    value.clear();
+    return {};
+}
+
+QString OutputTypeDescription(OutputType type)
+{
+    switch (type) {
+    case OutputType::BECH32M:
+        return QObject::tr("Lower fees · Better privacy");
+    case OutputType::BECH32:
+        return QObject::tr("Lower fees · Widely supported");
+    case OutputType::P2SH_SEGWIT:
+        return QObject::tr("Higher fees · Backward compatible");
+    case OutputType::LEGACY:
+        return QObject::tr("Higher fees · Not recommended");
+    case OutputType::UNKNOWN:
+        return {};
+    }
+    return {};
 }
 
 } // namespace
@@ -316,6 +308,7 @@ WalletQmlModel::WalletQmlModel(std::unique_ptr<interfaces::Wallet> wallet, QObje
 {
     m_wallet = std::move(wallet);
     m_activity_list_model = new ActivityListModel(this);
+    m_address_list_model = new AddressListModel(this);
     m_bump_transaction_model = new BumpTransactionModel(m_wallet.get(), this);
     m_coins_list_model = new CoinsListModel(this);
     m_send_recipients = new SendRecipientsListModel(this);
@@ -329,6 +322,7 @@ WalletQmlModel::WalletQmlModel(QObject* parent)
     : QObject(parent)
 {
     m_activity_list_model = new ActivityListModel(this);
+    m_address_list_model = new AddressListModel(this);
     m_bump_transaction_model = new BumpTransactionModel(nullptr, this);
     m_coins_list_model = new CoinsListModel(this);
     m_send_recipients = new SendRecipientsListModel(this);
@@ -348,6 +342,7 @@ WalletQmlModel::~WalletQmlModel()
     }
     delete m_fee_estimation_worker;
     delete m_activity_list_model;
+    delete m_address_list_model;
     delete m_coins_list_model;
     delete m_send_recipients;
     delete m_current_payment_request;
@@ -481,16 +476,6 @@ bool WalletQmlModel::canManagePassphrase() const
     return m_wallet && !m_wallet->privateKeysDisabled();
 }
 
-QString WalletQmlModel::newAddress(QString label)
-{
-    if (!m_wallet) {
-        return QString();
-    }
-    OutputType output_type = m_wallet->getDefaultAddressType();
-    util::Result<CTxDestination> dest{m_wallet->getNewDestination(output_type, label.toStdString())};
-    return QString::fromStdString(EncodeDestination(dest.value()));
-}
-
 bool WalletQmlModel::encryptWallet(const QString& passphrase)
 {
     clearSettingsError();
@@ -503,9 +488,9 @@ bool WalletQmlModel::encryptWallet(const QString& passphrase)
         return false;
     }
 
-    SecureString secure_passphrase{SecureStringFromQString(passphrase)};
+    SecureString secure_passphrase{QmlUtil::SecureStringFromQString(passphrase)};
     const bool encrypted{m_wallet->encryptWallet(secure_passphrase)};
-    ClearSecureString(secure_passphrase);
+    QmlUtil::ClearSecureString(secure_passphrase);
     if (!encrypted) {
         setSettingsError(tr("The wallet password could not be set."));
         return false;
@@ -531,11 +516,11 @@ bool WalletQmlModel::changeWalletPassphrase(const QString& old_passphrase, const
         return false;
     }
 
-    SecureString secure_old_passphrase{SecureStringFromQString(old_passphrase)};
-    SecureString secure_new_passphrase{SecureStringFromQString(new_passphrase)};
+    SecureString secure_old_passphrase{QmlUtil::SecureStringFromQString(old_passphrase)};
+    SecureString secure_new_passphrase{QmlUtil::SecureStringFromQString(new_passphrase)};
     const bool changed{m_wallet->changeWalletPassphrase(secure_old_passphrase, secure_new_passphrase)};
-    ClearSecureString(secure_old_passphrase);
-    ClearSecureString(secure_new_passphrase);
+    QmlUtil::ClearSecureString(secure_old_passphrase);
+    QmlUtil::ClearSecureString(secure_new_passphrase);
     if (!changed) {
         setSettingsError(tr("The current wallet password was incorrect."));
         return false;
@@ -570,6 +555,82 @@ void WalletQmlModel::clearSettingsError()
     setSettingsError(QString());
 }
 
+QVariantList WalletQmlModel::availableReceiveAddressTypes() const
+{
+    if (!m_wallet || !m_wallet->canGetAddresses()) {
+        return {};
+    }
+
+    QVariantList types;
+    const std::array ordered_types{
+        OutputType::BECH32M,
+        OutputType::BECH32,
+        OutputType::P2SH_SEGWIT,
+        OutputType::LEGACY,
+    };
+
+    for (const OutputType type : ordered_types) {
+        if (type == OutputType::BECH32M && (!m_wallet || !m_wallet->taprootEnabled())) {
+            continue;
+        }
+        QVariantMap item;
+        item.insert(QStringLiteral("id"), OutputTypeId(type));
+        item.insert(QStringLiteral("label"), OutputTypeLabel(type));
+        item.insert(QStringLiteral("description"), OutputTypeDescription(type));
+        types.append(item);
+    }
+    return types;
+}
+
+QString WalletQmlModel::defaultReceiveAddressType() const
+{
+    const QVariantList available_types = availableReceiveAddressTypes();
+    QSettings settings;
+    const QString saved_type{settings.value(persistedReceiveAddressTypeKey()).toString()};
+    for (const QVariant& item : available_types) {
+        const QVariantMap type{item.toMap()};
+        if (type.value(QStringLiteral("id")).toString() == saved_type) {
+            return saved_type;
+        }
+    }
+
+    const QString core_default{m_wallet ? OutputTypeId(m_wallet->getDefaultAddressType()) : QString()};
+    for (const QVariant& item : available_types) {
+        const QVariantMap type{item.toMap()};
+        if (type.value(QStringLiteral("id")).toString() == core_default) {
+            return core_default;
+        }
+    }
+
+    return available_types.empty()
+        ? QString{}
+        : available_types.front().toMap().value(QStringLiteral("id")).toString();
+}
+
+void WalletQmlModel::setDefaultReceiveAddressType(const QString& address_type)
+{
+    for (const QVariant& item : availableReceiveAddressTypes()) {
+        const QVariantMap type{item.toMap()};
+        if (type.value(QStringLiteral("id")).toString() == address_type) {
+            QSettings settings;
+            settings.setValue(persistedReceiveAddressTypeKey(), address_type);
+            settings.sync();
+            return;
+        }
+    }
+}
+
+QString WalletQmlModel::receiveAddressTypeLabel(const QString& address_type) const
+{
+    for (const QVariant& item : availableReceiveAddressTypes()) {
+        const QVariantMap type{item.toMap()};
+        if (type.value(QStringLiteral("id")).toString() == address_type) {
+            return type.value(QStringLiteral("label")).toString();
+        }
+    }
+    return {};
+}
+
 void WalletQmlModel::removeWallet()
 {
     if (!m_wallet) {
@@ -578,29 +639,65 @@ void WalletQmlModel::removeWallet()
     m_wallet->remove();
 }
 
-void WalletQmlModel::commitPaymentRequest()
+bool WalletQmlModel::setCurrentPaymentRequestAddress(QString address)
+{
+    if (!m_wallet || !m_current_payment_request || address.isEmpty()) {
+        return false;
+    }
+
+    const CTxDestination destination{DecodeDestination(address.toStdString())};
+    if (!IsValidDestination(destination)) {
+        return false;
+    }
+
+    m_current_payment_request->clear();
+    m_current_payment_request->setDestination(destination);
+    m_current_payment_request->setLabel(getAddressLabel(address));
+    return true;
+}
+
+bool WalletQmlModel::ensurePaymentRequestDestination()
 {
     if (!m_wallet || !m_current_payment_request) {
-        return;
+        return false;
     }
-
-    if (m_current_payment_request->id().isEmpty()) {
-        m_current_payment_request->setId(nextPaymentRequestId());
+    if (!m_current_payment_request->address().isEmpty()) {
+        return true;
     }
-
-    if (m_current_payment_request->address().isEmpty()) {
-        const OutputType output_type = m_wallet->getDefaultAddressType();
-        const auto destination{m_wallet->getNewDestination(output_type, m_current_payment_request->label().toStdString())};
-        if (!destination) {
-            return;
+    const QString address_type_id = m_current_payment_request->addressType().isEmpty()
+        ? defaultReceiveAddressType()
+        : m_current_payment_request->addressType();
+    OutputType output_type = m_wallet->getDefaultAddressType();
+    if (!address_type_id.isEmpty()) {
+        const auto parsed_type{ParseOutputType(address_type_id.toStdString())};
+        if (!parsed_type) {
+            return false;
         }
-        m_current_payment_request->setDestination(destination.value());
+        output_type = *parsed_type;
     }
+    const auto destination{m_wallet->getNewDestination(output_type, m_current_payment_request->label().toStdString())};
+    if (!destination || !IsValidDestination(destination.value())) {
+        return false;
+    }
+    m_current_payment_request->setAddressType(OutputTypeId(output_type));
+    m_current_payment_request->setDestination(destination.value());
+    return true;
+}
+
+bool WalletQmlModel::saveCurrentPaymentRequest()
+{
+    if (!m_wallet || !m_current_payment_request) {
+        return false;
+    }
+
+    const QString request_id_text = m_current_payment_request->id().isEmpty()
+        ? QString::number(nextPaymentRequestId())
+        : m_current_payment_request->id();
 
     bool parse_ok{false};
-    const int64_t request_id{m_current_payment_request->id().toLongLong(&parse_ok)};
+    const int64_t request_id{request_id_text.toLongLong(&parse_ok)};
     if (!parse_ok || request_id <= 0) {
-        return;
+        return false;
     }
 
     QmlRecentRequestEntry request_entry;
@@ -614,10 +711,68 @@ void WalletQmlModel::commitPaymentRequest()
     DataStream ss{};
     ss << request_entry;
 
-    m_wallet->setAddressReceiveRequest(
+    const bool saved{m_wallet->setAddressReceiveRequest(
         m_current_payment_request->destination(),
-        m_current_payment_request->id().toStdString(),
-        ss.str());
+        request_id_text.toStdString(),
+        ss.str())};
+    if (!saved) {
+        return false;
+    }
+
+    if (m_current_payment_request->id().isEmpty()) {
+        m_current_payment_request->setId(static_cast<unsigned int>(request_id));
+    }
+    return true;
+}
+
+bool WalletQmlModel::commitPaymentRequest()
+{
+    if (!m_wallet || !m_current_payment_request) {
+        return false;
+    }
+
+    if (!ensurePaymentRequestDestination()) {
+        if (m_wallet->isCrypted() && m_wallet->isLocked()) {
+            m_current_payment_request->setNeedsUnlock(true);
+        }
+        return false;
+    }
+    return saveCurrentPaymentRequest();
+}
+
+bool WalletQmlModel::commitPaymentRequestWithPassphrase(const QString& passphrase)
+{
+    if (!m_wallet || !m_current_payment_request) {
+        return false;
+    }
+
+    SecureString secure_passphrase{QmlUtil::SecureStringFromQString(passphrase)};
+    const auto result{TryUnlockWithPassphrase(*m_wallet, secure_passphrase)};
+    switch (result) {
+    case WalletUnlockResult::IncorrectPassphrase:
+        m_current_payment_request->setUnlockError(tr("The wallet password you entered was incorrect."));
+        return false;
+    case WalletUnlockResult::AlreadyUnlocked:
+    case WalletUnlockResult::UnlockedNowRelockRequired:
+        break;
+    }
+
+    const bool need_relock = result == WalletUnlockResult::UnlockedNowRelockRequired;
+    if (need_relock) {
+        refreshSecurityState();
+    }
+    WalletRelockGuard relock_guard{*m_wallet, [this] { refreshSecurityState(); }, need_relock};
+
+    if (!ensurePaymentRequestDestination()) {
+        return false;
+    }
+    if (!saveCurrentPaymentRequest()) {
+        return false;
+    }
+
+    m_current_payment_request->setNeedsUnlock(false);
+    m_current_payment_request->setUnlockError(QString());
+    return true;
 }
 
 unsigned int WalletQmlModel::nextPaymentRequestId() const
@@ -692,6 +847,136 @@ QString WalletQmlModel::getAddressLabel(const QString& address) const
     }
 
     return QString::fromStdString(label);
+}
+
+bool WalletQmlModel::setAddressLabel(const QString& address, const QString& label)
+{
+    if (!m_wallet || address.isEmpty()) {
+        return false;
+    }
+
+    const CTxDestination destination{DecodeDestination(address.toStdString())};
+    if (!IsValidDestination(destination)) {
+        return false;
+    }
+
+    wallet::AddressPurpose purpose{wallet::AddressPurpose::RECEIVE};
+    if (!m_wallet->getAddress(destination, nullptr, nullptr, &purpose)) {
+        return false;
+    }
+
+    return m_wallet->setAddressBook(destination, label.toStdString(), purpose);
+}
+
+std::vector<interfaces::WalletAddress> WalletQmlModel::getAddresses() const
+{
+    if (!m_wallet) {
+        return {};
+    }
+    return m_wallet->getAddresses();
+}
+
+std::map<QString, CAmount> WalletQmlModel::addressBalances() const
+{
+    std::map<QString, CAmount> balances;
+    if (!m_wallet) {
+        return balances;
+    }
+
+    for (const auto& coins_entry : m_wallet->listCoins()) {
+        for (const auto& [outpoint, tx_out] : coins_entry.second) {
+            CTxDestination destination;
+            if (!ExtractDestination(tx_out.txout.scriptPubKey, destination))
+                continue;
+
+            const QString address{QString::fromStdString(EncodeDestination(destination))};
+            if (address.isEmpty())
+                continue;
+
+            balances[address] += tx_out.txout.nValue;
+        }
+    }
+
+    return balances;
+}
+
+std::set<QString> WalletQmlModel::usedAddresses() const
+{
+    std::set<QString> addresses;
+
+    if (!m_wallet)
+        return addresses;
+
+    std::set<QString> receive_addresses;
+    for (const interfaces::WalletAddress& wallet_address : getAddresses()) {
+        if (wallet_address.purpose != wallet::AddressPurpose::RECEIVE || wallet_address.is_mine == wallet::ISMINE_NO) {
+            continue;
+        }
+
+        const QString address{QString::fromStdString(EncodeDestination(wallet_address.dest))};
+        if (!address.isEmpty()) {
+            receive_addresses.insert(address);
+        }
+    }
+
+    for (const interfaces::WalletTx& wallet_tx : m_wallet->getWalletTxs()) {
+        for (size_t i{0}; i < wallet_tx.txout_address.size(); ++i) {
+            if (i >= wallet_tx.txout_address_is_mine.size() || !wallet_tx.txout_address_is_mine[i])
+                continue;
+
+            if (i < wallet_tx.txout_is_change.size() && wallet_tx.txout_is_change[i])
+                continue;
+
+            const QString address{QString::fromStdString(EncodeDestination(wallet_tx.txout_address[i]))};
+            if (!address.isEmpty() && receive_addresses.count(address) > 0)
+                addresses.insert(address);
+        }
+    }
+
+    return addresses;
+}
+
+std::set<QString> WalletQmlModel::changeAddresses() const
+{
+    std::set<QString> addresses;
+    if (!m_wallet) {
+        return addresses;
+    }
+
+    std::set<COutPoint> change_outpoints;
+    for (const interfaces::WalletTx& wallet_tx : m_wallet->getWalletTxs()) {
+        if (!wallet_tx.tx)
+            continue;
+
+        const Txid txid{wallet_tx.tx->GetHash()};
+        for (size_t i{0}; i < wallet_tx.txout_is_change.size(); ++i) {
+            if (!wallet_tx.txout_is_change[i])
+                continue;
+
+            change_outpoints.insert(COutPoint{txid, static_cast<uint32_t>(i)});
+        }
+    }
+
+    for (const auto& coins_entry : m_wallet->listCoins()) {
+        for (const auto& [outpoint, tx_out] : coins_entry.second) {
+            if (tx_out.txout.nValue <= 0)
+                continue;
+
+            if (change_outpoints.count(outpoint) > 0) {
+                CTxDestination destination;
+                if (!ExtractDestination(tx_out.txout.scriptPubKey, destination))
+                    continue;
+
+                const QString address{QString::fromStdString(EncodeDestination(destination))};
+                if (address.isEmpty())
+                    continue;
+
+                addresses.insert(address);
+            }
+        }
+    }
+
+    return addresses;
 }
 
 std::unique_ptr<interfaces::Handler> WalletQmlModel::handleTransactionChanged(TransactionChangedFn fn)
@@ -872,7 +1157,7 @@ bool WalletQmlModel::prepareTransaction()
 
 bool WalletQmlModel::prepareTransactionWithPassphrase(const QString& passphrase)
 {
-    return prepareTransactionInternal(std::optional<SecureString>{SecureStringFromQString(passphrase)});
+    return prepareTransactionInternal(std::optional<SecureString>{QmlUtil::SecureStringFromQString(passphrase)});
 }
 
 bool WalletQmlModel::prepareTransactionInternal(std::optional<SecureString> passphrase)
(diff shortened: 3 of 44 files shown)


johnny9 (commented):
Can we try to reuse the toggle buttons in the NetworkTraffic page?

pseudoramdom:
> Can we try to reuse the toggle buttons in the NetworkTraffic page?

Good callout. 
I’ll update `SegmentedPicker` to use `ToggleButton` internally, so it gets the same visual interaction style as NetworkTraffic. I'd still keep the `SegmentedPicker` since it's a higher-level grouped control: it takes a model, owns currentIndex, emits selected(index, option), and avoids repeating the RowLayout + Repeater + checked/index wiring anywhere we need a segmented choice.

pseudoramdom:
Rebased on top of #546 (e5fab687ec07912e8d615e3ad9c4e73129672ac4)

pseudoramdom:
Rebased on top of `qt6` now 1ce12be74c15cba857a4e9a8e0e53b225972157d

johnny9:
image
the layout for the segmented pickers renders very strange for me. This seems to be another case of mac rendering different than linux which is very frustrating.

johnny9:
image


The following resolved the weird layout issue I was seeing. It looks good now. Double check these changes on your side.

SegmentedPicker follow-up from local testing:

- The unequal segment widths came from `RowLayout` preserving each `ToggleButton` implicit/preferred width. `Single-use` was naturally wider than `Change`, so `Layout.fillWidth: true` alone did not make the segments equal. Setting `Layout.preferredWidth: 1` and `Layout.minimumWidth: 0` gives each repeated button the same layout basis.
- The selected button looked offset because the picker used `anchors.fill` plus `anchors.margins` on the `contentItem`. The automation bridge showed the content row was inset to `35px` tall at `x/y = 5`, while each toggle button background was still `45px` tall. Moving the inset to `Control.padding: 5` makes the content row and button backgrounds the same height.

Patch:

```diff
diff --git a/qml/controls/SegmentedPicker.qml b/qml/controls/SegmentedPicker.qml
index 21d700d839..a74b9fba88 100644
--- a/qml/controls/SegmentedPicker.qml
+++ b/qml/controls/SegmentedPicker.qml
@@ -24,6 +24,7 @@ Control {
 
     implicitHeight: 45
     implicitWidth: 360
+    padding: 5
 
     background: Rectangle {
         color: Theme.color.neutral3
@@ -35,8 +36,6 @@ Control {
     }
 
     contentItem: RowLayout {
-        anchors.fill: parent
-        anchors.margins: 5
         spacing: 5
 
         Repeater {
@@ -48,6 +47,8 @@ Control {
 
                 Layout.fillWidth: true
                 Layout.fillHeight: true
+                Layout.preferredWidth: 1
+                Layout.minimumWidth: 0
                 autoExclusive: true
                 checked: index === root.currentIndex
                 text: root.optionText(modelData)
```

Local verification:

- `cmake --build build --target bitcoin-core-app -j2`
- Automation bridge flow reached the Addresses page and confirmed the picker rendered with equal-width, properly inset segments.
- `git diff --check -- qml/controls/SegmentedPicker.qml`

test/functional/qml_test_addresses.py
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env python3
```
johnny9, line 1: This test should be added to the github workflow for gui testing

qml/pages/wallet/RequestPayment.qml
johnny9: We need to fix this logic now. This should check request.id, not request.address; an address-book request already has an address but is still unpersisted.

Something like...
```
diff --git a/qml/pages/wallet/RequestPayment.qml b/qml/pages/wallet/RequestPayment.qml
index 94902dd01d..0000000000 100644
--- a/qml/pages/wallet/RequestPayment.qml
+++ b/qml/pages/wallet/RequestPayment.qml
@@ -18,6 +18,29 @@ Page {
 
     property WalletQmlModel wallet: walletController.selectedWallet
     property PaymentRequest request: wallet ? wallet.currentPaymentRequest : null
+    readonly property bool requestReady: root.request !== null
+    readonly property bool requestPersisted: root.requestReady && root.request.id !== ""
+    readonly property bool requestHasAddress: root.requestReady && root.request.address !== ""
+
+    function primaryActionText() {
+        if (root.requestPersisted) {
+            return qsTr("Copy payment request")
+        }
+        if (root.requestHasAddress) {
+            return qsTr("Create payment request")
+        }
+        return qsTr("Create bitcoin address")
+    }
+
+    function runPrimaryAction() {
+        if (!root.requestReady) {
+            return
+        }
+        if (!root.requestPersisted) {
+            root.wallet.commitPaymentRequest()
+            return
+        }
+        Clipboard.setText(root.request.address)
+    }
 
     ScrollView {
         clip: true
@@ -270,19 +293,8 @@ Page {
                     objectName: "requestPaymentCreateButton"
                     Layout.fillWidth: true
                     Layout.topMargin: 30
-                    text: root.request !== null && root.request.id !== ""
-                        ? qsTr("Copy payment request")
-                        : qsTr("Create bitcoin address")
-                    onClicked: {
-                        if (!root.request) {
-                            return
-                        }
-                        if (root.request.address === "") {
-                            root.wallet.commitPaymentRequest()
-                        } else {
-                            Clipboard.setText(root.request.address)
-                        }
-                    }
+                    text: root.primaryActionText()
+                    onClicked: root.runPrimaryAction()
                 }
```

qml/models/walletqmlmodel.cpp
johnny9: `getNewDestination()` can fail, but this QML path unwraps it with `dest.value()`. If address generation fails, for example an encrypted locked wallet needs the keypool topped up, this can assert/crash instead of prompting for the passphrase. The QML caller also closes the “New Address” popup unconditionally, so there is no visible failure state.

Here's a big patch based on this branch rebased onto current qt6 branch.

This changes address creation from a bare QString return into a structured result flow so getNewDestination() failure is handled instead of unwrapping dest.value(). The UI now keeps the “New Address” popup open on failure, shows an error, and uses the existing WalletPassphrasePopup when an encrypted locked wallet needs to unlock/top up the keypool. After a passphrase is submitted, the wallet is unlocked temporarily, address generation is retried, and the wallet is relocked.

The patch also refreshes AddressListModel only after successful address creation and adds unit coverage for:
- plain getNewDestination() failure
- encrypted locked wallet requiring a passphrase
- unlock/retry/relock success
- relock after retry failure

```
diff --git a/qml/models/addresslistmodel.cpp b/qml/models/addresslistmodel.cpp
index 6c9a7c7100..f1492115b8 100644
--- a/qml/models/addresslistmodel.cpp
+++ b/qml/models/addresslistmodel.cpp
@@ -160,12 +160,24 @@ void AddressListModel::refresh()
     rebuild();
 }
 
-QString AddressListModel::createSingleUseAddress(const QString& label)
+QVariantMap AddressListModel::createSingleUseAddress(const QString& label)
 {
     if (!m_wallet_model) return {};
-    const QString address{m_wallet_model->newAddress(label)};
-    refresh();
-    return address;
+    const QVariantMap result{m_wallet_model->createReceiveAddress(label)};
+    if (result.value(QStringLiteral("ok")).toBool()) {
+        refresh();
+    }
+    return result;
+}
+
+QVariantMap AddressListModel::createSingleUseAddressWithPassphrase(const QString& label, const QString& passphrase)
+{
+    if (!m_wallet_model) return {};
+    const QVariantMap result{m_wallet_model->createReceiveAddressWithPassphrase(label, passphrase)};
+    if (result.value(QStringLiteral("ok")).toBool()) {
+        refresh();
+    }
+    return result;
 }
 
 bool AddressListModel::setAddressLabel(const QString& address, const QString& label)
diff --git a/qml/models/addresslistmodel.h b/qml/models/addresslistmodel.h
index 718b0aa131..73aa5a1640 100644
--- a/qml/models/addresslistmodel.h
+++ b/qml/models/addresslistmodel.h
@@ -10,6 +10,7 @@
 #include <QAbstractListModel>
 #include <QString>
 #include <QVariantList>
+#include <QVariantMap>
 
 #include <vector>
 
@@ -61,7 +62,8 @@ public:
     void setShowUsed(bool show_used);
 
     Q_INVOKABLE void refresh();
-    Q_INVOKABLE QString createSingleUseAddress(const QString& label);
+    Q_INVOKABLE QVariantMap createSingleUseAddress(const QString& label);
+    Q_INVOKABLE QVariantMap createSingleUseAddressWithPassphrase(const QString& label, const QString& passphrase);
     Q_INVOKABLE bool setAddressLabel(const QString& address, const QString& label);
     Q_INVOKABLE QString addressAt(int row) const;
 
diff --git a/qml/models/walletqmlmodel.cpp b/qml/models/walletqmlmodel.cpp
index 79b80120ca..438497f4e8 100644
--- a/qml/models/walletqmlmodel.cpp
+++ b/qml/models/walletqmlmodel.cpp
@@ -35,6 +35,7 @@
 #include <QDateTime>
 #include <QMetaObject>
 #include <QRegularExpression>
+#include <QVariantMap>
 
 #include <array>
 #include <functional>
@@ -310,6 +311,35 @@ void ClearSecureString(SecureString& value)
     value.clear();
 }
 
+QVariantMap AddressCreationResult(bool ok, const QString& address, const QString& error, bool needs_unlock)
+{
+    return {
+        {QStringLiteral("ok"), ok},
+        {QStringLiteral("address"), address},
+        {QStringLiteral("error"), error},
+        {QStringLiteral("needsUnlock"), needs_unlock},
+    };
+}
+
+QVariantMap AddressCreationSuccess(const CTxDestination& destination)
+{
+    return AddressCreationResult(
+        /*ok=*/true,
+        QString::fromStdString(EncodeDestination(destination)),
+        QString(),
+        /*needs_unlock=*/false);
+}
+
+QVariantMap AddressCreationError(const QString& error)
+{
+    return AddressCreationResult(/*ok=*/false, QString(), error, /*needs_unlock=*/false);
+}
+
+QVariantMap AddressCreationNeedsUnlock(const QString& error)
+{
+    return AddressCreationResult(/*ok=*/false, QString(), error, /*needs_unlock=*/true);
+}
+
 } // namespace
 
 WalletQmlModel::WalletQmlModel(std::unique_ptr<interfaces::Wallet> wallet, QObject *parent)
@@ -437,14 +467,62 @@ QString WalletQmlModel::name() const
     return QString::fromStdString(m_wallet->getWalletName());
 }
 
-QString WalletQmlModel::newAddress(QString label)
+QVariantMap WalletQmlModel::createReceiveAddress(const QString& label)
+{
+    return createReceiveAddressInternal(label, std::nullopt);
+}
+
+QVariantMap WalletQmlModel::createReceiveAddressWithPassphrase(const QString& label, const QString& passphrase)
+{
+    return createReceiveAddressInternal(label, std::optional<SecureString>{SecureStringFromQString(passphrase)});
+}
+
+QVariantMap WalletQmlModel::createReceiveAddressInternal(const QString& label, std::optional<SecureString> passphrase)
 {
     if (!m_wallet) {
-        return QString();
+        if (passphrase.has_value()) {
+            ClearSecureString(*passphrase);
+        }
+        return AddressCreationError(tr("Wallet is not loaded."));
+    }
+
+    const auto get_new_destination = [&]() {
+        return m_wallet->getNewDestination(m_wallet->getDefaultAddressType(), label.toStdString());
+    };
+
+    const bool needs_unlock = m_wallet->isCrypted() && m_wallet->isLocked() && !m_wallet->privateKeysDisabled();
+    if (!passphrase.has_value()) {
+        const util::Result<CTxDestination> destination{get_new_destination()};
+        if (destination) {
+            return AddressCreationSuccess(*destination);
+        }
+        if (needs_unlock) {
+            return AddressCreationNeedsUnlock(tr("Enter your wallet password to create a new address."));
+        }
+        return AddressCreationError(LocalizedString(util::ErrorString(destination)));
+    }
+
+    bool relock{false};
+    if (needs_unlock) {
+        const bool unlocked = m_wallet->unlock(*passphrase);
+        ClearSecureString(*passphrase);
+        passphrase.reset();
+        if (!unlocked) {
+            return AddressCreationError(tr("The wallet password you entered was incorrect."));
+        }
+        relock = true;
+        refreshSecurityState();
+    } else {
+        ClearSecureString(*passphrase);
+        passphrase.reset();
+    }
+
+    WalletRelockGuard relock_guard{*m_wallet, [this] { refreshSecurityState(); }, relock};
+    const util::Result<CTxDestination> destination{get_new_destination()};
+    if (!destination) {
+        return AddressCreationError(LocalizedString(util::ErrorString(destination)));
     }
-    OutputType output_type = m_wallet->getDefaultAddressType();
-    util::Result<CTxDestination> dest{m_wallet->getNewDestination(output_type, label.toStdString())};
-    return QString::fromStdString(EncodeDestination(dest.value()));
+    return AddressCreationSuccess(*destination);
 }
 
 bool WalletQmlModel::setCurrentPaymentRequestAddress(QString address)
diff --git a/qml/models/walletqmlmodel.h b/qml/models/walletqmlmodel.h
index fecff08972..4e78bea03c 100644
--- a/qml/models/walletqmlmodel.h
+++ b/qml/models/walletqmlmodel.h
@@ -29,6 +29,7 @@
 #include <QObject>
 #include <QThread>
 #include <QTimer>
+#include <QVariantMap>
 
 class WalletQmlModel : public QObject
 {
@@ -86,7 +87,8 @@ public:
     Q_INVOKABLE void approveExternalSignerTransaction();
     Q_INVOKABLE bool prepareTransactionWithPassphrase(const QString& passphrase);
     Q_INVOKABLE bool sendTransaction();
-    Q_INVOKABLE QString newAddress(QString label);
+    Q_INVOKABLE QVariantMap createReceiveAddress(const QString& label);
+    Q_INVOKABLE QVariantMap createReceiveAddressWithPassphrase(const QString& label, const QString& passphrase);
     Q_INVOKABLE QString estimatedFeeForTarget(unsigned int target_blocks) const;
     Q_INVOKABLE int feeTargetIndex(unsigned int target_blocks) const;
     Q_INVOKABLE void scheduleFeeEstimates();
@@ -166,6 +168,7 @@ private:
     void subscribeToWalletSignals();
     void unsubscribeFromWalletSignals();
     void refreshSecurityState();
+    QVariantMap createReceiveAddressInternal(const QString& label, std::optional<SecureString> passphrase);
     bool prepareTransactionInternal(std::optional<SecureString> passphrase);
     bool sendTransactionInternal();
     bool unlockForAction(std::optional<SecureString>& passphrase, bool& relock);
diff --git a/qml/pages/wallet/AddressList.qml b/qml/pages/wallet/AddressList.qml
index 0f009425e2..535f55bd55 100644
--- a/qml/pages/wallet/AddressList.qml
+++ b/qml/pages/wallet/AddressList.qml
@@ -24,6 +24,7 @@ Page {
     property string selectedCategory: ""
     property string selectedScriptType: ""
     property bool selectedUsed: false
+    property string newAddressErrorText: ""
 
     signal back
     signal receiveRequested
@@ -35,6 +36,43 @@ Page {
         menu.open();
     }
 
+    function newAddressFallbackError(result) {
+        if (result && result.error && result.error.length > 0) {
+            return result.error;
+        }
+        return qsTr("New address generation failed.");
+    }
+
+    function handleNewAddressResult(result) {
+        if (result && result.ok) {
+            root.newAddressErrorText = "";
+            newAddressPassphrasePopup.errorText = "";
+            newAddressPassphrasePopup.close();
+            addressModel.category = AddressListModel.SingleUse;
+            newAddressPopup.close();
+            return true;
+        }
+
+        if (result && result.needsUnlock) {
+            root.newAddressErrorText = "";
+            newAddressPassphrasePopup.errorText = "";
+            newAddressPassphrasePopup.open();
+            return false;
+        }
+
+        const error = root.newAddressFallbackError(result);
+        if (newAddressPassphrasePopup.opened) {
+            newAddressPassphrasePopup.errorText = error;
+        } else {
+            root.newAddressErrorText = error;
+        }
+        return false;
+    }
+
+    function createNewAddress() {
+        return root.handleNewAddressResult(addressModel.createSingleUseAddress(newAddressLabel.text));
+    }
+
     header: NavigationBar2 {
         leftItem: NavButton {
             text: qsTr("Back")
@@ -189,6 +227,10 @@ Page {
             border.color: Theme.color.neutral4
             radius: 10
         }
+        onClosed: {
+            root.newAddressErrorText = "";
+            newAddressPassphrasePopup.close();
+        }
         contentItem: ColumnLayout {
             spacing: 16
 
@@ -204,19 +246,45 @@ Page {
                 Layout.fillWidth: true
                 placeholderText: qsTr("Add note...")
             }
+            CoreText {
+                objectName: "newAddressErrorText"
+                visible: text.length > 0
+                text: root.newAddressErrorText
+                color: Theme.color.red
+                font.pixelSize: 15
+                wrapMode: Text.WordWrap
+            }
             ContinueButton {
                 objectName: "newAddressCreateButton"
                 Layout.fillWidth: true
                 text: qsTr("Create")
-                onClicked: {
-                    addressModel.createSingleUseAddress(newAddressLabel.text);
-                    addressModel.category = AddressListModel.SingleUse;
-                    newAddressPopup.close();
-                }
+                onClicked: root.createNewAddress()
             }
         }
     }
 
+    WalletPassphrasePopup {
+        id: newAddressPassphrasePopup
+        parent: Overlay.overlay
+        width: Math.min(420, root.width - 40)
+        popupObjectName: "newAddressPassphrasePopup"
+        passphraseFieldObjectName: "newAddressPassphraseField"
+        errorTextObjectName: "newAddressPassphraseErrorText"
+        cancelButtonObjectName: "newAddressPassphraseCancelButton"
+        confirmButtonObjectName: "newAddressPassphraseConfirmButton"
+        titleText: qsTr("Enter wallet password")
+        descriptionText: qsTr("Enter your wallet password to create a new address.")
+        confirmText: qsTr("Unlock and create")
+        busyConfirmText: qsTr("Unlocking...")
+        onSubmitted: (passphrase) => {
+            newAddressPassphrasePopup.busy = true;
+            const result = addressModel.createSingleUseAddressWithPassphrase(newAddressLabel.text, passphrase);
+            newAddressPassphrasePopup.busy = false;
+            root.handleNewAddressResult(result);
+        }
+    }
+
     Popup {
         id: detailsPopup
         anchors.centerIn: Overlay.overlay
@@ -286,6 +354,8 @@ Page {
                         textSize: 15
                         padding: 10
                         onClicked: {
+                            root.newAddressErrorText = "";
+                            newAddressPassphrasePopup.errorText = "";
                             newAddressLabel.text = "";
                             newAddressPopup.open();
                         }
diff --git a/test/mocks/mockwallet.h b/test/mocks/mockwallet.h
index 55644d528c..4a681ec9c6 100644
--- a/test/mocks/mockwallet.h
+++ b/test/mocks/mockwallet.h
@@ -100,9 +100,13 @@ class MockWallet : public StubWallet
 {
 public:
     std::function<util::Result<CTransactionRef>(const std::vector<wallet::CRecipient>&, const wallet::CCoinControl&, bool, int&, CAmount&)> createTransactionHandler;
+    std::function<util::Result<CTxDestination>(OutputType, const std::string&)> getNewDestinationHandler;
 
     util::Result<CTxDestination> getNewDestination(const OutputType type, const std::string& label) override
     {
+        if (getNewDestinationHandler) {
+            return getNewDestinationHandler(type, label);
+        }
         return getNewDestinationValue(type, label);
     }
 
diff --git a/test/test_addresslistmodel.cpp b/test/test_addresslistmodel.cpp
index 97c6e04776..1e558e8c01 100644
--- a/test/test_addresslistmodel.cpp
+++ b/test/test_addresslistmodel.cpp
@@ -6,6 +6,8 @@
 
 #include <test/mocks/mockwallet.h>
 
+#include <QVariantMap>
+
 #include <addresstype.h>
 #include <chainparams.h>
 #include <key_io.h>
@@ -151,7 +153,9 @@ void AddressListModelTests::labelsCanBeCreatedAndEdited()
     WalletQmlModel wallet_model{std::move(wallet)};
     AddressListModel* model{wallet_model.addressListModel()};
 
-    const QString address{model->createSingleUseAddress(QStringLiteral("first label"))};
+    const QVariantMap result{model->createSingleUseAddress(QStringLiteral("first label"))};
+    QVERIFY(result.value(QStringLiteral("ok")).toBool());
+    const QString address{result.value(QStringLiteral("address")).toString()};
     QCOMPARE(address, QString::fromStdString(EncodeDestination(created)));
     QCOMPARE(model->rowCount(), 1);
     QCOMPARE(model->data(model->index(0), AddressListModel::LabelRole).toString(), QStringLiteral("first label"));
diff --git a/test/test_walletqmlmodel.cpp b/test/test_walletqmlmodel.cpp
index 22c4b1850b..5c72785cdb 100644
--- a/test/test_walletqmlmodel.cpp
+++ b/test/test_walletqmlmodel.cpp
@@ -18,6 +18,7 @@
 
 #include <QSignalSpy>
 #include <QSemaphore>
+#include <QVariantMap>
 
 #include <algorithm>
 #include <atomic>
@@ -96,7 +97,9 @@ public:
     int unlock_calls{0};
     int lock_calls{0};
     int commit_calls{0};
+    int get_new_destination_calls{0};
     std::vector<std::string> unlock_passphrases;
+    std::vector<std::string> new_destination_labels;
     std::vector<bool> create_transaction_sign_args;
     std::vector<bool> fill_psbt_sign_args;
 
@@ -120,6 +123,11 @@ public:
         locked = false;
         return true;
     };
+    std::function<util::Result<CTxDestination>(OutputType, const std::string&)> get_new_destination_fn = [this](OutputType, const std::string& label) -> util::Result<CTxDestination> {
+        ++get_new_destination_calls;
+        new_destination_labels.push_back(label);
+        return DecodeDestination(VALID_MAINNET_ADDRESS.toStdString());
+    };
     std::function<std::optional<common::PSBTError>(std::optional<int>,
                                                    bool,
                                                    bool,
@@ -146,6 +154,10 @@ public:
     }
     bool unlock(const SecureString& wallet_passphrase) override { return unlock_fn(wallet_passphrase); }
     bool isLocked() override { return locked; }
+    util::Result<CTxDestination> getNewDestination(const OutputType type, const std::string& label) override
+    {
+        return get_new_destination_fn(type, label);
+    }
     util::Result<CTransactionRef> createTransaction(const std::vector<wallet::CRecipient>& recipients,
                                                     const wallet::CCoinControl& coin_control,
                                                     bool sign,
@@ -211,6 +223,10 @@ private Q_SLOTS:
     void scheduleFeeEstimates_usesSelectedCoinsInCoinControl();
     void scheduleFeeEstimates_debouncesRapidRestarts();
     void transactionChangedEmitsBalanceChanged();
+    void createReceiveAddressReturnsErrorWhenGenerationFails();
+    void createReceiveAddressOnLockedWalletRequiresPassword();
+    void createReceiveAddressWithPassphraseUnlocksRetriesAndRelocks();
+    void createReceiveAddressWithPassphraseRelocksWhenGenerationFails();
     void prepareTransactionOnLockedWalletRequiresPassword();
     void prepareTransactionWithPrivateKeysDisabledDoesNotRequirePassword();
     void prepareTransactionWithPassphraseForwardsUtf8Bytes();
@@ -690,6 +706,100 @@ void WalletQmlModelTests::transactionChangedEmitsBalanceChanged()
     QCOMPARE(model.balance(), QStringLiteral("75.00000000"));
 }
 
+void WalletQmlModelTests::createReceiveAddressReturnsErrorWhenGenerationFails()
+{
+    NiceMock<MockWallet>* wallet{nullptr};
+    auto model = MakeWalletModel(wallet);
+    int get_new_destination_calls{0};
+    OutputType requested_type{OutputType::UNKNOWN};
+    std::string requested_label;
+    wallet->getNewDestinationHandler = [&](OutputType type, const std::string& label) -> util::Result<CTxDestination> {
+        ++get_new_destination_calls;
+        requested_type = type;
+        requested_label = label;
+        return util::Error{Untranslated("keypool empty")};
+    };
+
+    const QVariantMap result{model->createReceiveAddress(QStringLiteral("new label"))};
+    QVERIFY(!result.value(QStringLiteral("ok")).toBool());
+    QVERIFY(!result.value(QStringLiteral("needsUnlock")).toBool());
+    QCOMPARE(result.value(QStringLiteral("address")).toString(), QString());
+    QCOMPARE(result.value(QStringLiteral("error")).toString(), QStringLiteral("keypool empty"));
+    QCOMPARE(get_new_destination_calls, 1);
+    QCOMPARE(requested_type, OutputType::BECH32);
+    QCOMPARE(requested_label, std::string{"new label"});
+}
+
+void WalletQmlModelTests::createReceiveAddressOnLockedWalletRequiresPassword()
+{
+    FakePasswordWallet* wallet{nullptr};
+    auto model = MakeWalletModel(wallet);
+    wallet->get_new_destination_fn = [wallet](OutputType, const std::string& label) -> util::Result<CTxDestination> {
+        ++wallet->get_new_destination_calls;
+        wallet->new_destination_labels.push_back(label);
+        return util::Error{Untranslated("keypool empty")};
+    };
+
+    const QVariantMap result{model->createReceiveAddress(QStringLiteral("locked label"))};
+    QVERIFY(!result.value(QStringLiteral("ok")).toBool());
+    QVERIFY(result.value(QStringLiteral("needsUnlock")).toBool());
+    QCOMPARE(result.value(QStringLiteral("error")).toString(), QStringLiteral("Enter your wallet password to create a new address."));
+    QCOMPARE(wallet->get_new_destination_calls, 1);
+    QCOMPARE(wallet->unlock_calls, 0);
+    QCOMPARE(wallet->lock_calls, 0);
+    QVERIFY(wallet->locked);
+}
+
+void WalletQmlModelTests::createReceiveAddressWithPassphraseUnlocksRetriesAndRelocks()
+{
+    FakePasswordWallet* wallet{nullptr};
+    auto model = MakeWalletModel(wallet);
+    wallet->get_new_destination_fn = [wallet](OutputType, const std::string& label) -> util::Result<CTxDestination> {
+        ++wallet->get_new_destination_calls;
+        wallet->new_destination_labels.push_back(label);
+        if (wallet->locked) {
+            return util::Error{Untranslated("keypool empty")};
+        }
+        return DecodeDestination(VALID_MAINNET_ADDRESS.toStdString());
+    };
+
+    const QString passphrase{QString::fromUtf8("pässwörd-₿")};
+    const QVariantMap result{model->createReceiveAddressWithPassphrase(QStringLiteral("receive label"), passphrase)};
+
+    QVERIFY(result.value(QStringLiteral("ok")).toBool());
+    QVERIFY(!result.value(QStringLiteral("needsUnlock")).toBool());
+    QCOMPARE(result.value(QStringLiteral("address")).toString(), VALID_MAINNET_ADDRESS);
+    QCOMPARE(wallet->get_new_destination_calls, 1);
+    QCOMPARE(wallet->new_destination_labels.size(), size_t{1});
+    QCOMPARE(wallet->new_destination_labels.front(), std::string{"receive label"});
+    QCOMPARE(wallet->unlock_calls, 1);
+    QCOMPARE(wallet->unlock_passphrases.size(), size_t{1});
+    QCOMPARE(wallet->unlock_passphrases.front(), passphrase.toUtf8().toStdString());
+    QCOMPARE(wallet->lock_calls, 1);
+    QVERIFY(wallet->locked);
+}
+
+void WalletQmlModelTests::createReceiveAddressWithPassphraseRelocksWhenGenerationFails()
+{
+    FakePasswordWallet* wallet{nullptr};
+    auto model = MakeWalletModel(wallet);
+    wallet->get_new_destination_fn = [wallet](OutputType, const std::string& label) -> util::Result<CTxDestination> {
+        ++wallet->get_new_destination_calls;
+        wallet->new_destination_labels.push_back(label);
+        return util::Error{Untranslated("keypool still empty")};
+    };
+
+    const QVariantMap result{model->createReceiveAddressWithPassphrase(QStringLiteral("receive label"), QStringLiteral("secret"))};
+
+    QVERIFY(!result.value(QStringLiteral("ok")).toBool());
+    QVERIFY(!result.value(QStringLiteral("needsUnlock")).toBool());
+    QCOMPARE(result.value(QStringLiteral("error")).toString(), QStringLiteral("keypool still empty"));
+    QCOMPARE(wallet->get_new_destination_calls, 1);
+    QCOMPARE(wallet->unlock_calls, 1);
+    QCOMPARE(wallet->lock_calls, 1);
+    QVERIFY(wallet->locked);
+}
+
 void WalletQmlModelTests::prepareTransactionOnLockedWalletRequiresPassword()
 {
     FakePasswordWallet* wallet{nullptr};

```

johnny9: The address list cache can get stale while the page is open.

A minimal fix is to refresh the address model from wallet status/transaction notifications and derive payment-request availability from the refreshed `used` state.

```
diff --git a/qml/models/addresslistmodel.cpp b/qml/models/addresslistmodel.cpp
@@
-        entry.can_create_payment_request = true;
+        entry.can_create_payment_request = !used;
         entries.push_back(entry);
     }

diff --git a/qml/models/walletqmlmodel.cpp b/qml/models/walletqmlmodel.cpp
@@
     m_handler_status_changed = handleStatusChanged([this]() {
         QMetaObject::invokeMethod(this, [this]() {
             refreshSecurityState();
+            if (m_address_list_model) {
+                m_address_list_model->refresh();
+            }
             Q_EMIT balanceChanged();
         }, Qt::QueuedConnection);
     });
     m_handler_transaction_changed = handleTransactionChanged([this](const uint256&, ChangeType) {
         QMetaObject::invokeMethod(this, [this] {
+            if (m_address_list_model) {
+                m_address_list_model->refresh();
+            }
             Q_EMIT balanceChanged();
         }, Qt::QueuedConnection);
     });
 }
```

johnny9 (commented):
Found a few things worth tackling in the context of this PR.

1. Payment request flow shortcut requires some adjustment to the RequestPayment page logic.
2. New address creation failures should be handled
3. Addresses in the list can go stale, watch for changes.
5. Segment picker layout needs adjustment

qml/models/walletqmlmodel.cpp
```diff
@@ -500,6 +521,130 @@ QString WalletQmlModel::getAddressLabel(const QString& address) const
...
+        const Txid txid{wallet_tx.tx->GetHash()};
+        for (size_t i{0}; i < wallet_tx.txout_is_change.size(); ++i) {
+            if (!wallet_tx.txout_is_change[i])
+                continue;
+
+            change_outpoints.insert(COutPoint{txid, static_cast<uint32_t>(i)});
+        }
+    }
+
+    for (const auto& [destination, coins] : m_wallet->listCoins()) {
```
jarolrod: is it correct to just use `ListCoins` here? 

This treats the `listCoins()` map key as the change address, but that key is only for grouping -> `ListCoins()` groups coins with `FindNonChangeParentOutput()`, so a change output can be grouped under the parent receive/input address instead of the actual output script destination.

That means this can display the wrong address for change and attribute balances to the wrong address.

In the widgets gui we handle this by using the `listCoins()` key as a group header *only*, then we derive each output’s actual destination from `tx_out.txout.scriptPubKey`.

I think we should go ahead and basically do the same thing.

The following diff should express what I mean:

<details>

```diff
diff --git a/qml/models/walletqmlmodel.cpp b/qml/models/walletqmlmodel.cpp
index 6a538ed09..873c3ae0b 100644
--- a/qml/models/walletqmlmodel.cpp
+++ b/qml/models/walletqmlmodel.cpp
@@ -35,6 +35,7 @@
 
 #include <array>
 #include <optional>
+#include <tuple>
 
 namespace {
 constexpr unsigned int DEFAULT_STANDARD_FEE_TARGET{2};
@@ -555,13 +556,17 @@ std::map<QString, CAmount> WalletQmlModel::addressBalances() const
         return balances;
     }
 
-    for (const auto& [destination, coins] : m_wallet->listCoins()) {
-        const QString address{QString::fromStdString(EncodeDestination(destination))};
+    for (const auto& coins_entry : m_wallet->listCoins()) {
+        for (const auto& coin : coins_entry.second) {
+            const auto& tx_out{std::get<1>(coin)};
+            CTxDestination destination;
+            if (!ExtractDestination(tx_out.txout.scriptPubKey, destination))
+                continue;
 
-        if (address.isEmpty())
-            continue;
+            const QString address{QString::fromStdString(EncodeDestination(destination))};
+            if (address.isEmpty())
+                continue;
 
-        for (const auto& [outpoint, tx_out] : coins) {
             balances[address] += tx_out.txout.nValue;
         }
     }
@@ -626,18 +631,21 @@ std::set<QString> WalletQmlModel::changeAddresses() const
         }
     }
 
-    for (const auto& [destination, coins] : m_wallet->listCoins()) {
-        const QString address{QString::fromStdString(EncodeDestination(destination))};
-        if (address.isEmpty())
-            continue;
-
-        for (const auto& [outpoint, tx_out] : coins) {
+    for (const auto& coins_entry : m_wallet->listCoins()) {
+        for (const auto& [outpoint, tx_out] : coins_entry.second) {
             if (tx_out.txout.nValue <= 0)
                 continue;
 
             if (change_outpoints.count(outpoint) > 0) {
+                CTxDestination destination;
+                if (!ExtractDestination(tx_out.txout.scriptPubKey, destination))
+                    continue;
+
+                const QString address{QString::fromStdString(EncodeDestination(destination))};
+                if (address.isEmpty())
+                    continue;
+
                 addresses.insert(address);
-                break;
             }
         }
     }
diff --git a/test/test_addresslistmodel.cpp b/test/test_addresslistmodel.cpp
index 97c6e0477..5fa29b85c 100644
--- a/test/test_addresslistmodel.cpp
+++ b/test/test_addresslistmodel.cpp
@@ -178,10 +178,11 @@ void AddressListModelTests::changeAddressesComeFromUnspentChangeOutputs()
     WalletQmlModel wallet_model{std::move(wallet)};
     AddressListModel* model{wallet_model.addressListModel()};
     wallet_ptr->m_txs.insert(change_tx);
-    wallet_ptr->m_coins[change].push_back({COutPoint{txid, 1}, change_out});
+    wallet_ptr->m_coins[receive].push_back({COutPoint{txid, 1}, change_out});
     model->setCategory(AddressListModel::Change);
 
     QCOMPARE(model->rowCount(), 1);
+    QCOMPARE(model->data(model->index(0), AddressListModel::AddressRole).toString(), QString::fromStdString(EncodeDestination(change)));
     QCOMPARE(model->data(model->index(0), AddressListModel::CategoryRole).toString(), QStringLiteral("change"));
     QCOMPARE(model->data(model->index(0), AddressListModel::DisplayAmountRole).toString(), QStringLiteral("₿ 2.00000000"));
     QCOMPARE(model->data(model->index(0), AddressListModel::HasAmountRole).toBool(), true);

```

</details>

qml/pages/wallet/RequestPayment.qml
pseudoramdom: I see, I'll give this a try.

qml/models/walletqmlmodel.cpp
pseudoramdom: Ah I see. `listCoins()` groups by wallet ownership/parent output.
I'll update this to derive each coin’s actual destination from `txout.scriptPubKey`

qml/models/walletqmlmodel.cpp
pseudoramdom: I'm leaning towards removing "+ New Address" button from the Addresses list page. It requires wallet unlock check, user should be able to pick the address type and set a note; which is essentially the Receive Tab.

pseudoramdom:
Picked changes from #563 and rebased. Addressed the feedback as well.

johnny9:
double check the functional tests.

qml/models/walletqmlmodel.h
```diff
@@ -6,10 +6,12 @@
 #define BITCOIN_QML_MODELS_WALLETQMLMODEL_H
 
 #include <qml/models/activitylistmodel.h>
+#include <qml/models/addresslistmodel.h>
 #include <qml/models/bumptransactionmodel.h>
 #include <qml/models/coinslistmodel.h>
 #include <qml/models/paymentrequest.h>
 #include <qml/models/sendrecipientslistmodel.h>
+#include <qml/models/signverifymessagemodel.h>
```
jarolrod, line 14: this is in the first commit, but it doesn't seem to be added until a later commit, so the first commit doesn't compile on its own

Not exactly a blocker for me, but worthwhile to fix

johnny9 (commented):
The functional test regression came from c31ba7d985 which rewired EllipsisMenuToggleItem.qml. The likely fix is to call toggle() on the right object.  It seems we now have a bunch of special cases for "clicked" so we should have the test bridge actually inject mouse pressed/release on the center of the object (x, y + width,height/2)

qml/models/signverifymessagemodel.cpp
```diff
@@ -0,0 +1,163 @@
...
+    relock = false;
+    if (!m_wallet || !m_wallet->isCrypted() || !m_wallet->isLocked()) {
+        return true;
+    }
+    if (!passphrase.has_value()) {
+        setSigningStatus(tr("Enter your wallet password to sign this message."), true);
+        return false;
+    }
+
+    const SecureString secure_passphrase{passphrase->toStdString()};
```
jarolrod: we should stick to using our SecureStringFromQString function instead of passphrase->toStdString.

To avoid confusion in the future, may be worthwhile to make it a helper function and put it to use like so:

<details>

```diff
diff --git a/qml/models/signverifymessagemodel.cpp b/qml/models/signverifymessagemodel.cpp
index 7470b00a0..226d03f00 100644
--- a/qml/models/signverifymessagemodel.cpp
+++ b/qml/models/signverifymessagemodel.cpp
@@ -7,7 +7,7 @@
 #include <common/signmessage.h>
 #include <key_io.h>
 #include <qml/models/walletrelockguard.h>
-#include <support/allocators/secure.h>
+#include <qml/util.h>
 
 namespace {
 std::optional<PKHash> LegacyP2PKHFromAddress(const QString& address)
@@ -124,8 +124,10 @@ bool SignVerifyMessageModel::unlockForSigning(const std::optional<QString>& pass
         return false;
     }
 
-    const SecureString secure_passphrase{passphrase->toStdString()};
-    if (!m_wallet->unlock(secure_passphrase)) {
+    SecureString secure_passphrase{QmlUtil::SecureStringFromQString(*passphrase)};
+    const bool unlocked{m_wallet->unlock(secure_passphrase)};
+    QmlUtil::ClearSecureString(secure_passphrase);
+    if (!unlocked) {
         setSigningStatus(tr("The wallet password you entered was incorrect."));
         return false;
     }
diff --git a/qml/models/walletqmlmodel.cpp b/qml/models/walletqmlmodel.cpp
index 56510f069..7a9904aa8 100644
--- a/qml/models/walletqmlmodel.cpp
+++ b/qml/models/walletqmlmodel.cpp
@@ -15,6 +15,7 @@
 #include <qml/models/signverifymessagemodel.h>
 #include <qml/models/walletrelockguard.h>
 #include <qml/models/walletqmlmodeltransaction.h>
+#include <qml/util.h>
 
 #include <chainparams.h>
 #include <consensus/amount.h>
@@ -33,7 +34,6 @@
 #include <wallet/coincontrol.h>
 #include <wallet/wallet.h>
 
-#include <QByteArray>
 #include <QDateTime>
 #include <QMetaObject>
 #include <QRegularExpression>
@@ -263,26 +263,6 @@ QString LocalizedString(const bilingual_str& value)
     return QString::fromStdString(value.translated.empty() ? value.original : value.translated);
 }
 
-SecureString SecureStringFromQString(const QString& value)
-{
-    QByteArray bytes{value.toUtf8()};
-    SecureString secure;
-    secure.assign(bytes.constData(), bytes.constData() + bytes.size());
-    if (!bytes.isEmpty()) {
-        memory_cleanse(bytes.data(), static_cast<size_t>(bytes.size()));
-    }
-    return secure;
-}
-
-void ClearSecureString(SecureString& value)
-{
-    if (value.empty()) {
-        return;
-    }
-    memory_cleanse(value.data(), value.size());
-    value.clear();
-}
-
 QString OutputTypeId(OutputType type)
 {
     return QString::fromStdString(FormatOutputType(type));
@@ -514,9 +494,9 @@ bool WalletQmlModel::encryptWallet(const QString& passphrase)
         return false;
     }
 
-    SecureString secure_passphrase{SecureStringFromQString(passphrase)};
+    SecureString secure_passphrase{QmlUtil::SecureStringFromQString(passphrase)};
     const bool encrypted{m_wallet->encryptWallet(secure_passphrase)};
-    ClearSecureString(secure_passphrase);
+    QmlUtil::ClearSecureString(secure_passphrase);
     if (!encrypted) {
         setSettingsError(tr("The wallet password could not be set."));
         return false;
@@ -542,11 +522,11 @@ bool WalletQmlModel::changeWalletPassphrase(const QString& old_passphrase, const
         return false;
     }
 
-    SecureString secure_old_passphrase{SecureStringFromQString(old_passphrase)};
-    SecureString secure_new_passphrase{SecureStringFromQString(new_passphrase)};
+    SecureString secure_old_passphrase{QmlUtil::SecureStringFromQString(old_passphrase)};
+    SecureString secure_new_passphrase{QmlUtil::SecureStringFromQString(new_passphrase)};
     const bool changed{m_wallet->changeWalletPassphrase(secure_old_passphrase, secure_new_passphrase)};
-    ClearSecureString(secure_old_passphrase);
-    ClearSecureString(secure_new_passphrase);
+    QmlUtil::ClearSecureString(secure_old_passphrase);
+    QmlUtil::ClearSecureString(secure_new_passphrase);
     if (!changed) {
         setSettingsError(tr("The current wallet password was incorrect."));
         return false;
@@ -1125,7 +1105,7 @@ bool WalletQmlModel::prepareTransaction()
 
 bool WalletQmlModel::prepareTransactionWithPassphrase(const QString& passphrase)
 {
-    return prepareTransactionInternal(std::optional<SecureString>{SecureStringFromQString(passphrase)});
+    return prepareTransactionInternal(std::optional<SecureString>{QmlUtil::SecureStringFromQString(passphrase)});
 }
 
 bool WalletQmlModel::prepareTransactionInternal(std::optional<SecureString> passphrase)
@@ -1133,7 +1113,7 @@ bool WalletQmlModel::prepareTransactionInternal(std::optional<SecureString> pass
     clearTransactionStatus();
     if (!m_wallet || !m_send_recipients || m_send_recipients->recipients().empty()) {
         if (passphrase.has_value()) {
-            ClearSecureString(*passphrase);
+            QmlUtil::ClearSecureString(*passphrase);
             passphrase.reset();
         }
         setTransactionStatus(tr("Enter at least one valid recipient to continue."));
@@ -1499,7 +1479,7 @@ bool WalletQmlModel::unlockForAction(std::optional<SecureString>& passphrase, bo
     relock = false;
     if (!m_wallet || !m_wallet->isCrypted() || !m_wallet->isLocked()) {
         if (passphrase.has_value()) {
-            ClearSecureString(*passphrase);
+            QmlUtil::ClearSecureString(*passphrase);
             passphrase.reset();
         }
         return true;
@@ -1509,7 +1489,7 @@ bool WalletQmlModel::unlockForAction(std::optional<SecureString>& passphrase, bo
     }
 
     const bool unlocked = m_wallet->unlock(*passphrase);
-    ClearSecureString(*passphrase);
+    QmlUtil::ClearSecureString(*passphrase);
     passphrase.reset();
     if (!unlocked) {
         setTransactionStatus(tr("The wallet password you entered was incorrect."));
diff --git a/qml/util.cpp b/qml/util.cpp
index c2532f02e..065491a57 100644
--- a/qml/util.cpp
+++ b/qml/util.cpp
@@ -4,9 +4,13 @@
 
 #include <qml/util.h>
 
+#include <support/cleanse.h>
+
 #include <cassert>
+#include <cstddef>
 #include <string>
 
+#include <QByteArray>
 #include <QQuickWindow>
 #include <QSGRendererInterface>
 #include <QString>
@@ -33,4 +37,24 @@ QString GraphicsApi(QQuickWindow* window)
     assert(false);
 }
 
+SecureString SecureStringFromQString(const QString& value)
+{
+    QByteArray bytes{value.toUtf8()};
+    SecureString secure;
+    secure.assign(bytes.constData(), bytes.constData() + bytes.size());
+    if (!bytes.isEmpty()) {
+        memory_cleanse(bytes.data(), static_cast<std::size_t>(bytes.size()));
+    }
+    return secure;
+}
+
+void ClearSecureString(SecureString& value)
+{
+    if (value.empty()) {
+        return;
+    }
+    memory_cleanse(value.data(), value.size());
+    value.clear();
+}
+
 } // namespace QmlUtil
diff --git a/qml/util.h b/qml/util.h
index 6df5d2c2b..32930a5eb 100644
--- a/qml/util.h
+++ b/qml/util.h
@@ -5,6 +5,8 @@
 #ifndef BITCOIN_QML_UTIL_H
 #define BITCOIN_QML_UTIL_H
 
+#include <support/allocators/secure.h>
+
 #include <QString>
 #include <QtGlobal>
 
@@ -22,6 +24,9 @@ namespace QmlUtil {
 */
 QString GraphicsApi(QQuickWindow* window);
 
+SecureString SecureStringFromQString(const QString& value);
+void ClearSecureString(SecureString& value);
+
 } // namespace QmlUtil
 
 #endif // BITCOIN_QML_UTIL_H
diff --git a/qml/walletqmlcontroller.cpp b/qml/walletqmlcontroller.cpp
index a16cd57bf..50c5af3f4 100644
--- a/qml/walletqmlcontroller.cpp
+++ b/qml/walletqmlcontroller.cpp
@@ -5,6 +5,7 @@
 #include <qml/walletqmlcontroller.h>
 
 #include <qml/models/walletqmlmodel.h>
+#include <qml/util.h>
 
 #include <common/args.h>
 #include <common/settings.h>
@@ -21,7 +22,6 @@
 #include <stdexcept>
 #include <utility>
 
-#include <QByteArray>
 #include <QDir>
 #include <QFileInfo>
 #include <QMetaObject>
@@ -43,26 +43,6 @@ QString JoinWarnings(const std::vector<bilingual_str>& warnings)
     return lines.join('\n');
 }
 
-SecureString SecureStringFromQString(const QString& value)
-{
-    QByteArray bytes{value.toUtf8()};
-    SecureString secure;
-    secure.assign(bytes.constData(), bytes.constData() + bytes.size());
-    if (!bytes.isEmpty()) {
-        memory_cleanse(bytes.data(), static_cast<size_t>(bytes.size()));
-    }
-    return secure;
-}
-
-void ClearSecureString(SecureString& value)
-{
-    if (value.empty()) {
-        return;
-    }
-    memory_cleanse(value.data(), value.size());
-    value.clear();
-}
-
 bool IsBackupLikeFile(const QFileInfo& file_info)
 {
     const QString file_name = file_info.fileName().toLower();
@@ -312,10 +292,10 @@ bool WalletQmlController::createSingleSigWallet(const QString &name, const QStri
         setWalletCreateError(tr("Wallets are still loading. Try again in a moment."));
         return false;
     }
-    SecureString secure_passphrase{SecureStringFromQString(passphrase)};
+    SecureString secure_passphrase{QmlUtil::SecureStringFromQString(passphrase)};
     const std::string wallet_name{name.toStdString()};
     auto wallet{m_node.walletLoader().createWallet(wallet_name, secure_passphrase, wallet::WALLET_FLAG_DESCRIPTORS, m_warning_messages)};
-    ClearSecureString(secure_passphrase);
+    QmlUtil::ClearSecureString(secure_passphrase);
     setWalletLoadWarnings(JoinWarnings(m_warning_messages));
     if (wallet) {
         const QString loaded_wallet_name = QString::fromStdString((*wallet)->getWalletName());
@@ -436,7 +416,7 @@ void WalletQmlController::migrateWallet(const QString& path, const QString& pass
         setWalletMigrationError(tr("Wallets are still loading. Try again in a moment."));
         return;
     }
-    startWalletMigration(path, SecureStringFromQString(passphrase));
+    startWalletMigration(path, QmlUtil::SecureStringFromQString(passphrase));
 }
 
 void WalletQmlController::clearWalletMigrationStatus()
@@ -937,7 +917,7 @@ void WalletQmlController::startWalletMigration(const QString& path, SecureString
     clearWalletMigrationStatus();
 
     if (path.trimmed().isEmpty()) {
-        ClearSecureString(passphrase);
+        QmlUtil::ClearSecureString(passphrase);
         setWalletMigrationError(tr("Choose a wallet to update."));
         Q_EMIT walletMigrationFailed();
         return;
@@ -945,7 +925,7 @@ void WalletQmlController::startWalletMigration(const QString& path, SecureString
 
     const QString wallet_reference = resolveManagedWalletReference(path);
     if (wallet_reference.isEmpty()) {
-        ClearSecureString(passphrase);
+        QmlUtil::ClearSecureString(passphrase);
         setWalletMigrationError(tr("The selected wallet is not available in the wallet directory."));
         Q_EMIT walletMigrationFailed();
         return;
@@ -960,7 +940,7 @@ void WalletQmlController::startWalletMigration(const QString& path, SecureString
 
     QTimer::singleShot(0, m_worker, [this, wallet_reference, passphrase = std::move(passphrase)]() mutable {
         auto result = m_node.walletLoader().migrateWallet(wallet_reference.toStdString(), passphrase);
-        ClearSecureString(passphrase);
+        QmlUtil::ClearSecureString(passphrase);
 
         if (!result) {
             const QString error = QString::fromStdString(util::ErrorString(result).translated);
```
</details>

qml/models/walletqmlmodel.cpp
```diff
@@ -578,29 +665,57 @@ void WalletQmlModel::removeWallet()
...
+            if (!parsed_type) {
+                return false;
+            }
+            output_type = *parsed_type;
+        }
         const auto destination{m_wallet->getNewDestination(output_type, m_current_payment_request->label().toStdString())};
-        if (!destination) {
-            return;
+        if (!destination || !IsValidDestination(destination.value())) {
+            return false;
```
jarolrod, line 680: We should have retry logic like the qt widgets gui does, and here is some more context as to why: https://github.com/bitcoin/bitcoin/pull/2904

johnny9: can check if isLocked first

pseudoramdom: Ah interesting! Just today I learned why we might need to ask for passphrase if we need to replenish the key pool. I'll look in to this tomorrow.

johnny9:
Can you go ahead and fix https://github.com/bitcoin-core/gui-qml/issues/712 when you address the current round of comments. It should just be a simple adjustment to the default focus of the page.

pseudoramdom:
Also fixes #712

johnny9 (commented):
In file included from ../../test/mocks/mockwallet.h:21,
                 from ../../test/test_addresslistmodel.cpp:7:
In copy constructor ‘interfaces::WalletTxOut::WalletTxOut(const interfaces::WalletTxOut&)’,
    inlined from ‘constexpr std::_Head_base<_Idx, _Head, false>::_Head_base(_UHead&&) [with _UHead = interfaces::WalletTxOut&; long unsigned int _Idx = 1; _Head = interfaces::WalletTxOut]’ at /usr/include/c++/16/tuple:213:4,
    inlined from ‘constexpr std::_Tuple_impl<_Idx, _Head>::_Tuple_impl(_UHead&&) [with _UHead = interfaces::WalletTxOut&; long unsigned int _Idx = 1; _Head = interfaces::WalletTxOut]’ at /usr/include/c++/16/tuple:571:38,
    inlined from ‘constexpr std::_Tuple_impl<_Idx, _Head, _Tail ...>::_Tuple_impl(_UHead&&, _UTail&& ...) [with _UHead = COutPoint; _UTail = {interfaces::WalletTxOut&}; <template-parameter-2-3> = void; long unsigned int _Idx = 0; _Head = COutPoint; _Tail = {interfaces::WalletTxOut}]’ at /usr/include/c++/16/tuple:315:38,
    inlined from ‘constexpr std::tuple< <template-parameter-1-1> >::tuple(_UTypes&& ...) [with _UTypes = {COutPoint, interfaces::WalletTxOut&}; _Elements = {COutPoint, interfaces::WalletTxOut}]’ at /usr/include/c++/16/tuple:985:44,
    inlined from ‘void AddressListModelTests::changeAddressesComeFromUnspentChangeOutputs()’ at ../../test/test_addresslistmodel.cpp:186:43:
../../bitcoin/src/interfaces/wallet.h:417:8: warning: ‘change_out.interfaces::WalletTxOut::time’ may be used uninitialized [-Wmaybe-uninitialized]
  417 | struct WalletTxOut
      |        ^~~~~~~~~~~
../../test/test_addresslistmodel.cpp: In member function ‘void AddressListModelTests::changeAddressesComeFromUnspentChangeOutputs()’:
../../test/test_addresslistmodel.cpp:179:29: note: ‘change_out’ declared here
  179 |     interfaces::WalletTxOut change_out;
      |                             ^~~~~~~~~~
[ 96%] Linking CXX executable ../bin/bitcoinqml_unit_tests

johnny9 (approved):
ACK 26d91ab48175b1b585c780780b725affc2eba114

I believe all major issues have been addressed and there is sufficient test coverage of the features.

There is a minor quirk with the verification UX flow when the signature remains visible after editing the message and is worth looking at as a part of a design review of the Preview release.

pseudoramdom:
> There is a minor quirk with the verification UX flow when the signature remains visible after editing the message and is worth looking at as a part of a design review of the Preview release.

Good catch. Created https://github.com/bitcoin-core/gui-qml/issues/717 to track this

qml/models/walletqmlmodel.cpp
pseudoramdom: Took this feedback. Implemented retry logic. Used this chance to refactor the unlock related to `walletunlock.h`

qml/models/walletqmlmodel.cpp
```diff
@@ -694,6 +855,136 @@ QString WalletQmlModel::getAddressLabel(const QString& address) const
...
+                const QString address{QString::fromStdString(EncodeDestination(destination))};
+                if (address.isEmpty())
+                    continue;
+
+                addresses.insert(address);
+            }
+        }
+    }
+
+    return addresses;
```
achow101, line 985: ISTM enumerating the change addresses should be something that is handled in the wallet, not in the GUI.

I'm also not really a fan of this approach as it iterates every single transaction in the wallet, which can be a slow process. Although I concede that there probably isn't a better way to list out change addresses since we don't store them.

achow101:
The placement of these pages seems a bit odd to me as they aren't really settings. It took me a little bit to find them.

pseudoramdom:
> The placement of these pages seems a bit odd to me as they aren't really settings. It took me a little bit to find them.

It is not ideal, agree. I think there's a revised design of settings page which we haven't implemented. Maybe we can revisit the entry point for this then? Cc @johnny9

qml/models/signverifymessagemodel.h
```diff
@@ -0,0 +1,61 @@
...
+    Q_INVOKABLE void clear();
+    Q_INVOKABLE void clearSigningStatus();
+
+Q_SIGNALS:
+    void signingErrorChanged();
+    void signingNeedsUnlockChanged();
+    void signatureChanged();
+
+private:
+    bool signMessageInternal(const QString& address, const QString& message, const std::optional<QString>& passphrase);
```
jarolrod, line 48: why QString for passphrase here? It's an unecessary QString that can live for a long time in memory, should be SecureString here

Can do the following:

<details>

```diff
diff --git a/qml/models/signverifymessagemodel.cpp b/qml/models/signverifymessagemodel.cpp
index 264aeeee5..119dc3dc0 100644
--- a/qml/models/signverifymessagemodel.cpp
+++ b/qml/models/signverifymessagemodel.cpp
@@ -52,20 +52,28 @@ bool SignVerifyMessageModel::signMessage(const QString& address, const QString&
 
 bool SignVerifyMessageModel::signMessageWithPassphrase(const QString& address, const QString& message, const QString& passphrase)
 {
-    return signMessageInternal(address, message, passphrase);
+    return signMessageInternal(address, message, QmlUtil::SecureStringFromQString(passphrase));
 }
 
-bool SignVerifyMessageModel::signMessageInternal(const QString& address, const QString& message, const std::optional<QString>& passphrase)
+bool SignVerifyMessageModel::signMessageInternal(const QString& address, const QString& message, std::optional<SecureString> passphrase)
 {
     clearSigningStatus();
     setSignature(QString());
 
     if (!m_wallet) {
+        if (passphrase.has_value()) {
+            QmlUtil::ClearSecureString(*passphrase);
+            passphrase.reset();
+        }
         setSigningStatus(tr("No wallet is selected."));
         return false;
     }
     const auto pkhash{LegacyP2PKHFromAddress(address)};
     if (!pkhash) {
+        if (passphrase.has_value()) {
+            QmlUtil::ClearSecureString(*passphrase);
+            passphrase.reset();
+        }
         setSigningStatus(tr("Enter a legacy P2PKH bitcoin address."));
         return false;
     }
@@ -114,13 +122,21 @@ void SignVerifyMessageModel::clearSigningStatus()
     setSigningStatus(QString());
 }
 
-bool SignVerifyMessageModel::unlockForSigning(const std::optional<QString>& passphrase, bool& relock)
+bool SignVerifyMessageModel::unlockForSigning(std::optional<SecureString>& passphrase, bool& relock)
 {
     relock = false;
     if (!m_wallet) {
+        if (passphrase.has_value()) {
+            QmlUtil::ClearSecureString(*passphrase);
+            passphrase.reset();
+        }
         return true;
     }
     if (!m_wallet->isCrypted() || !m_wallet->isLocked()) {
+        if (passphrase.has_value()) {
+            QmlUtil::ClearSecureString(*passphrase);
+            passphrase.reset();
+        }
         return true;
     }
     if (!passphrase.has_value()) {
@@ -128,8 +144,9 @@ bool SignVerifyMessageModel::unlockForSigning(const std::optional<QString>& pass
         return false;
     }
 
-    SecureString secure_passphrase{QmlUtil::SecureStringFromQString(*passphrase)};
-    switch (TryUnlockWithPassphrase(*m_wallet, secure_passphrase)) {
+    const auto result{TryUnlockWithPassphrase(*m_wallet, *passphrase)};
+    passphrase.reset();
+    switch (result) {
     case WalletUnlockResult::IncorrectPassphrase:
         setSigningStatus(tr("The wallet password you entered was incorrect."));
         return false;
diff --git a/qml/models/signverifymessagemodel.h b/qml/models/signverifymessagemodel.h
index 88a50a2ed..283375952 100644
--- a/qml/models/signverifymessagemodel.h
+++ b/qml/models/signverifymessagemodel.h
@@ -6,6 +6,7 @@
 #define BITCOIN_QML_MODELS_SIGNVERIFYMESSAGEMODEL_H
 
 #include <interfaces/wallet.h>
+#include <support/allocators/secure.h>
 
 #include <functional>
 #include <optional>
@@ -45,8 +46,8 @@ Q_SIGNALS:
     void signatureChanged();
 
 private:
-    bool signMessageInternal(const QString& address, const QString& message, const std::optional<QString>& passphrase);
-    bool unlockForSigning(const std::optional<QString>& passphrase, bool& relock);
+    bool signMessageInternal(const QString& address, const QString& message, std::optional<SecureString> passphrase);
+    bool unlockForSigning(std::optional<SecureString>& passphrase, bool& relock);
     void setSigningStatus(const QString& error, bool needs_unlock = false);
     void setSignature(const QString& signature);
     void notifySecurityStateChanged();
```
</details>

qml/models/signverifymessagemodel.cpp
```diff
@@ -0,0 +1,171 @@
...
+    if (result != SigningResult::OK) {
+        setSigningStatus(QString::fromStdString(SigningResultString(result)));
+        return false;
+    }
+
+    setSignature(QString::fromStdString(signature));
+    return true;
+}
+
+bool SignVerifyMessageModel::verifyMessage(const QString& address, const QString& message, const QString& signature) const
```
jarolrod, line 92: is there a reason we're not doing the detailed error message that we get from MessageVerify?

For example we have `MessageVerificationResult::ERR_ADDRESS_NO_KEY`, `MessageVerificationResult::ERR_INVALID_ADDRESS:`, `MessageVerificationResult::ERR_MALFORMED_SIGNATURE` etc etc that all can give a detailed error message.

This is something the qt widgets gui does.

jarolrod:
ACK https://github.com/bitcoin-core/gui-qml/commit/26d91ab48175b1b585c780780b725affc2eba114

I'm fine with remaining points being tackled as follow-ups, since this PR is already large

johnny9:
> > The placement of these pages seems a bit odd to me as they aren't really settings. It took me a little bit to find them.
> 
> It is not ideal, agree. I think there's a revised design of settings page which we haven't implemented. Maybe we can revisit the entry point for this then? Cc @johnny9

We will be designing a better home for these pages at some point this summer and have already started looking at removing a lot of things from the Settings page. The Settings page is a bit of a dumping ground right now for features that don't have a good home in the current design.

qml/models/signverifymessagemodel.h
johnny9: #719

qml/models/signverifymessagemodel.cpp
johnny9: #718

qml/models/walletqmlmodel.cpp
johnny9: #720

qml/models/signverifymessagemodel.cpp
pseudoramdom: Probably an oversight. I don't recall referring the qt widgets for this, which I should have.

qml/models/walletqmlmodel.cpp
pseudoramdom: Are we open to the idea of exposing a `Wallet::listChangeAddresses()` interface? 
An alternative would be to add a `bool is_change` property to `WalletTxOut`. That way we can just filter via `listCoins()` and avoid the walking through the transactions.

PR #535, opened 2026-03-13
Decouple BanListModel from qt BantableModel

johnny9:
Decouples `BanListModel` from `qt/bantablemodel.h` and removes the stale `qt/guiutil.h` include from `qml/bitcoin.cpp`.

Together these remove the remaining `#include <qt/...>` references from the QML runtime source scan.

Related to #505.

Testing:
- `git grep -n -E "#include <qt/|#include <bitcoin/src/qt/|qt/guiutil|GUIUtil|qt/bantablemodel" -- qml test CMakeLists.txt || true`

PR #562, opened 2026-04-30
qml: remove unused qt guiutil include

johnny9:
This removes the stale `qt/guiutil.h` include from `qml/bitcoin.cpp`. The file no longer references `GUIUtil`, so the include is unnecessary and keeps a `bitcoin/src/qt` dependency visible in the QML runtime source scan.

Related to #505.

Testing:
- `git grep -n -E "#include <qt/guiutil.h>|GUIUtil" -- qml test CMakeLists.txt || true`
- `cmake -S . -B /tmp/gui-qml-remove-guiutil-build -DBUILD_APP_TESTS=ON`
- `cmake --build /tmp/gui-qml-remove-guiutil-build --target bitcoinqml -j2`

### DIFF
--- a/qml/bitcoin.cpp
+++ b/qml/bitcoin.cpp
@@ -18,7 +18,6 @@
 #include <node/context.h>
 #include <node/interface_ui.h>
 #include <noui.h>
-#include <qt/guiutil.h>
 #include <qml/appmode.h>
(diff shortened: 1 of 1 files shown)


johnny9:
Folded this one-line include cleanup into #535 so the src/qt dependency cleanup is reviewed in one PR.

PR #564, opened 2026-05-04
qml: Add typography design tokens to Theme

pseudoramdom:
Currently every text hardcodes `font.family: "Inter"` and raw pixel sizes.
This PR establishes a typography layer of the design system inside the existing `Theme`.
It introduces semantic text roles so consumers can write font: `Theme.text.body.font` instead

Additionally, add a debug-only preview page to visually verify all roles in one place. 
It lives under Settings → Display → Theme, in a new "Developer" section that's hidden in release builds.

Screenshot 2026-05-04 at 11 55 12 AM
Screenshot 2026-05-04 at 11 41 50 AM
Screenshot 2026-05-04 at 11 42 04 AM

### DIFF
--- a/qml/bitcoin.cpp
+++ b/qml/bitcoin.cpp
@@ -20,6 +20,7 @@
 #include <noui.h>
 #include <qml/appmode.h>
 #include <qml/bitcoinamount.h>
+#include <qml/buildinfo.h>
 #include <qml/clipboard.h>
 #ifdef __ANDROID__
 #include <qml/androidnotifier.h>
@@ -129,8 +130,10 @@ bool InitErrorMessageBox(
     QQmlApplicationEngine engine;
 
     AppMode app_mode = SetupAppMode();
+    BuildInfo build_info;
 
     qmlRegisterSingletonInstance<AppMode>("org.bitcoincore.qt", 1, 0, "AppMode", &app_mode);
+    qmlRegisterSingletonInstance<BuildInfo>("org.bitcoincore.qt", 1, 0, "BuildInfo", &build_info);
     engine.rootContext()->setContextProperty("message", QString::fromStdString(message.translated));
     engine.load(QUrl(QStringLiteral("qrc:///qml/pages/initerrormessage.qml")));
     if (engine.rootObjects().isEmpty()) {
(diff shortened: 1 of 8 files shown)


pseudoramdom:
@johnny9 We have some places where we still reference the old "Inter" font family on `qt6`. Instead of fixing that, I'd like to merge this so that we refer to the semantic text styles from the `Theme` going forward.

johnny9 (approved):
ACK 1d613eed81799b17cb812e7cba1d33c18faed239

GBKS:
Nice one. Are these based on the typographic scale defined in Figma and [on the website](https://bitcoincore.app/typography/)?

GBKS:
And maybe a small note about naming these tokens. I find that semantic tokens like [the ones Apple uses](https://developer.apple.com/design/human-interface-guidelines/typography#Medium) not really useful for application design. Terms like callout, subhead and footnote seem to stem from print design and don't carry over well. Which is why we went for a simple laddering of header and body sizes for the core app, with the main differentiator being that header sizes are semibold.

If we really wanted to flesh this out, we could use a similar system I also like for color variables. Generic colors like green, blue, red, which map to semantic ones like success, progress, and error. We could use the scale we have defined in Figma and on the website, and then come up with some specific semantic names (screen title, form label...) that match the needs of our application (not unrelated print terms).

Just some thoughts and ideas. But we should definitely not overcomplicate it.

pseudoramdom:
> Are these based on the typographic scale defined in Figma and [on the website](https://bitcoincore.app/typography/)?

Yes, the values are based on the [website](https://bitcoincore.app/typography/).

> And maybe a small note about naming these tokens. I find that semantic tokens like [the ones Apple uses](https://developer.apple.com/design/human-interface-guidelines/typography#Medium) not really useful for application design. Terms like callout, subhead and footnote seem to stem from print design and don't carry over well. Which is why we went for a simple laddering of header and body sizes for the core app, with the main differentiator being that header sizes are semibold.
> 
> If we really wanted to flesh this out, we could use a similar system I also like for color variables. Generic colors like green, blue, red, which map to semantic ones like success, progress, and error. We could use the scale we have defined in Figma and on the website, and then come up with some specific semantic names (screen title, form label...) that match the needs of our application (not unrelated print terms).
> 
> Just some thoughts and ideas. But we should definitely not overcomplicate it.

This is good feedback, agree on moving away from print-derived terms. And yes, having specific semantic names like formLabel, button, menuItem would carry their intent more clearly at the call site than a numeric scale.

Same with semantic colors -> success, progress/info, error

Issue #560, opened 2026-04-29
Feature Parity 23: Watch-only Wallet Support


johnny9:
## Summary
Add watch-only wallet parity so QML can create/import wallets with private keys disabled, surface them clearly as watch-only, and gate receive/send/signing actions according to wallet capabilities.

## Complexity
- `L`

## Dependencies
- `10`
- `14`
- `18`
- `02`

## Global Constraints
- All implementation lives under `gui-qml/`.
- No GUI runtime dependency on `bitcoin/src/qt` components. Reused logic must be moved/ported into `gui-qml/`.
- Do not change `interfaces::Wallet` or `interfaces::Node` in this version.
- Issue completion must satisfy `DefinitionOfDone.md`.

## Scope
- Add a watch-only wallet option to the add/create wallet flow.
- Create descriptor watch-only wallets with private keys disabled and no passphrase-only steps that do not apply.
- Expose wallet capability metadata from the backend/model layer, including private-key-disabled state and address-generation availability.
- Import/load existing watch-only wallets and show a consistent wallet type in the selector, wallet badge, and wallet settings.
- Add or route descriptor/address import UX for watch-only material; if the current interfaces are insufficient, record the exact gap in the interface proposals log.
- Gate receive, send, sign-message, backup, and passphrase actions based on watch-only capabilities.
- Route transaction creation toward PSBT/export flow when direct signing or broadcast is not available.
- Keep balances, activity, and coin selection correct for watch-only and solvable-only outputs.

## Acceptance Criteria (DoD-Aligned)
1. User can create a watch-only wallet from the add wallet flow, restart the app, and see it reload as watch-only.
2. User can import/load an existing watch-only wallet and the wallet selector/settings clearly identify it as watch-only.
3. Actions requiring private keys are disabled or rerouted with explicit explanations; no signing/broadcast path silently fails.
4. Receive behavior is correct for watch-only wallets: unavailable when no receive descriptors exist, available when imported descriptors support address generation.
5. Balances, activity rows, and coin selection distinguish watch-only funds from directly spendable funds.
6. Missing descriptor import support or other interface gaps are captured as concrete follow-up entries instead of being hidden in UI code.

## Required Tests
- C++ unit tests: Unit tests for wallet creation flags, wallet capability metadata, and watch-only balance/action classification.
- QML QTest: QTests for watch-only create/import flows, wallet type display, and gated receive/send/signing states.
[...]

PR #565, opened 2026-05-04
Watch only wallet

epicleafies:
Added the menu for all wallet options when adding a wallet.
Added the watch only wallet creation flow. Currently password protection isn't supported, may want to consider for the future.
Added unit, qml and functional tests for new features.
Fixed bug where the gui would crash when leaving the on-boarding sequence.
Fixed bug where long wallet names would break the wallet name text boxes.
One TODO added. The wallet creation process should be moved to its own worker thread to avoid UI freezes. Out of scope for this PR.
#560 

image

image

image

image


pseudoramdom:
Pulled this into https://github.com/bitcoin-core/gui-qml/pull/727. We can close this.

PR #547, opened 2026-04-09
External signer wallet create and send

johnny9:
06-mock-signer-detected
07-add-wallet-flow-opened
09-external-wallet-form-opened
11-external-wallet-created
19-no-signer-review-displayed
23-signer-approval-completed
20-no-signer-review-error-surfaced

PR #551, opened 2026-04-20
Wallet close, settings, and mempool information

johnny9:
05-wallet-settings-opened
image
image
image
image

Closes #509 
Closes #515
Closes #698

PR #686, opened 2026-05-06
qml: Validate wallet name during wallet creation

### LINKED ISSUE #571: Create wallet: validate wallet name availability before continuing
## Code Gap
The QML create-wallet flow accepts any syntactically valid wallet name and only surfaces duplicate-name failures after the user continues through the flow and attempts wallet creation.

Qt prevents or reports invalid create-wallet options before the expensive create step. For Preview, the QML flow should not let users invest in the password/confirmation path when the wallet name is already unavailable.

## Work Needed
- Check the wallet name against discovered wallet directory entries before leaving the name step.
- Also account for already-loaded wallets.
- Surface duplicate or unavailable wallet-name errors on the name page.
- Apply the same name availability behavior to external-signer wallet creation.
- Add coverage for available name, duplicate closed wallet, duplicate loaded wallet, and invalid name.

pseudoramdom:
Fixes https://github.com/bitcoin-core/gui-qml/issues/571

When creating a wallet, the name is checked for
- invalid/empty names
- discovered wallet names in wallet dir
- existing loaded wallets

The same checks are performed for external-signer wallets.

### Review notes
~~This PR is stacked on top of #547.~~ Now rebased on top of `qt6`

## Screenshots
Screenshot 2026-05-06 at 8 50 36 AM

### DIFF
--- a/qml/pages/wallet/CreateName.qml
+++ b/qml/pages/wallet/CreateName.qml
@@ -79,7 +79,7 @@ Page {
             text: root.walletNameError.length > 0 ? root.walletNameError : walletController.walletLoadError
             color: Theme.color.red
             horizontalAlignment: Text.AlignLeft
-            font.pixelSize: 14
+            font: Theme.text.caption.font
         }
 
         ContinueButton {
@@ -97,7 +97,7 @@ Page {
                     root.walletNameError = availabilityError
                     return
                 }
-                root.walletName = walletNameInput.text.trim()
+                root.walletName = walletNameInput.text
                 root.next()
             }
         }
--- a/qml/pages/wallet/CreateName.qml
+++ b/qml/pages/wallet/CreateName.qml
@@ -20,6 +20,7 @@ Page {
     property int walletType: CreateName.WalletType.SingleSig
     property string initialWalletName: ""
     readonly property bool externalSignerWallet: walletType === CreateName.WalletType.ExternalSigner
+    property string walletNameError: ""
     background: null
 
     Component.onCompleted: {
@@ -61,22 +62,24 @@ Page {
             placeholderText: root.externalSignerWallet
                 ? qsTr("Eg. Hardware wallet...")
                 : qsTr("Eg. My bitcoin wallet...")
-            validator: RegularExpressionValidator { regularExpression: /^[a-zA-Z0-9_]{1,20}$/ }
             onTextChanged: {
                 walletController.clearWalletLoadStatus()
                 root.walletName = walletNameInput.text
+                root.walletNameError = ""
                 continueButton.enabled = walletNameInput.text.length > 0
             }
         }
 
         CoreText {
-            visible: walletController.walletLoadError.length > 0
+            objectName: "walletNameError"
             Layout.fillWidth: true
             Layout.leftMargin: 20
             Layout.rightMargin: 20
+            visible: root.walletNameError.length > 0 || walletController.walletLoadError.length > 0
+            text: root.walletNameError.length > 0 ? root.walletNameError : walletController.walletLoadError
             color: Theme.color.red
-            wrapMode: Text.WordWrap
-            text: walletController.walletLoadError
+            horizontalAlignment: Text.AlignLeft
+            font.pixelSize: 14
         }
 
         ContinueButton {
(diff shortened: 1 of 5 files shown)


pseudoramdom:
I could've written unit test, but the scaffolding for test_walletcontroller is all in #551 already. 
Felt it's easier to add later rather than picking that commit and dropping it later when #551 is merged. Thoughts @johnny9 ?

johnny9:
> I could've written unit test, but the scaffolding for test_walletcontroller is all in #551 already. 
> Felt it's easier to add later rather than picking that commit and dropping it later when #551 is merged. Thoughts @johnny9 ?

I'm fine with that.

pseudoramdom:
558cf9c1b722adbbf294a6c9901a3cc0d57caadd
Rebased on top of `qt6` now

qml/walletqmlcontroller.cpp
```diff
@@ -306,6 +319,55 @@ bool WalletQmlController::walletPathExists(const QString& path) const
...
+        }
+    }
+
+    return false;
+}
+
+QString WalletQmlController::walletNameAvailabilityError(const QString& name) const
+{
+    const QString candidate = trimmedWalletName(name);
+    static const QRegularExpression valid_name(QStringLiteral("^[A-Za-z0-9_]{1,20}$"));
```
MarnixCroes: IMO this makes sense.

Note that bitcoin-qt, allows special characters, more than 20 characters and does not trim.
IDK if this was a purposeful decision (back then).

qml/walletqmlcontroller.cpp
```diff
@@ -306,6 +319,55 @@ bool WalletQmlController::walletPathExists(const QString& path) const
...
+
+    if (candidate.isEmpty()) {
+        return tr("Enter a wallet name.");
+    }
+
+    if (!valid_name.match(candidate).hasMatch()) {
+        return tr("Wallet names can use 1-20 letters, numbers, and underscores.");
```
MarnixCroes:
```suggestion
        return tr("Wallet names can use up to 20 characters (letters, numbers and underscores).");
```

MarnixCroes: nit

MarnixCroes (commented):
558cf9c1b722adbbf294a6c9901a3cc0d57caadd

tested a bit, works as expected ✅

qml/walletqmlcontroller.cpp
davidgumberg: in bitcoin-qt and bitcoin core wallet names are not distinct from paths, so whatever characters are permitted in a path are permitted in a wallet name..

This validation rule disallows wallet names that use utf-8 I am not sure if that's intentional?

I wonder if wallet name rules belong to the wallet itself rather than the gui controller. Related is that users with existing wallets that violate these rules won't be able to load them in the gui.

johnny9: validation rules in the gui can give faster feedback to the user so they are useful but the current validation rules are too restrictive. it shouldn't prevent the user from loading their 🚀🚀🚀🚀🚀🚀 wallet but we should allow utf-8 wallet names since it is valid

image

pseudoramdom: Agree. I'll relax the regex to allow utf-8. 
Additionally, David suggested moving the wallet name validation to wallet and exposing a wallet interface for it. (as a follow-up, not required for this PR)

qml/pages/wallet/CreateName.qml
```diff
@@ -61,22 +62,24 @@ Page {
             placeholderText: root.externalSignerWallet
                 ? qsTr("Eg. Hardware wallet...")
                 : qsTr("Eg. My bitcoin wallet...")
```
jarolrod, line 64: here's an interesting nit I noticed, should I placeholder test represent valid names, as in should the placeholder test instead be: "my_bitcoin_wallet"

pseudoramdom:
Hi @GBKS - is there a design spec to show validation errors? Would it be toast?

pseudoramdom:
Pulled this into https://github.com/bitcoin-core/gui-qml/pull/727. Closing this PR.

PR #687, opened 2026-05-06
qml: Remove back button after wallet creation

### LINKED ISSUE #572: Create wallet: remove back navigation after successful creation
## Code Gap
After a wallet is successfully created, `CreateConfirm` and `CreateBackup` still expose Back navigation. That lets the user navigate back into pre-create steps even though the wallet already exists and has become the selected wallet.

This can lead to duplicate-create attempts, confusing wallet state, and unclear recovery from the post-create backup step.

## Work Needed
- Remove or replace Back navigation after the wallet has been created.
- Ensure the success and backup steps have a single clear route forward or out of the wizard.
- Prevent duplicate create attempts from post-success navigation.
- Preserve the ability to cancel before wallet creation begins.
- Add coverage for unencrypted and encrypted wallet creation success paths.

pseudoramdom:
Fixes https://github.com/bitcoin-core/gui-qml/issues/572

This change adds a small shared navigation policy to `PageStack` and ties `NavigationBar2` into it. `PageStack` now exposes `canGoBack` and `goBack()`, 
- pages can opt out by setting `navigationBackEnabled: false` 
- `NavigationBar2` can now take a navigationStack and render a default Back button based on that stack’s `canGoBack`, calling `goBack()` when clicked.

The wallet creation flow now uses this shared behavior instead of hand-rolling Back buttons and back signal.

### DIFF
--- a/qml/controls/NavigationBar2.qml
+++ b/qml/controls/NavigationBar2.qml
@@ -10,6 +10,12 @@ Pane {
     property alias leftItem: left_section.contentItem
     property alias centerItem: center_section.contentItem
     property alias rightItem: right_section.contentItem
+    property var navigationStack: null
+    property bool showBackButton: navigationStack ? navigationStack.canGoBack : false
+    property string backButtonObjectName: ""
+    property string backButtonText: qsTr("Back")
+
+    signal backClicked
 
     background: null
     padding: 4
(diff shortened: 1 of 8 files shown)


johnny9 (approved):
ACK 658882b21cdfeb70c4800c3e8e8c0c1b753eb46d

Issue #707, opened 2026-05-17
Activity filters: add advanced date range, amount, and Other filters


johnny9:
## Code Gap
The QML Activity page has the currently designed filter set: search text, date presets, primary transaction type filters, reset behavior, filtered-empty state, and filtered CSV export.

Qt widgets also support additional transaction-history filters that are not part of the current QML design: minimum amount filtering, a custom date range, a `Last month` date preset, and a distinct `Other` transaction type filter.

## Work Needed
- Add a minimum amount filter to Activity if retained for parity.
- Add custom date range filtering if retained for parity.
- Add the `Last month` date preset if retained for parity.
- Add an `Other` type filter instead of folding `Transaction::Other` into Sent.
- Ensure advanced filters apply consistently to the visible list and CSV export.
- Keep the filtered-empty state accurate for amount/date-range/type filters.
- Add coverage for minimum amount, custom date range, `Last month`, `Other`, combined filters, reset, and export of filtered results.

## Target
Post-v1.0 / parity follow-up unless the final Activity design brings any of these filters into v1.0 scope.

## Split From
- #569

## Code References
- `qml/pages/wallet/Activity.qml`
- `qml/models/activityfilterproxymodel.cpp`
- `bitcoin/src/qt/transactionview.cpp:285`
- `bitcoin/src/qt/transactionview.cpp:303`

epicleafies:
Working on this.

Issue #706, opened 2026-05-17
Receive QR: add platform share action


johnny9:
## Gap
The receive request QR flow can copy the URI/address and save a QR image, but it does not expose a platform-level share action such as an OS share sheet.

Qt widgets do not provide this behavior; they support saving, copying, and dragging the QR image. This is useful on mobile and some desktop environments, but it is not required for v1.0 Qt parity.

## Work
Add a platform-appropriate share action for receive request information, likely sharing the BIP21 URI and/or QR image depending on platform support.

Keep copy URI, copy address, copy QR image, and save QR image behavior intact.

## Target
Post-v1.0.

## Split From
- #655

## Code References
- `qml/components/QRCodePopup.qml`
- `qml/pages/wallet/PaymentRequestDetail.qml`
- `bitcoin/src/qt/qrimagewidget.cpp:23`

PR #556, opened 2026-04-26
qml: Hide wallet change outputs in activity list

pseudoramdom:
This fixes an issue where the wallet activity list would show change outputs as separate outgoing & incoming transactions for normal sends.

Tests added:

  - Verifies sender change outputs are hidden.
  - Verifies incoming payments to change addresses are still shown.

Fixes https://github.com/bitcoin-core/gui-qml/issues/555

**Note** - I used an LLM agent to generate the unit tests. Looked fine to me at first glance.

PR #708, opened 2026-05-17
Activity filtering and Receive request

### LINKED ISSUE #457: BIP21 URI support in Receive page QR encoding
### Please describe the feature you'd like to see added.

The Receive page's generated QR code should follow the BIP21 URI spec.

https://bips.xyz/21

### Is your feature related to a problem, if so please describe it.

_No response_

### Describe the solution you'd like

_No response_

### Describe any alternatives you've considered

_No response_

### Please leave any additional context

_No response_

### LINKED ISSUE #518: Generation, History, QR, and BIP21 Sharing

### LINKED ISSUE #555: When sending from a wallet, the activity list shows 2 entries for change outputs
When sending from a wallet, the activity list shows the these rows
- A sent entry for the recipient output
- A sent entry for the wallet’s own change output
- A received entry for the wallet’s own change output

Image

#### Expected behavior
Only recipient outputs should be shown for outgoing transactions. Change outputs should not appear in the activity list.

### LINKED ISSUE #569: Activity filters: add designed search, date, type filters, and filtered export
## Code Gap
The QML Activity page needs the designed Activity filtering flow for the current release scope: text search, date presets, transaction type filters, reset behavior, filtered-empty state, and CSV export of the currently filtered rows.

This issue intentionally tracks the current QML design rather than full Qt widget transaction-table parity. Advanced parity filters are split to #707.

## Work Needed
- Add a search control for Activity rows.
- Match search against row label, address, and transaction id where available.
- Add date preset filters for all dates, today, this week, this month, and this year.
- Add type filters for all types, received, sent, sent to yourself, mined, and payment requests.
- Keep payment request rows filterable independently from fulfilled transaction rows.
- Reset filters when the search/filter UI is dismissed.
- Show a distinct filtered-empty state when rows exist but no row matches the active filters.
- Export the currently filtered Activity rows to CSV.
- Keep CSV export aligned with visible filter state and current display unit.
- Add C++ unit and GUI functional coverage for filtering, reset, filtered-empty state, and filtered export.

## Split Out
- #707 tracks minimum amount, custom date range, `Last month`, and `Other` type filters as advanced parity follow-up work.

## Code References
- `qml/models/activityfilterproxymodel.cpp`
- `test/test_activityfilterproxymodel.cpp`
- `test/functional/qml_test_activity_filter_export.py`

### LINKED ISSUE #632: Activity amounts: fix sent transaction sign display
## Code Gap
Sent transaction rows can be displayed with the wrong sign. `Transaction::fromWalletTx()` stores sent debits as negative values, while `Transaction::prettyAmount()` computes `credit - debit`. That can make sent Activity rows appear positive, while CSV export uses `credit + debit` and can disagree with the visible list.

## Work Needed
- Settle the Activity transaction amount sign convention for sent, received, self-transfer, generated, conflicted, and abandoned rows.
- Fix the visible amount formatting so sent transactions display as outgoing amounts.
- Keep CSV export and detail-page amounts consistent with the visible row.
- Add unit coverage for `Transaction::prettyAmount()` and `ActivityListModel::NetAmountSatRole` using sent and received transactions.
- Add GUI coverage for at least one sent row amount.

### LINKED ISSUE #641: Activity export: match display unit and release CSV behavior
## Code Gap
The QML Activity CSV export works, but it always writes an `Amount (BTC)` column and formats amounts in BTC. Qt exports transaction history using the current display unit in the amount column. This can make QML export disagree with what the user sees when the app is set to sats.

## Work Needed
- Use the current display unit in the Activity CSV amount column title and amount formatting.
- Keep exported values consistent with visible Activity row amounts after the amount-sign fix.
- Decide whether filtered pending payment requests should remain exportable or be excluded from transaction-history export.
- Preserve CSV quoting and failure handling.
- Add coverage for BTC and sats display units, filtered export, sent/received signs, and pending request export behavior.

### LINKED ISSUE #650: Receive addresses: handle address generation and unlock failures
## Gap
Receive address generation can fail silently, and `WalletQmlModel::newAddress()` dereferences `dest.value()` without checking the result. The receive page ignores `commitPaymentRequest()` failure and shows no recovery path.

## Work
Check address-generation results, propagate localized errors to QML, handle locked/encrypted wallet unlock when needed, and keep the UI from showing a created request when no address was created.

## Code References
- `qml/models/walletqmlmodel.cpp:507`
- `qml/models/walletqmlmodel.cpp:616`
- `qml/pages/wallet/RequestPayment.qml:296`
- `bitcoin/src/qt/receivecoinsdialog.cpp:173`

### LINKED ISSUE #651: Receive requests: show and copy full BIP21 URI after creation
## Gap
After creating a request, the immediate QML receive screen copies and QR-encodes the bare address. Amount, label, and message are dropped unless the user reaches the detail screen.

## Work
Use the full `bitcoin:` URI payload for the request QR and primary copy action after request creation. Keep an explicit copy-address action available for users who only need the address.

## Code References
- `qml/pages/wallet/RequestPayment.qml:284`
- `qml/pages/wallet/RequestPayment.qml:343`
- `qml/models/paymentrequest.cpp:138`
- `bitcoin/src/qt/receiverequestdialog.cpp:47`

### LINKED ISSUE #652: Receive requests: add address type selection
## Gap
QML always creates receive addresses using the wallet default output type. Qt widgets let the user choose supported address types such as Legacy, P2SH-SegWit, Bech32, and Taproot.

## Work
Expose supported receive output types in the request flow and pass the selected type into address generation for new requests and new receive addresses.

## Code References
- `qml/models/walletqmlmodel.cpp:627`
- `bitcoin/src/qt/receivecoinsdialog.cpp:89`
- `bitcoin/src/qt/receivecoinsdialog.cpp:154`

### LINKED ISSUE #653: Receive history: add visible requested-payment history
## Gap
QML has a receive request history model, but the Receive page only exposes a hidden automation count. Qt widgets show a visible requested-payments history table.

## Work
Add a visible requested-payment history surface under Receive, backed by `ReceiveRequestHistoryModel`, with enough detail to identify prior requests independently of Activity transaction status.

## Code References
- `qml/pages/wallet/RequestPayment.qml:29`
- `qml/models/receiverequesthistorymodel.cpp:95`
- `bitcoin/src/qt/receivecoinsdialog.cpp:81`

### LINKED ISSUE #654: Receive history: add show and remove actions for saved requests
## Gap
The backend can load and remove saved receive requests, but QML has no Receive-page actions to show an old request or delete it from history.

## Work
Wire show/open and remove actions from the receive history UI to `loadPaymentRequest()` and `removeReceiveRequest()`, including confirmation or undo behavior if the design requires it.

## Code References
- `qml/models/walletqmlmodel.cpp:684`
- `qml/models/walletqmlmodel.cpp:698`
- `bitcoin/src/qt/receivecoinsdialog.cpp:192`
- `bitcoin/src/qt/receivecoinsdialog.cpp:221`

### LINKED ISSUE #655: Receive QR: enable save image action
## Gap
Payment request detail can copy the URI/address and show the QR code, but the QR image needs a dedicated save-to-file action for Qt widget parity.

Qt widgets support saving the QR image as a PNG, plus copying/dragging the image from the QR widget. Platform-level share behavior is tracked separately in #706 as post-v1.0 work.

## Work
Add or keep a save/download action for the receive request QR image.

The saved image should encode the same BIP21 QR payload shown in the payment request detail flow. Keep copy URI and copy address behavior intact.

## Target
v1.0 Qt parity.

## Split Out
- #706 tracks platform share behavior for post-v1.0.

## Code References
- `qml/components/QRCodePopup.qml`
- `qml/pages/wallet/PaymentRequestDetail.qml`
- `bitcoin/src/qt/receiverequestdialog.cpp:49`
- `bitcoin/src/qt/qrimagewidget.cpp:117`

johnny9:
This PR consolidates the receive-request and Activity work into one branch so the receive history, payment request detail flow, Activity filtering/export, and transaction/request linking are handled together.

- Supersedes #553
- Supersedes #559
- Supersedes #556

- Adds wallet-backed receive request history with BIP21 URI/QR metadata, persistence, malformed-entry handling, detail pages, copy/download QR actions, use-as-template, and delete-from-history actions.
- Lets fulfilled payment requests stop being duplicate Activity rows while keeping matching request metadata reachable from the transaction details page.
- Adds Activity search, date/type filters, filtered-empty state, and filtered CSV export using the current display unit.
- Hides wallet change outputs from Activity and fixes sent transaction amount signs.
- Adds receive request address type selection and persisted Receive option toggles.
- Adds unit, QML, and functional coverage for receive requests, Activity filtering/export, display-unit export, saved request actions, and sent amount signs.

image

image

image

image

image

image

image

## Closes / Fixes

Closes #518
- Implements wallet-backed request generation/history, BIP21 QR payloads, persistence, selection/copy/remove, and malformed-entry coverage. Platform share is split out to #706.

Fixes #457
- Receive QR payloads now use the full BIP21 URI.

Fixes #555
- Activity hides wallet change outputs for outgoing transactions.

Closes #569
- Implements the currently designed Activity search, date/type filters, reset behavior, filtered-empty state, and filtered CSV export. Advanced date range, min amount, and Other filters are split out to #707.

Fixes #632
- Fixes sent transaction amount sign display and keeps Activity/export/detail amount conventions aligned.

Fixes #641
- Activity CSV export uses the current display unit in the amount header and exported amount formatting.

Fixes #650
- Handles receive address generation failures, locked encrypted wallet unlock prompts, localized errors, and prevents showing a created request when no address was generated.

Fixes #651
- The generated receive request exposes/copies the full BIP21 URI while keeping address copy available.

Fixes #652
- Adds receive request address type selection and passes the selected type into request address generation.

Fixes #653
- Adds visible requested-payment history.

Fixes #654
- Adds show/open, use-as-template, and delete-from-history actions for saved receive requests.

Fixes #655
- Adds QR image download/save-to-file support. Platform share remains split out to #706.

## Not closed by this PR

- #706: platform share action is intentionally post-v1.0.
- #707: advanced Activity filters are intentionally split out.


qml/models/receiverequestentry.h
```diff
@@ -0,0 +1,54 @@
...
+{
+    static constexpr int CURRENT_VERSION{2};
+    int nVersion{CURRENT_VERSION};
+    int64_t id{0};
+    QDateTime date;
+    QmlReceiveRequestRecipient recipient;
+
+    SERIALIZE_METHODS(QmlRecentRequestEntry, obj)
+    {
+        int64_t date_timet;
```
jarolrod, line 47: I know this PR is an early stage, but this being `int64_t` then being serialized and deserialized as such differs from how qt widgets does it, which it does `unsigned int` - the big difference here is writing and reading 8 bytes vs 4 bytes

This should mean that requests made in qt widgets can't really be read correctly in QML, so an existing GUI user moving to QML, may see their old requests disappear. 

We should make an effort to maintain cross compatibility. So, compatibility with Qt Widgets gui entries being read in QML. But also, compatibility (as much as possible) with QML entries begin read by Qt widgets (should a user ever need to move back or use another project that uses the qt widgets gui/or qt widgets gui style serialization and deser).

Worthwhile to add and maintain a cross-compat func test.

Also, something to keep in mind for future development.

johnny9: Yeah this is a problem with the wallet interface. This is a good PR to add the backwards compatibility check with serializing the requests. https://github.com/bitcoin/bitcoin/issues/34629 I raised the issue here but even the solution to the issue likely requires careful migration when adding new fields which I think we will want to add.

For the test itself, I can probably download a previous bitcoin-qt release and check that it doesn't crash at start up. I believe I can do it by just loading a wallet that has a receive request generated from our gui-qml and then launch v31 bitcoin-qt and it will attempt to read the receive requests on load.

qml/models/walletqmlmodel.cpp
```diff
@@ -443,73 +455,169 @@ QString WalletQmlModel::newAddress(QString label)
...
     if (!m_wallet || !m_current_payment_request) {
-        return;
+        return false;
     }
 
-    if (m_current_payment_request->id().isEmpty()) {
+    const bool is_update = !m_current_payment_request->id().isEmpty();
+
+    if (!is_update) {
         m_current_payment_request->setId(nextPaymentRequestId());
```
jarolrod: we set a new ID here, and so this is mutating m_current_payment_request and there is a non empty id. When qml goes to render it uses the `id !== ""` check to display a request.

This will happen even if `m_wallet->getNewDestination` below fails. 

Realistically, and if we build out the other wallet models correctly, this *should* be a mainly unreachable state, but should be noted. And is better overall, to just be defensive here -> so we should allow getNewDestination to succeed first, then setID -> so we can completely avoid this case.

qml/models/receiverequestentry.h
johnny9: Ill make an issue to track this compatibility test. I think that' a good one.

qml/res/icons/share.png
pseudoramdom, line 1: Can we start using SVG versions of all of these icons? 
They're all available at https://bitcoinicons.com

pseudoramdom:
Screenshot 2026-05-20 at 11 25 52 AM

The search icon is rendering bad on macOS

qml/components/CopiedToast.qml
```diff
@@ -0,0 +1,63 @@
...
+Rectangle {
+    id: root
+
+    property bool shown: false
+    property int visibleDurationMs: 1500
+    property string text: qsTr("Copied")
+    property color toastColor: Theme.color.green
+    property color contentColor: Theme.color.neutral0
+    property url iconSource: "image://images/check"
+
```
pseudoramdom: Oh we have a `Toast` component now that takes in icon, message, action.
We can maybe add the timer part to `Toast.qml` and get rid of this?

johnny9: Yeah I will remove this toast and I forgot I need to fix how this toast is working since it wasn't operating like a pop up. Will address this

qml/res/icons/share.png
johnny9: added the svg

qml/components/CopiedToast.qml
johnny9: I made a ToastBanner and ToastPopup

qml/pages/wallet/PaymentRequestDetail.qml
```diff
@@ -0,0 +1,478 @@
...
+    objectName: "paymentRequestDetailPage"
+    background: null
+
+    signal done()
+
+    property WalletQmlModel wallet: walletController.selectedWallet
+    property PaymentRequest request: wallet ? wallet.currentPaymentRequest : null
+    property bool detailsExpanded: false
+
+    function formatRelativeTime(isoString) {
```
pseudoramdom: This looks like it should live in cpp layer than qml. Thoughts?

pseudoramdom: Looks like we have a similar func in `Transaction::dateTimeString()` . Maybe move it to a dateutils ?

qml/pages/wallet/PaymentRequestDetail.qml
```diff
@@ -0,0 +1,478 @@
...
+                spacing: 4
+
+                CoreText {
+                    Layout.fillWidth: true
+                    visible: root.request !== null && root.request.amount.satoshi > 0
+                    text: {
+                        if (!root.request) return ""
+                        var label = root.request.amount.unitLabel
+                        if (root.request.amount.unit === BitcoinAmount.SAT && root.request.amount.satoshi !== 1)
+                            label = "sats"
```
pseudoramdom: Can we bake this in `BitcoinAmount`? 

```cpp
QString BitcoinAmount::unitLabel() const
  {
      switch (m_unit) {
      case Unit::BTC: return QStringLiteral("₿");  // glyph, no translation
      case Unit::SAT:
          return qAbs(m_satoshi) == 1
              ? tr("sat",  "unit label, singular")
              : tr("sats", "unit label, plural");
      }
      assert(false);

```

pseudoramdom: If we add a 
```cpp
 QString BitcoinAmount::displayWithUnit() const
  {
      return toDisplay() + QStringLiteral(" ") + unitLabel();
  }
```
this can be simplified to 
```cpp
 text: root.request ? qsTr("Requesting %1").arg(root.request.amount.displayWithUnit) : ""
```

test/qml/tst_requestpayment.qml
```diff
@@ -0,0 +1,151 @@
...
+import QtQuick.Controls 2.15
+import QtTest 1.2
+
+TestCase {
+    name: "RequestPayment"
+    when: windowShown
+    width: 400
+    height: 200
+
+    function formatRelativeTime(isoString) {
```
pseudoramdom: This is odd. Are we duplicating the `formatRelativeTime` logic to the test and testing the test version? The implementations could drift and we'd be left testing an incorrect implementation. We should probably remove this func.
The test should test the real thing :)

test/functional/qml_test_receive.py
```diff
@@ -0,0 +1,158 @@
...
+        _create_request(gui, "0.0001", "Alice", "pizza")
+        qr_code = gui.get_property("paymentRequestDetailQRCode", "code")
+        assert qr_code.startswith("bitcoin:"), f"QR payload missing BIP21 prefix: {qr_code!r}"
+        assert "amount=0.00010000" in qr_code, f"QR payload missing amount: {qr_code!r}"
+        assert "label=Alice" in qr_code, f"QR payload missing label: {qr_code!r}"
+        assert "message=pizza" in qr_code, f"QR payload missing message: {qr_code!r}"
+        print(f"[qml_receive_requests] created request with QR: {qr_code}")
+
+        _stop_gui(harness)
+        time.sleep(0.5)
```
pseudoramdom: Can we avoid `sleep` here?

qml/models/activitylistmodel.h
```diff
@@ -60,6 +61,8 @@ class ActivityListModel : public QAbstractListModel
     void updateTransaction(const uint256& hash, const interfaces::WalletTxStatus& wtx,
                            int num_blocks, int64_t block_time);
     int findTransactionIndex(const uint256& hash) const;
+    int findPendingRequestIndex(const QString& address) const;
+    void fulfillPendingRequest(int index, const QSharedPointer<Transaction>& real_tx);
     void removePendingRequestForAddress(const QString& address);
```
pseudoramdom: `removePendingRequestForAddress` does not have any callers. Remove?

qml/pages/wallet/RequestPayment.qml
```diff
@@ -9,30 +9,53 @@ import Qt.labs.settings 1.0
...
-    readonly property bool requestPersisted: root.requestReady && root.request.id !== ""
-    readonly property bool requestHasAddress: root.requestReady && root.request.address !== ""
-
-    function selectedReceiveAddressType() {
-        if (root.request && root.request.address === "" && root.request.addressType !== undefined && root.request.addressType.length > 0) {
-            return root.request.addressType
+    property string selectedReceiveAddressType: ""
+    property var availableAddressTypes: {
+        if (!root.wallet) return []
+        if (typeof root.wallet.availableReceiveAddressTypes === "function") {
```
pseudoramdom: This looks like some runtime check. Do we need this?

pseudoramdom: LLM says we can just do this? 
```cpp
property var availableAddressTypes: root.wallet ? root.wallet.receiveAddressTypes : []

function resetSelectedReceiveAddressType() {
    root.selectedReceiveAddressType = root.wallet ? root.wallet.defaultReceiveAddressType : "bech32"
}
```

test/qml/tst_requestpayment.qml
pseudoramdom: Looks like this was resolved in later commits.

qml/pages/wallet/PaymentRequestDetail.qml
pseudoramdom: Looks like this was resolved in later commits.

qml/bitcoin_qml.qrc
```diff
@@ -174,9 +182,11 @@
         <file alias="pending-light">res/icons/pending-light.png</file>
         <file alias="plus">res/icons/plus.png</file>
         <file alias="refresh">res/icons/refresh.png</file>
+        <file alias="qr-code">res/icons/qr-code.png</file>
         <file alias="plus-big-filled">res/icons/plus-big-filled.png</file>
         <file alias="pending">res/icons/pending.png</file>
         <file alias="search">res/icons/search.png</file>
+        <file alias="share">res/icons/share.png</file>
```
pseudoramdom, line 190: This should use the .svg?

qml/pages/wallet/ActivityDetails.qml
pseudoramdom: This seems to have drifted from the logic Activity.qml which has extra 
```cpp
if (delegate.isPendingRequest) {
    Theme.color.purple
}
```
Should we wrap this logic to common place?

qml/pages/wallet/ActivityDetails.qml
```diff
@@ -54,6 +56,32 @@ Page {
...
+        }
+        return qsTr("Payment request")
+    }
+
+    function paymentRequestSubtitle(request) {
+        var amount = request && request.amountDisplay && request.amountDisplay.length > 0
+            ? request.amountDisplay
+            : qsTr("No amount")
+        if (request && request.date && request.date.length > 0) {
+            return amount + " - " + request.date
```
pseudoramdom: This will break i8n (assuming date will change with locale)? 

```suggestion
            return qsTr("%1 - %2").arg(amount).arg(request.date)
```

pseudoramdom:
Screenshot 2026-05-21 at 1 32 43 PM
nit: Text color in the toast popup should probably be white. Also padding seems less on the right.

qml/pages/wallet/PaymentRequestDetail.qml
johnny9: yeah we should move as much as we can into cpp

qml/pages/wallet/ActivityDetails.qml
johnny9: yeah we can have a share function in a util.js. I'll think a bit about what can maybe go in cpp to simplify.

johnny9:
> Screenshot 2026-05-20 at 11 25 52 AM
> 
> The search icon is rendering bad on macOS

Fixed the button. It renders correctly on macOs
image

qml/pages/wallet/ActivityDetails.qml
johnny9: Added a common ActivityTransactionVisuals component that the pages both bind to

qml/bitcoin_qml.qrc
johnny9: going to stick with the embedded png for now. it is more efficient but we can reevaluate this icon flow.

johnny9:
> Screenshot 2026-05-21 at 1 32 43 PM nit: Text color in the toast popup should probably be white. Also padding seems less on the right.


https://github.com/user-attachments/assets/b07185d9-dfbd-4a47-8f94-41144a205184

johnny9:
I believe all comments have been addresses.

johnny9:
Update from [a4fd527](https://github.com/bitcoin-core/gui-qml/commit/a4fd5275f49f400c518f06a7ca01ae1a95e049a0) to [ad94733](https://github.com/bitcoin-core/gui-qml/commit/ad9473341242feca9389aa1f2c73142b33207f69)

- I dropped a change to dealing with IBD for determining if the wallet is syncing (you can see on regtest Activity doesn't know its not doing ibd so it thinks its syncing). This should be handled with a set of changes dealing with node events. #723 to track

pseudoramdom:
Reviewed till https://github.com/bitcoin-core/gui-qml/commit/a4fd5275f49f400c518f06a7ca01ae1a95e049a0 yesterday.
Followup commits look good. ACK ad9473341242feca9389aa1f2c73142b33207f69

pseudoramdom (approved):

qml/models/receiverequestentry.h
jarolrod: sorry if redundant, but did we make an issue in our repo to track this

qml/models/activitylistmodel.cpp
jarolrod, line 78: I don't think this is working as intended, look at the following set of actions

1. Open Receive.
2. Click Name field, set field to "Old Label", enter an amount
3. Click Generate payment request
4. Click Edit
5. Change Name to "New label"
6. Click Update payment request
7. Open Activity

And you will still see "Old Label"

Image

johnny9: Likely missing a signal. Will root cause and add an issue.

johnny9: #725

qml/models/activitylistmodel.cpp
```diff
@@ -203,6 +313,43 @@ int ActivityListModel::findTransactionIndex(const uint256& hash) const
...
+
+    for (int i = 0; i < m_transactions.size(); ++i) {
+        if (m_transactions[i]->isPendingRequest && m_transactions[i]->address == address) {
+            return i;
+        }
+    }
+    return -1;
+}
+
+void ActivityListModel::fulfillPendingRequest(int index, const QSharedPointer<Transaction>& real_tx)
```
jarolrod, line 328: There seems to be a big hole in the implementation of the Activity list around address reuse or duplicate requests. This removes the address from the pending lookup after fulfilling just one pending row, but core can store multiple receive requests for the same address. So we can still have another pending row for this address sitting in `m_transactions`.

After this line runs, `findPendingRequestIndex(address)` will return `-1` for future payments to that same address, even though a pending row still exists. The next payment gets inserted as a separate real transaction row, and the old pending request stays stuck as pending.

I think this needs either the same `address_still_pending` check used in `removePendingReceiveRequest()`, or maybe a clearer policy that same address pending requests are ambiguous and should all be removed or hidden once the address has a real transaction. 

Might be a bigger design/system thinking task to resolve this.

johnny9: I need to think about this but the idea was to have all payment requests listed under the address activity details is if there is a transaction for that address. If the filter is selected for payment requests on the activity list it should show all payment requests. I might need to double check that one.

johnny9: OK I think there might be a gap with what I expected. Likely need to raise an issue for showing payment request for used addresses in the activity list when payment request filter is selected. That might not be working right.

johnny9: #726

PR #716, opened 2026-05-19
Language and Units Functional Test Fix

### LINKED ISSUE #713: Functional test leaves app language set to Spanish after running
The language-switching functional test does not reset the user's language preference back to the system default after the test finishes. After running the test suite, the next launch of bitcoin-core-app starts in Spanish (the last language exercised by the test) instead of the system locale.
This causes a real-world annoyance for developers who run the functional tests against their personal datadir, and also makes test ordering subtly meaningful — any later test that assumes English strings can fail intermittently depending on what ran before it.
This is the same shape of leak as #714, on a different setting.

**Reproduction**
1. From a clean state, launch bitcoin-core-app — UI is in the system locale (English on most CI machines and dev environments).
2. Run the language functional test.
3. Quit the app and relaunch — UI is now in Spanish.

**Expected**
The test should restore the persisted language setting to whatever it was before the test ran (or, if the slot was empty, leave it empty so the next launch uses the system locale).

**Acceptance criteria**
1. After running the language functional test, bitcoin-core-app launches in the same language it would have launched in before the test ran.
2. Running the test in a loop does not progressively change the persisted language.
3. The teardown runs even if the test fails partway through (try/finally or fixture cleanup).

### LINKED ISSUE #714: Functional test leaves app display unit set to SAT after running
The display-unit functional test does not reset the user's display unit preference back to the default after the test finishes. After running the test suite, the next launch of bitcoin-core-app displays amounts in SAT (the unit selected by the test) instead of BTC.

This is the same shape of leak as #713, on a different setting.

**Reproduction**

1. From a clean state, launch bitcoin-core-app — amounts display in BTC.
2. Run the display-unit functional test.
3. Quit the app and relaunch — amounts now display in SAT.

**Expected**

The test should restore the persisted display unit to whatever it was before the test ran, so re-running the suite is idempotent and the user's preference is not silently changed.

**Acceptance criteria**

1. After running the functional test, bitcoin-core-app launches with the same display unit it would have launched with before the test ran.
2. Running the test in a loop does not progressively change the persisted display unit.
3. The cleanup runs even if the test fails partway through.

**Notes**

Relevant code: `test/functional/qml_test_settings_display.py`.

epicleafies:
Fixed language not resetting after functional tests. Fixes #713
Fixed units not resetting after functional tests. Fixes #714

### DIFF
--- a/test/functional/qml_test_settings_display.py
+++ b/test/functional/qml_test_settings_display.py
@@ -46,6 +46,20 @@ def navigate_to_display_settings(gui):
     print("  Navigated to Display settings page")
 
 
+def reset_display_unit_to_btc(gui):
+    """Reset the persisted display unit to BTC.
+
+    Precondition: caller is on the SettingsDisplay page with gotoDisplayUnit
+    visible. Callers should wrap invocations in `try/except QmlDriverError:
+    pass` for best-effort teardown that does not mask the original test failure.
+    """
+    gui.click("gotoDisplayUnit")
+    gui.wait_for_page("settingsDisplayUnitPage", timeout_ms=5000)
+    gui.click("displayUnitBTC")
+    gui.click("settingsDisplayUnitBack")
+    gui.wait_for_page("gotoDisplayUnit", timeout_ms=5000)
+
+
 def reset_language_to_system_default(gui):
     """Reset the persisted language to the System default (empty tag).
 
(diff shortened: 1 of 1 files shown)


johnny9 (commented):
Cleanup is good to have but the harness should set the config directory for the application into a unique temporary directory to isolate every test from eachother so if cleanup doesnt happen or something goes wrong in our script each test case can start clean.

johnny9:
The format for linking to an issue is "Fixes #<issue number>" or "Closes #<issue number>"

johnny9:
I cherry-picked your changes and rolled them into #724 as a part of its updates to the functional tests

PR #697, opened 2026-05-14
Have -disablewallet remove all walletmodels and validate node-only mode with a functional test

johnny9:
node-only mode using -disablewallet en espanol

image
image
image

Closes #506 
Closes #691

PR #558, opened 2026-04-26
Add MiniBlockClock to the Navigation Bar

johnny9:
Adds the mini version of the BlockClock dial. This mirrors the block clock states (Connecting, IBD, Blocks, Paused, and Error). 

Additionally, fixed a bug with the lambda capture for the update handler that was causing the IBD % to flicker between the current % and 0%. (83b445529de293a61a4aa429919e55454ce8557a)

Screenshot from 2026-04-26 01-38-26
Screenshot from 2026-04-26 01-37-48
Screenshot from 2026-04-26 01-37-32

closes #493

PR #724, opened 2026-05-23
Node status (mini blockclock) and runtime feedback (Errors/Warnings)

### LINKED ISSUE #324: App/block clock does not reflect network disconnection
The application & the block clock do not display an error state when network connection is lost.

Something like this? Not sure if it is implemented.

Screenshot 2023-04-18

**Actual behavior**
This one is on the build from #311 

Screenshot 2023-04-18

This one is from a older build.

Screenshot 2023-04-17 at 12 27 36

I've observed this more than once.

Repro: Turn off wifi when the app is running. I think it also occurred when the app is launched and the wifi is off.

Running macOS 12.5.6.

### LINKED ISSUE #493: block status estimation glitching
https://github.com/user-attachments/assets/cbdaffa1-471f-42f6-a646-88266daa4808

04b2a1ff5f924d248d8d2224d2e9ab99f27b9855

### LINKED ISSUE #506: Establish DoD Test Gates for Parity Work
## Summary
Create repeatable test scaffolding so every parity feature can satisfy DoD without ad-hoc test design.

## Complexity
- `L`

## Dependencies
- `01`

## Global Constraints

## Scope
- Define test matrix per feature type (C++ unit, QML QTest, Python E2E).
- Add missing testbridge commands/selectors needed for parity flows.
- Add CI-friendly deterministic test data/setup patterns for wallet + node scenarios.
- Document feature template for acceptance criteria and required tests.

1. A reusable issue/test template exists and is applied to all issues in this folder.
2. At least one representative feature from node, wallet, and send/receive domains has full DoD test triad passing.
3. Automation selectors (`objectName`) standards are documented and enforced in changed QML files.

## Required Tests
- C++ unit tests: Unit tests for any new automation support utilities.
- QML QTest: QTest validating critical selectors and state transitions are stable.
- Python E2E (`qml/test/testbridge.h`): One end-to-end pipeline run covering success and failure paths for representative flows.

## Implementation Notes
- This issue reduces risk and rework for every downstream parity issue.

## Done Criteria for This Issue
- [x] Create test target for unittesting and qml testing
- [x] Create python based end-to-end test scripts
- [ ] Representative Node/Wallet/SendReceive E2E scenarios exist and pass in CI.

### LINKED ISSUE #601: Runtime dialogs: bridge core message/question prompts to QML
## Code Gap
Qt widgets installs GUI handlers for runtime core prompts after the main window is created, including `uiInterface.ThreadSafeMessageBox` and `uiInterface.ThreadSafeQuestion`.

The QML app currently handles early startup errors through `InitErrorMessageBox()`, but there is no app-level QML dialog bridge for runtime node messages/questions once the main application is running. That leaves important node prompts without a release-level QML equivalent.

## Work Needed
- Add a QML-facing dialog service for runtime core message and question prompts.
- Route `ThreadSafeMessageBox` and `ThreadSafeQuestion` through the QML app after startup.
- Preserve blocking/non-blocking behavior, button mapping, severity, and fatal/shutdown semantics expected by the node interface.
- Ensure dialogs work before and after wallet UI is loaded.
- Cover the prompt paths with appropriate automated or documented manual coverage.

### LINKED ISSUE #602: Startup: surface fatal initialization failures in the app shell
## Code Gap
The QML app shell does not currently surface failures from the asynchronous startup path clearly.

Early startup failures before the main QML shell loads are already handled by `initerrormessage.qml` through the temporary `ThreadSafeMessageBox` hook. The gap is after the shell has loaded and `QmlInitExecutor::initialize()` runs `appInitMain()` on the init thread.

Today that path can leave the shell in an ambiguous state:
- wallet mode can keep showing loading/ghost UI without an actionable fatal message
- node-only mode can show the block clock error state, but without the actual startup failure text
- `runawayException` is not wired into `NodeModel`
- `QmlInitExecutor::handleRunawayException()` surfaces node warning text instead of the actual exception context

## Scope
Do not add a dedicated startup progress page or repurpose block sync status for startup progress.

The existing UI should remain the normal loading/status surface:
- wallet mode keeps using the existing ghost/loading pattern while initialization is in progress
- node-only mode keeps using the block clock for normal node/block status
- block-tip notifications and `verificationProgress` remain steady-state sync status, not the primary startup failure mechanism

The work here is to add a clear fatal startup failure path for async initialization errors.

## Work Needed
- Capture fatal init error text during the async `appInitMain()` path, including error-style message-box/init failure text.
- Expose a minimal startup failure contract from `NodeModel`, for example `startupError` plus the existing `faulted` state.
- On `initializeResult(false, ...)`, set `faulted` and expose the captured startup error text, with a generic fallback if no specific text was captured.
- Wire `QmlInitExecutor::runawayException` into `NodeModel` and surface the real exception context instead of `node.getWarnings()`.
- Show a shell-level fatal dialog/overlay when the async startup failure text is present.
- The fatal UI should include the actionable message text and a `Close` action that requests clean shutdown through `nodeModel.requestShutdown()`.
- Preserve the existing early initialization dialog for pre-shell failures.

## Coverage
Add coverage for:
- init failure after the QML shell has loaded exposes a startup error and sets `faulted`
- runaway startup exception exposes the actual exception text and sets `faulted`
- the shell fatal dialog appears when startup error text is present
- the close action from the fatal state requests shutdown
- normal wallet loading and node block-clock status continue to use the existing UI paths

### LINKED ISSUE #606: Connection status: distinguish inbound, outbound, and total peer counts
## Code Gap
`NodeModel` currently exposes `numOutboundPeers`, but the value is updated from the node connection-count notification, which reports total connections. QML then treats that total as an outbound count in places like the block clock and peer indicator.

That makes the exposed QML contract inaccurate:
- inbound peers can be counted as outbound peers
- total connection state is hidden behind an outbound-specific property name
- consumers cannot distinguish inbound-only, outbound-only, and mixed peer states
- UI code cannot clearly express whether it needs total connectivity or directional peer counts

## Scope
The focus of this issue is to clean up the peer-count properties exposed to QML and make them accurate.

This is not a request to design a new connection-status surface. Existing UI should continue to own presentation. The work should only update existing consumers where they are currently reading the wrong property.

## Proposed QML Contract
Expose explicit peer-count properties from `NodeModel`:
- `numPeers`: total connected peers / connections
- `numInboundPeers`: inbound connected peers
- `numOutboundPeers`: outbound connected peers
- `maxNumOutboundPeers`: outbound peer target/capacity used by existing outbound peer indicator UI

The source of truth should be the node interface counts by direction, matching the Qt client model pattern:
- inbound: `ConnectionDirection::In`
- outbound: `ConnectionDirection::Out`
- total: `ConnectionDirection::Both`

The existing connection-count notification can be used as the refresh trigger, but it should not be assigned directly to `numOutboundPeers`.

## Consumer Expectations
Update existing QML consumers to use the property that matches their intent:
- general connected/disconnected state should use `numPeers > 0`
- the block clock peer dots should remain outbound-peer related: `numOutboundPeers` scaled against `maxNumOutboundPeers`
- the peer dots should not count inbound peers, because their current design is tied to outbound peer capacity / target fill
- peer-count summaries should be able to show total, inbound, and outbound values where a peer monitoring surface already exists

An inbound-only node should be considered connected for general status, but the outbound peer dots should still show zero outbound peer capacity filled.

A zero-peer node should remain disconnected/connecting in existing status UI.

Paused network state should continue to be driven by `nodeModel.pause`, not by peer-count derivation.

## Work Needed
- Add accurate total, inbound, and outbound peer-count properties to `NodeModel`.
- Refresh all three counts from directional node counts when the connection-count notification fires.
- Initialize the counts from the node interface so the properties are accurate before the first notification.
- Keep `numOutboundPeers` semantically outbound-only after this change.
- Update existing QML reads so total connection state and directional peer counts are not conflated.
- Keep the block clock peer dots using outbound peers against `maxNumOutboundPeers`.
- Avoid broad UI redesign beyond correcting consumers of the peer-count properties.

## Test Validation
Add focused validation for the exposed property contract and its consumers:
- zero peers: total, inbound, and outbound are all zero; connected UI remains in the disconnected/connecting path
- outbound-only peers: total and outbound are nonzero; inbound is zero; general connected state is true; outbound peer dots are active
- inbound-only peers: total and inbound are nonzero; outbound is zero; general connected state is true; outbound peer dots remain empty/looking for outbound peers
- mixed peers: total equals inbound plus outbound from the node interface
- connection-count notifications refresh all directional counts instead of assigning the notification value directly to outbound
- block clock connected/disconnected state uses total peer count
- block clock peer dots use outbound count against `maxNumOutboundPeers`
- paused network state remains controlled by `pause`

### LINKED ISSUE #607: Node warnings: surface current warnings and alert changes in QML
## Code Gap
Qt widgets exposes current node warnings through `ClientModel::getStatusBarWarnings()`, which reads `interfaces::Node::getWarnings().translated`, and refreshes that status when `handleNotifyAlertChanged(...)` fires.

The QML `NodeModel` does not currently expose equivalent persistent node-warning state. As a result, QML cannot show warnings that are present at startup, warnings that appear later, warning text changes, or warning clears.

Runtime message/question dialogs are tracked separately. This issue is for ongoing node warning state that should remain visible while monitoring the node.

## Scope
The focus of this issue is to expose accurate current node-warning state to QML and connect it to the block clock warning states from the design.

Design reference: the Bitcoin Core App Design Figma node `4904-111548` shows two alert states for the block clock status. The intended UX for this issue is to replace the mini block-clock/status affordance with an alert icon while node warnings are present, show the warning string(s) under the main block clock, and show a condensed warning in the alert icon tooltip.

This is not a request to redesign the app shell, add a notification system, or change fatal startup handling. The implementation should follow the existing block clock/status surface and the referenced alert-state design.

Out of scope:
- wallet load/create/import warnings surfaced by `WalletQmlController`
- transient modal dialogs from `handleMessageBox(...)` / `handleQuestion(...)`
- fatal async startup failure handling
- debug-log parsing
- new warning categories or new Core warning sources
- broad navigation or app-shell redesign
- hand-maintained QML mappings from translated warning text to semantic warning labels

## Proposed QML Contract
Expose current node warnings from `NodeModel`, for example:
- `warnings`: current translated warning text from `interfaces::Node::getWarnings().translated`, empty when there are no warnings
- `hasWarnings`: convenience boolean derived from whether `warnings` is non-empty
- `warningSummary`: a short display string suitable for compact status/tooltip text, derived mechanically from the current warning text

The exact property names can follow local naming conventions, but QML should not need to call into Core directly or infer warning state from unrelated status fields.

The model should:
- initialize warning state from `node.getWarnings()` when `NodeModel` is created
- subscribe to `node.handleNotifyAlertChanged(...)`
- refresh the warning state from `node.getWarnings()` whenever the alert-change notification fires
- emit QML change notifications when warning text is added, changed, or cleared

## Warning Summary Rules
`warningSummary` should be a constrained preview of the actual warning state, not a separate semantic classification.

Suggested behavior:
- normalize warning text by stripping or splitting Core's joined-warning separator such as `<hr />`
- collapse whitespace/newlines for compact display
- for one warning, show a short truncated preview of that warning
- for multiple warnings, show a count-oriented summary such as `2 node warnings` where the compact status surface cannot fit all warning text
- use `warningSummary` for the alert icon tooltip
- keep the full warning text visible under the block clock for single and multiple warnings

Avoid parsing translated warning text into custom categories such as `Clock out of sync` unless Core later exposes stable warning IDs/categories to QML.

## Rendering Expectations
Use the block clock status design for node warnings:
- when `hasWarnings` is false, keep the existing block clock status behavior
- when `hasWarnings` is true, replace the mini block-clock/status affordance with the alert icon shown in the design
- hovering the alert icon should show a tooltip containing the condensed warning summary
- show the current full warning string(s) under the main block clock
- warning text should remain visible while Core reports warnings and disappear once Core reports no warnings
- if there are multiple warnings, show them in a readable form under the block clock; do not accidentally expose raw Core separators

Expected placement:
- wallet desktop mode: the block clock tab/status area should enter its warning state when node warnings are present
- node-only mode: the main block clock status should enter its warning state when node warnings are present
- warning state is app/node status, not network-label status, and should be visible on mainnet as well as non-mainnet networks

Warnings should remain non-blocking status. They should not be modeled as one-shot dismissible alerts. If the design adds click details, dismissing the detail view must not hide the warning while Core still reports it.

Warnings may be joined by Core with HTML-style separators such as `<hr />`. The tooltip should use the normalized/constrained `warningSummary`. The block clock warning text should either render joined warnings intentionally as rich text or convert them into readable plain text. Raw separators should not appear accidentally in the UI.

Example warning text that may be shown:
- `This is a pre-release test build - use at your own risk - do not use for mining or merchant applications`
- `Your computer's date and time appear to be more than ... minutes out of sync with the network...`
- `Warning: We do not appear to fully agree with our peers! You may need to upgrade, or other nodes may need to upgrade.`
- `Unknown new rules activated (versionbit X)`

## Work Needed
- Add current warning state properties to `NodeModel`.
- Add alert-change subscription wiring in `NodeModel`.
- Refresh warnings from `interfaces::Node::getWarnings()` on construction and on alert changes.
- Ensure warning clears propagate to QML, not just warning additions.
- Add mechanical warning-summary generation for compact warning display.
- Update the block clock status surface to enter the design's alert state when warnings are present.
- Replace the mini block-clock/status affordance with the alert icon while warnings are present.
- Add tooltip behavior on the alert icon using the condensed `warningSummary`.
- Show current full warning string(s) under the main block clock.
- Keep wallet warnings and runtime modal messages separate from node warning state.
- Avoid broad UI redesign beyond exposing and displaying the warning state in the block clock status surface.

## Test Validation
Add focused validation for the model contract and QML consumer behavior:
- warning present at `NodeModel` construction is exposed to QML
- no warning at construction exposes an empty warning state / false `hasWarnings`
- `handleNotifyAlertChanged` refreshes warning text from `getWarnings()`
- warning text changes emit the relevant QML change notification
- warning clear emits the relevant QML change notification and clears the block clock warning state
- repeated alert-change notifications with unchanged text do not emit unnecessary property changes
- single-warning summary is normalized and constrained for compact display
- multiple-warning summary uses a count-oriented label where compact display is needed
- alert icon tooltip shows the condensed warning summary
- full warning text remains available under the block clock for single and multiple warnings
- multiple warnings joined by Core are displayed readably, without accidental raw separator artifacts
- block clock status enters the alert state when warnings are present
- mini block-clock/status affordance is replaced by the alert icon while warnings are present
- block clock warning text appears when warnings are present and disappears when warnings clear

### LINKED ISSUE #608: Peer actions: handle disconnect, ban, and unban failures in QML
## Code Gap

The QML peer details page exposes Disconnect and Ban actions, and the banned-peers page exposes Unban. The C++ model methods return success/failure, but the QML handlers currently ignore those results.

Visible peer-control actions should not silently fail for Preview/Beta.

## Qt Behavior Snapshot

The existing Qt Widgets peer actions do not do extra recovery work on failure:

- `RPCConsole::disconnectSelectedNode()` calls `m_node.disconnectById(id)` and only clears the selected peer when it returns `true`. A failure is silent and leaves the current details/selection intact.
- `RPCConsole::banSelectedNode()` calls `m_node.ban(...)` and `m_node.disconnectByAddress(...)` without inspecting either return value. It clears the selected peer and refreshes the ban table regardless.
- `RPCConsole::unbanSelectedNode()` delegates to `BanTableModel::unban()`, which returns `false` for an invalid entry or backend failure. The RPC console refreshes the ban table only if at least one selected unban succeeds. Total failure is silent.

The backend interface only returns booleans for these operations, not detailed error strings. In practice, false can mean stale peer/list state, invalid input, no matching node, or unavailable `connman`/`banman`.

## Implementation Direction

- Stop ignoring peer-action return values in QML.
- Show a user-visible error dialog, or an existing equivalent error surface, when disconnect, ban, or unban returns `false`.
- Keep failure handling generic; do not invent backend-specific error reasons that the interface does not provide.
- Do not mark `NodeModel::faulted` for these action failures. They are action-level failures, not node initialization/runtime fault states.
- Do not add retry loops or additional backend recovery behavior. Refresh the affected peer/ban models after action attempts where useful so stale UI state can resolve.
- Treat successful actions as visible through normal UI state changes rather than adding success dialogs.

Suggested failure behavior:

- Disconnect failure: keep the peer details page open and show a generic error such as “Could not disconnect peer. The peer may already be disconnected or the node state may have changed.”
- Ban failure: validate the ban target and duration before submitting. If submission fails, show a generic error and do not silently close the flow in a way that looks successful.
- Unban failure: change `BanListModel::unbanAt(row)` to return `bool` so QML can show an error when the backend rejects the unban or the row is stale/invalid.

## Work Needed

- Use the existing `bool` results from `NodeModel::disconnectPeer()` and `NodeModel::banPeer()` in `PeerDetails.qml`.
- Change `BanListModel::unbanAt()` from `void` to `bool`, returning `false` for invalid rows or failed `m_node.unban(...)`.
- Add a small, reusable error dialog/surface if there is no existing suitable one.
- Keep the current peer-disappeared behavior coherent: if the selected peer disappears while the details page is open, leave the details page or disable actions rather than letting actions target stale data silently.
- Refresh affected peer and ban-list state after disconnect, ban, and unban attempts where the model can be stale.

## Acceptance Criteria

- Failed Disconnect is visible to the user and does not look like success.
- Failed Ban is visible to the user and does not look like success.
- Failed Unban is visible to the user and does not remove the row as though it succeeded.
- Invalid or stale ban targets/durations are rejected before or at submission with visible feedback.
- Successful disconnect, ban, and unban still update the peer details/list or ban list as they do today.
- No new behavior depends on parsing backend error text; the implementation uses the boolean success/failure API.

## Suggested Coverage

- C++ unit coverage for `NodeModel::disconnectPeer()` false, `NodeModel::banPeer()` false, and invalid ban address/duration behavior.
- C++ unit coverage for `BanListModel::unbanAt()` success, invalid row false, and backend false.
- QML coverage that the Disconnect, Ban, and Unban handlers show the failure surface when their model call returns `false`.
- Existing functional peer-management tests can remain focused on happy-path regtest behavior unless a reliable failure scenario is added separately.

### LINKED ISSUE #609: Block status: add detailed sync state, header sync, and ETA diagnostics
## Code Gap
The QML block clock shows basic verification progress and an estimated remaining time, but it does not expose the detailed Qt sync diagnostics from the status bar/modal overlay.

Missing release-level details include header sync/presync state, last block time, blocks left, progress per hour, clearer ETA behavior, and a detailed view users can open while the node is syncing.

## Work Needed
- Bridge header tip updates and distinguish header presync, header sync, block sync, disk indexing, and peer-connection states where the node provides them.
- Expose last block time, progress percentage, progress per hour, estimated time left, and blocks/headers left where available.
- Add a QML detail surface for block/sync status rather than relying only on the compact block clock.
- Fix ETA update behavior so it is not stale or effectively stuck at estimating during normal sync.
- Add coverage for initial block download, header sync, synced, paused, and no-peer states.

### LINKED ISSUE #610: Node information: add Qt debug-window information diagnostics
## Code Gap
Qt widgets exposes a debug-window Information tab with node and network diagnostics such as client version/user agent, data directory, blocks directory, startup time, network name, local addresses, connection counts, block count, last block time, and warnings.

The QML app has separate About, Storage, Peers, Network Traffic, Mempool, Debug Log, and Console surfaces, but does not yet provide an equivalent consolidated node information diagnostic surface.

## Work Needed
- Add a QML node information diagnostic page or section.
- Include the release-scoped information fields from the Qt debug-window Information tab.
- Include data directory and blocks directory values, not just the configured data directory label.
- Include startup time, network name, connection split, block count, last block time, and current warnings where available.
- Add copy behavior for diagnostic fields if selected for v1.0.
- Add coverage for node-only and wallet-enabled modes.

### LINKED ISSUE #691: Honor runtime `-disablewallet` in app routing so onboarding exits to node-only (`nodeStack`) and validate with `test/functional/qml_test_disablewallet_boot.py`.
add a -disablewallet argument to stop the wallet models from being added to the qml engine. If that is not enough, include a github workflow that builds without wallet and run the node functional tests with it.

### LINKED ISSUE #710: Hide or clarify Mempool Information when running blocksonly
## Problem
The Settings screen shows a `Mempool Information` entry even when the node is running with `-blocksonly`.

Today the row is unconditional in `qml/pages/node/NodeSettings.qml`, and `NodeModel` always exposes and polls mempool values through `refreshMempoolInfo()` / `fetchMempoolInfo()`. There is no blocksonly or transaction-relay gate in QML or `NodeModel`.

This can make the settings UI look like regular mempool behavior is available/configurable in blocksonly mode, even though transaction relay is disabled and Core applies blocksonly-specific mempool behavior.

## Expected behavior
When running with `-blocksonly`, the GUI should either:

- hide the `Mempool Information` settings entry, or
- clearly present the page as blocksonly-limited behavior, if we still want to expose the configured max mempool limit.

## Notes

### LINKED ISSUE #713: Functional test leaves app language set to Spanish after running

### LINKED ISSUE #714: Functional test leaves app display unit set to SAT after running

### LINKED ISSUE #723: Activity empty state should not infer syncing from verification progress
## Problem

The QML Activity empty state currently uses chain verification progress as a proxy for whether wallet activity is still syncing. That can show `Syncing wallet activity...` when the node is already past initial sync but `verificationProgress` is still below the UI threshold.

For an empty wallet that is not actively syncing, this makes the Activity page look like data is still catching up instead of showing the normal empty-wallet copy.

## Expected behavior

Activity empty-state copy should be based on an explicit node sync/IBD state, not a `verificationProgress < 0.9999` threshold.

When the wallet has no activity and the node is not syncing, Activity should show the normal empty state:

- `No activity yet`
- `Once you send or receive bitcoin, your transactions will appear here.`

When the wallet has no activity and the node is still syncing, Activity can continue to show the syncing-specific copy.

## Work Needed

- Expose a clear sync state from `NodeModel` if one is not already available.
- Update `Activity.qml` to use that sync state for the source-empty Activity copy.
- Keep filtered-empty behavior separate from source-empty behavior.
- Keep this compatible with the broader Activity sync-warning work tracked in #634.
- Add coverage for:
  - empty wallet while node is syncing
  - empty wallet while node is not syncing but verification progress is below the old threshold
  - normal non-empty Activity behavior

## Code References

- `qml/models/nodemodel.h`
- `qml/models/nodemodel.cpp`
- `test/qml/tst_activity.qml`
- `test/test_nodemodel.cpp`

## Related

- #634 tracks the broader out-of-sync warning/banner work around balances and Activity.

johnny9:
This PR supersedes:
- #558
- #697
- #716

- Adds the mini BlockClock/navigation indicator and fixes queued node-status updates so block progress does not flicker from stale callback data.
- Surfaces runtime node message/question prompts, startup fatal errors, node warnings, and node information diagnostics in QML.
- Adds directional peer counts, peer-action failure dialogs, and blocksonly-aware mempool settings visibility.
- Honors runtime `-disablewallet` by keeping wallet models out of the QML engine and routing onboarding to node-only mode.
- Fixes Activity empty-state sync copy to use explicit block-sync state instead of verification-progress thresholds.
- Makes the display-settings functional tests restore and sandbox persisted QSettings state.
- Adds unit, QML, and functional coverage for node status, runtime feedback, BlockClock, disablewallet, blocksonly, peer actions, Activity empty states, and settings persistence.

01-startup-warning-runtime-dialog
02-node-only-actions-row
03-node-warnings-popup
04-node-information-popup
image
image

## Closes / Fixes

Fixes #324
- BlockClock/node status now reflects disconnected/no-peer states through explicit node status inputs and mini BlockClock coverage.

Fixes #493
- Queues node model block-tip and connection-count updates by value so IBD progress does not flicker with stale callback data.

Closes #506
- Adds representative node, wallet, and send/receive functional coverage to the GUI functional workflow, alongside unit/QML coverage for the new status surfaces.

Closes #601
- Routes runtime `ThreadSafeMessageBox` and question prompts through QML runtime dialogs after startup, preserving blocking answers for questions.

Closes #602
- Captures async initialization failures and runaway startup exceptions in `NodeModel`, then shows a shell-level fatal error popup with shutdown handling.

Fixes #606
- Exposes total, inbound, and outbound peer counts separately and updates status consumers so total connectivity is not confused with outbound capacity.

Closes #607
- Exposes node warning state in QML, refreshes it from alert changes, and adds warning summary/detail surfaces.

Fixes #608
- Shows user-visible failures for disconnect, ban, and unban actions instead of silently ignoring failed peer operations.

Closes #609
- Adds header-sync/block-sync state and sync diagnostics through the BlockClock status and node information surfaces without adding a dedicated sync modal.

Closes #610
- Adds a QML node information popup with release-scoped diagnostics such as version, directories, network, peers, block state, and warnings.

Closes #691
- Honors runtime `-disablewallet` by omitting wallet context objects/models and validating node-only onboarding/settings flow with functional coverage.

Fixes #710
- Hides the mempool information settings entry when running in blocksonly mode and covers the behavior with QML/functional tests.

Fixes #713
- Restores the language setting after the display-settings functional test and runs the test in a temporary QSettings sandbox.

Fixes #714
- Restores the display unit after the display-settings functional test and runs the test in a temporary QSettings sandbox.

Fixes #723
- Activity source-empty copy now follows explicit block-sync state, so a fresh empty wallet no longer shows syncing copy solely because verification progress is below the old threshold.


qml/components/MiniBlockClock.qml
```diff
@@ -0,0 +1,186 @@
...
+    property real iconSize: 18
+    property bool pageSelected: false
+    property bool connected: nodeModel.numPeers > 0
+    property bool synced: nodeModel.verificationProgress > 0.999
+    property bool paused: nodeModel.pause
+    property bool faulted: nodeModel.faulted
+    property var networkStatusModelRef: typeof networkStatusModel !== "undefined" ? networkStatusModel : null
+    property bool offline: networkStatusModelRef !== null && networkStatusModelRef.networkOffline
+
+    readonly property bool showOfflineState: !root.paused && !root.faulted && root.offline
```
jarolrod: This is a behavior change from the big block clock because here paused has higher precedence than offline, while the main BlockClock and tooltip show offline ahead of paused. 

If both `nodeModel.pause` and `networkStatusModel.networkOffline` are true, the main clock is in the `Offline` state but the mini will show Paused.

Should evaluate what is our correct order of operations for these states

qml/components/NodeInformationPopup.qml
```diff
@@ -0,0 +1,121 @@
...
+    objectName: "nodeInformationPopup"
+
+    property var rows: []
+    readonly property int informationRowCount: informationRepeater.count
+    readonly property string firstInformationValue: rows.length > 0 ? rows[0].value : ""
+    readonly property int lastInformationValueLineCount: informationRepeater.count > 0 ? informationRepeater.itemAt(informationRepeater.count - 1).valueLineCount : 0
+    readonly property int lastInformationValueWrapMode: informationRepeater.count > 0 ? informationRepeater.itemAt(informationRepeater.count - 1).valueWrapMode : Text.NoWrap
+    readonly property int contentMargin: 28
+
+    onAboutToShow: rows = nodeModel.nodeInformationRows()
```
jarolrod, line 22: while unlikely to be actually user reachable in practice, this code is still fragile like so and can lead to a crash:

This popup can be opened while node startup is still running. In that window, `nodeInformationRows()` can call `getHeaderTip()`, `getNumBlocks()`, and `getLastBlockTime()` before `ChainstateManager` has been created, and those paths assert on `chainman()`.

qml/models/nodemodel.cpp
```diff
@@ -5,35 +5,119 @@
...
+    if (style & CClientUIInterface::ICON_ERROR) {
+        return QStringLiteral("image://images/error");
+    }
+    if (style & CClientUIInterface::ICON_WARNING) {
+        return QStringLiteral("image://images/alert-filled");
+    }
+    return QStringLiteral("image://images/info-filled");
+}
+
+QString RuntimeDialogPrimaryText(unsigned int style, bool question)
```
jarolrod: any reason why we only handle a subset of the CClientUIInterface button mask sets, the qt widgets gui handles all of them

jarolrod: to expand on the above

qt widgets just passes `style & CClientUIInterface::BTN_MASK` directly to `QMessageBox`, so it supports the full Core button set. 

Since here we are reimplementing that contract, I think it should either map all `BTN_*` values here or explicitly reject/test unsupported combinations. Right now `Ignore`, `Help`, and `Reset` silently fall through.

johnny9: I went back and reorganized around this so that it follows the node interface ui buttons instead of abstracting it away. It should now have the correct buttons by using DialogButtonBox.

qml/components/NodeInformationPopup.qml
johnny9: added an initialization flag to prevent this and a regression test as well

qml/components/MiniBlockClock.qml
johnny9: fixed

pseudoramdom:
The buttons look enlarged on macOS 
Screenshot 2026-05-26 at 2 25 57 PM

johnny9:
> The buttons look enlarged on macOS 

Forced the style and added implicit width/height to the icons. you might need to `brew install qtdeclarative` to add the basic qml styles.

image

qml/components/blockclockdial.cpp
```diff
@@ -43,6 +47,19 @@ void BlockClockDial::setupConnectingGradient(const QPen & pen)
     m_connecting_gradient.setColorAt(1, "transparent");
 }
 
+void BlockClockDial::setupSyncedGradient(const QRectF& bounds)
+{
```
pseudoramdom, line 51: This is called on every paint() ? Not sure if it's a performance hit, but If so, can we update them only when there are updates.

i.e. exit early here `if (!m_synced_gradient_needs_update) return`

And set `m_synced_gradient_needs_update = true` in `setPenWidth`, `setConfirmationColors` etc.

We can also revisit perf later.

test/functional/qml_test_disablewallet_boot.py
```diff
@@ -0,0 +1,207 @@
...
+        "desktopWalletSettingsTabButton",
+    } & objects
+    assert not unexpected, (
+        "Wallet UI should not be instantiated in -disablewallet mode: "
+        f"{sorted(unexpected)}"
+    )
+
+    wallet_settings_visible = gui.get_property("settingsWallet", "visible")
+    assert wallet_settings_visible is False, (
+        "External Signer settings row should be hidden in -disablewallet mode, "
```
pseudoramdom: nit: Should be Wallet settings and not External signer

qml/components/ActionErrorPopup.qml
```diff
@@ -0,0 +1,71 @@
...
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+
+import "../controls"
+
+Popup {
+    id: root
+
+    property string title: qsTr("Action failed")
```
pseudoramdom: This is fine for now. Instead of creating one-off alerts, I'd love for us to create generic Alert popups that can take actions, mark an action as destructive (button is red for actions like ban). 

```
AlertPopup {
      id: deleteAlert
      title: qsTr("Delete wallet?")
      message: qsTr("Are you sure you want to delete this wallet?")

      AlertAction {
          text: qsTr("Cancel")
          role: AlertAction.Cancel
      }
      AlertAction {
          text: qsTr("Delete")
          role: AlertAction.Destructive
          onTriggered: walletController.removeWallet(walletId)
      }
  }
```

qml/components/ActionErrorPopup.qml
```diff
@@ -0,0 +1,71 @@
...
+        id: columnLayout
+        anchors.fill: parent
+        spacing: 0
+
+        CoreText {
+            Layout.fillWidth: true
+            Layout.preferredHeight: 56
+            text: root.title
+            bold: true
+            font.pixelSize: 20
```
pseudoramdom:
```
font: Theme.text.subtitle.font
lineHeight: Theme.text.subtitle.lineHeight
```

qml/components/ActionErrorPopup.qml
```diff
@@ -0,0 +1,71 @@
...
+            Layout.fillWidth: true
+        }
+
+        CoreText {
+            objectName: "actionErrorMessage"
+            Layout.fillWidth: true
+            Layout.margins: 20
+            text: root.message
+            color: Theme.color.neutral8
+            font.pixelSize: 15
```
pseudoramdom:
```
font: Theme.text.description.font
lineHeight: Theme.text.description.lineHeight
```

qml/models/nodemodel.cpp
```diff
@@ -473,9 +483,10 @@ void NodeModel::ConnectToBlockTipSignal()
 
     m_handler_notify_block_tip = m_node.handleNotifyBlockTip(
         [this]([[maybe_unused]] SynchronizationState state, interfaces::BlockTip tip, double verification_progress) {
-            QMetaObject::invokeMethod(this, [this, block_height = tip.block_height, block_time = tip.block_time, verification_progress] {
+            QMetaObject::invokeMethod(this, [this, state, block_height = tip.block_height, block_time = tip.block_time, verification_progress] {
                 setBlockTipHeight(block_height);
                 setVerificationProgress(verification_progress);
+                setBlockSyncActive(block_height > 0 && state != SynchronizationState::POST_INIT);
```
pseudoramdom: This is intentionally different from line 447 where we use `isInitialBlockDownload` instead of `SynchronizationState::POST_INIT`?

johnny9: yeah, in the case of isInitialBlockDownload, it is a initial state query to the node so we know where we are starting at. the state comes from the handler so we get accurate updates with that without having to do another query using the node interface call.

johnny9:
abort-retry-ignore-button-mask
database-read-error
deprecated-checkpoints-warning
reindex-question-ok-abort

Here are examples of the runtimeDialog messages from the node's interface_ui hooks

qml/components/AlertPopup.qml
```diff
@@ -0,0 +1,125 @@
...
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import QtQml 2.15
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+
+import "../controls"
+
+Popup {
```
pseudoramdom, line 12: This is great!

pseudoramdom:
Looking good!
ACK 0a622e80a70302ca8d927b6e0946213ff8101cb1

pseudoramdom (approved):

qml/models/nodemodel.cpp
jarolrod, line 119: Not a blocker for this PR, but we need to add this here, can include here or add on a follow-up pr, or make an issue
```suggestion
    unsubscribeFromCoreSignals();
    setMempoolInfoPollingActive(false);
```

jarolrod:
ACK https://github.com/bitcoin-core/gui-qml/commit/0a622e80a70302ca8d927b6e0946213ff8101cb1

qml/models/nodemodel.cpp
johnny9: #730

PR #727, opened 2026-05-26
Watch-only wallet, wallet loading status

### LINKED ISSUE #571: Create wallet: validate wallet name availability before continuing

### LINKED ISSUE #573: Create wallet password: show mismatch feedback before enabling Continue
## Code Gap
The create-wallet password page disables Continue until the password fields match, but it does not tell the user why the action is unavailable.

This is not a primary release blocker, but the Beta flow should make failed password confirmation understandable without requiring trial and error.

## Work Needed
- Show inline feedback when both password fields are non-empty and do not match.
- Keep the existing confirmation acknowledgement requirement.
- Ensure the error clears immediately when the passwords match.
- Add coverage for mismatch, match, empty fields, and acknowledgement toggle behavior.

### LINKED ISSUE #574: Close wallet: add confirmation, warning text, and failure handling
## Code Gap
The wallet selector now has a close control for loaded wallets, but the action closes immediately. Qt asks for confirmation and warns that closing a wallet for too long can require a resync when pruning is enabled.

The QML close path also does not surface a failure state if the selected wallet disappears or cannot be closed cleanly.

## Work Needed
- Add a confirmation dialog before closing a wallet from the selector.
- Include pruning/resync warning text when appropriate for the selected release behavior.
- Surface close failures or stale-wallet cases instead of silently returning.
- Keep wallet selection coherent after closing the selected wallet.
- Add coverage for close current wallet, close non-current loaded wallet, cancel, stale wallet, and final wallet closed.

### LINKED ISSUE #616: Create wallet: move standard wallet creation off the QML thread with progress state
## Code Gap
External-signer creation, wallet import, wallet load, and wallet migration run through the wallet controller worker path. Standard single-sig wallet creation currently calls `walletLoader().createWallet()` synchronously from `WalletQmlController::createSingleSigWallet()`.

This can block the QML/UI thread during wallet creation and does not provide the same progress/busy behavior used by the other wallet lifecycle flows.

## Work Needed
- Move standard and encrypted wallet creation onto the wallet controller worker thread.
- Add a busy/progress state to the create password page for both Skip and Continue paths.
- Disable duplicate submissions while creation is in progress.
- Preserve warning and error reporting from wallet creation.
- Ensure the created wallet becomes selected after async completion.
- Add coverage for unencrypted creation, encrypted creation, creation failure, warning reporting, and duplicate-click prevention.

### LINKED ISSUE #617: Watch-only wallet creation: add no-private-keys wallet flow
## Code Gap
The Qt create-wallet dialog supports creating a wallet with private keys disabled. The QML Add Wallet intro copy advertises view-only wallets, and imported watch-only wallets can be described, but there is no generic QML flow to create a no-private-keys/watch-only wallet.

This is a primary wallet type for release. External-signer wallet creation is not a substitute because it requires a detected signer and sets the external signer flag.

## Work Needed
- Add a wallet type choice for no-private-keys/watch-only wallet creation.
- Create the wallet with descriptor and disable-private-keys flags without requiring an external signer.
- Disable incompatible encryption/password choices for no-private-keys wallets, matching the selected Qt behavior.
- Make the created wallet's settings and send/receive affordances reflect watch-only state.
- Coordinate with Send work for unsigned/PSBT creation from watch-only wallets.
- Add coverage for successful watch-only creation, duplicate name, settings metadata, password/encryption gating, and first-use routing.

### LINKED ISSUE #618: Wallet open: show load progress, warnings, and errors from wallet selection
## Code Gap
Selecting a closed wallet from the wallet selector can trigger an async wallet load. The controller tracks `walletLoadInProgress`, `walletLoadWarnings`, and `walletLoadError`, but the selector/app-shell path does not present a clear loading, warning, or failure state for normal wallet-open attempts.

The popup closes immediately, so an open failure can be easy to miss or appear as no action.

## Work Needed
- Show loading state when opening a closed wallet from the selector.
- Surface load warnings and errors in the selector or app shell.
- Keep selection state coherent if loading fails.
- Preserve migration-required routing for legacy wallets.
- Add coverage for successful open, failed open, warning-only open, already-open selection, and migration-required open.

### LINKED ISSUE #621: Wallet metadata: complete key-scheme and private-key state display
## Code Gap
Wallet settings expose basic key-scheme/private-key/external-signer text, but the current model only distinguishes watch-only from single-key for `keyScheme()`. Imported multi-key wallets, blank wallets, external-signer wallets, and no-private-keys wallets need clearer release-level metadata.

This also affects which wallet settings/actions should be offered for wallets that cannot hold private keys.

## Work Needed
- Expand wallet metadata detection for single-key, multi-key, watch-only/no-private-keys, blank, and external-signer wallets.
- Display the selected release metadata consistently in Wallet settings and import success surfaces.
- Gate password/security settings for no-private-keys or other wallet types where Qt disables them.
- Add coverage for each wallet type retained for v1.0.

### LINKED ISSUE #699: Wallet selector: long wallet names are cut off in badges and list rows
## Summary
Long wallet names should remain identifiable in both the selected wallet badge and the wallet selector list.

Image

Parent: #577

## Problem
`WalletBadge` clips its content and uses non-wrapping wallet name text, while `WalletSelect.qml` reuses the badge as a fixed-width, fixed-height list delegate. With long regtest or test wallet names, the visible text is cut off in both the active wallet badge and the wallet list, making similarly named wallets hard to distinguish.

## Work Needed
- Define a consistent wallet-name overflow strategy for `WalletBadge` and `WalletSelect.qml` rows.
- Keep the selected-wallet badge and selector rows visually stable while still exposing enough of the wallet name to identify it.
- Consider middle/right elision plus a way to see the full wallet name on desktop, such as tooltip text, where that matches existing UI patterns.
- Verify long names do not overlap icons, balance text, row borders, or adjacent controls.

## Acceptance Criteria
1. A long wallet name remains identifiable in the selected wallet badge.
2. A long wallet name remains identifiable in each wallet selector row.
3. Badge/list layout stays stable with long names, including checked and hover states.
4. Test coverage or a documented manual test uses multiple long, similar wallet names.

### LINKED ISSUE #700: Wallet selector: improve scrolling for long wallet lists
## Summary
The wallet selector should remain fast and fully usable when the wallet directory contains a long list of wallets.

Image

Parent: #577

## Problem
With many regtest wallets, the wallet selector list scrolls slowly and the long list can be cut off. `WalletSelect.qml` currently uses a clipped popup, a fixed-width column, a `ListView` with a height derived from `Math.min(listView.count * 34, 300)`, and a trailing `AddWalletButton`. The selector needs to handle large wallet counts without hiding rows or making the add-wallet action unreachable.

## Work Needed
- Make the wallet selector size itself against the available viewport/window space instead of relying only on a fixed 300px list cap.
- Ensure the `ListView`, scrollbar, popup clipping, and `AddWalletButton` cooperate so the bottom of a long list is not cut off.
- Improve long-list scrolling responsiveness; consider delegate reuse, cache behavior, and avoiding unnecessary layout churn during scroll.
- Validate behavior with a large number of wallets, including enough rows to require scrolling.

## Acceptance Criteria
1. A long wallet list scrolls smoothly enough to select any wallet without laggy interaction.
2. The list is not visually cut off at the bottom of the popup/window.
3. The `Add Wallet` action remains visible or reachable when the wallet list is long.
4. Test coverage or a documented manual test exercises a long wallet list, for example 50+ wallets.

### LINKED ISSUE #711: Validate wallet display names to avoid duplicate or misleading wallet labels
## Problem
Wallet display-name editing currently allows duplicate names and names that conflict with other wallet names/display names.

`WalletQmlController::setWalletDisplayName()` only trims the requested alias and stores or removes it in `QSettings`. It does not check whether the display name duplicates another loaded/listed wallet name or another stored display name.

Because wallet selector rows use the display-name role from `WalletListModel`, two distinct wallets can appear with the same visible label. This can make wallet selection and wallet settings ambiguous, especially after renaming one wallet to match another wallet's name or alias.

## Expected behavior
The GUI should prevent or clearly reject wallet display names that would make wallets indistinguishable in user-facing wallet lists.

Possible approaches:
- Reject display names that match another wallet's real name.
- Reject display names that match another wallet's stored display name.
- Show an inline validation error in Wallet Settings when the requested display name is empty/duplicate/conflicting as appropriate.

## Notes

Relevant code:
- `qml/walletqmlcontroller.cpp`: `setWalletDisplayName()`
- `qml/models/walletlistmodel.cpp`: wallet selector display-name role

## Release Target
Preview

pseudoramdom:
This PR consolidates watch-only wallet creation and tracks per-wallet load states. It also bundles in small improvements to wallet picker view.

- Watch-only wallet - Supersedes https://github.com/bitcoin-core/gui-qml/pull/565 fixes #617 
- Other wallet creation improvements
  -  Move wallet creation off UI thread and show loading indicator (fixes #616)
    - <details> <summary>Screenshot</summary>Screenshot 2026-05-26 at 11 33 20 AM</details> 
  - Validate wallet name availability before creation - supersedes #686 which fixes #571, fixes #711
  - Show inline password mismatch error in wallet creation flow (fixes #573)
  - <details> <summary>Screenshot</summary> Screenshot 2026-05-26 at 9 53 47 AM</details>  

Wallet picker improvements 
- Introduce per-wallet loading state in the wallet picker
- Each wallet entry in the picker will show the icon reflecting the wallet key scheme and load status. It shows the balance if loaded. (fixes #621 )
- <details> <summary>Screenshot</summary>Screenshot 2026-05-26 at 11 40 35 AM</details> 
- Show an error if a wallet failed to load (fixes #618)
- <details><summary>Screenshot</summary>Screenshot 2026-05-13 at 1 40 37 PM</details> 
- Show confirmation popup before closing wallet (fixes #574)
- <details><summary>Screenshot</summary>Screenshot 2026-05-26 at 9 52 35 AM</details> 
- Truncate long wallet names and show tooltip on hover and handle long wallets list (fixes #699, fixes #700)
- <details><summary>Screenshot</summary>Screenshot 2026-05-26 at 11 12 58 AM</details>


.github/workflows/gui-functional-tests.yml
```diff
@@ -112,3 +112,4 @@ jobs:
           python3 test/functional/qml_test_wallet_migration.py
           python3 test/functional/qml_test_wallet_rbf.py
           python3 test/functional/qml_test_wallet_restore.py
+          python3 test/functional/qml_test_watchonly_wallet.py
```
johnny9, line 120: add create_wallet_name functional test as well as the wallet selector functional test to the workflow

qml/controls/CoreTextField.qml
```diff
@@ -9,6 +9,7 @@ TextField {
     id: root
 
     property bool hideText: false
+    property bool showText: false
```
johnny9: this property seems redundant

qml/walletqmlcontroller.cpp
```diff
@@ -389,6 +435,83 @@ bool WalletQmlController::createExternalSignerWallet(const QString& name)
...
+        raw_wallet->ConnectScriptPubKeyManNotifiers();
+    }
+
+    if (descriptors_added == 0) {
+        setWalletLoadError(tr("Failed to import descriptors into watch-only wallet."));
+        return;
+    }
+
+    QMutexLocker locker(&m_wallets_mutex);
+    m_selected_wallet = new WalletQmlModel(std::move(*wallet_result));
```
johnny9: we should let the wallet loader notification handler create the wallet model. doing it this way creates a duplicate model in the list and closing the wallet will have issues.

johnny9: something like

1. In createWatchOnlyWallet(), validate name/xpub, set loading state, and mark the expected wallet name as “defer adoption”.
2. Let createWallet() fire NotifyWalletLoaded().
3. In handleLoadWallet(), if the loaded wallet name matches the deferred watch-only create, stash the interfaces::Wallet and return without registering a model.
4. Import both descriptors into the returned wallet.
5. If descriptor import fails, unload/remove the just-created wallet and report create error.
6. If descriptor import succeeds, call handleLoadWallet(std::move(deferred_wallet)) once, with m_pending_wallet_load_action = WalletLoadAction::Create, so the normal path registers the model and emits walletCreateSucceeded().

qml/pages/wallet/CreateWalletWizard.qml
```diff
@@ -18,9 +18,31 @@ PageStack {
...
+    }
+
+    Connections {
+        target: walletController
+        enabled: root.waitingForInit
+        function onInitializedChanged() {
+            if (walletController.initialized) {
+                root.waitingForInit = false
+                walletController.createWatchOnlyWallet(root.walletName, root.xpub)
+                if (walletController.isWalletLoaded) {
```
johnny9: this should probably be based on a different condition. If another wallet is already loaded and watch-only creation fails at qml/walletqmlcontroller.cpp (line 456) or qml/walletqmlcontroller.cpp (line 502), the wizard can still push the success screen.

johnny9:
```
property bool creatingWatchOnly: false

Connections {
    target: walletController

    function onWalletCreateSucceeded() {
        if (!root.creatingWatchOnly) return
        root.creatingWatchOnly = false
        root.push(watchOnlyConfirm)
    }

    function onWalletCreateErrorChanged() {
        if (root.creatingWatchOnly && walletController.walletCreateError.length > 0) {
            root.creatingWatchOnly = false
        }
    }

    function onWalletLoadErrorChanged() {
        if (root.creatingWatchOnly && walletController.walletLoadError.length > 0) {
            root.creatingWatchOnly = false
        }
    }
}
```

qml/walletqmlcontroller.cpp
```diff
@@ -389,6 +435,83 @@ bool WalletQmlController::createExternalSignerWallet(const QString& name)
...
+            std::string error;
+            auto parsed = Parse(info.desc_str, keys, error, /*require_checksum=*/false);
+            if (parsed.empty()) {
+                continue;
+            }
+
+            wallet::WalletDescriptor w_desc(
+                std::move(parsed.at(0)),
+                TicksSinceEpoch<std::chrono::seconds>(Now<NodeSeconds>()),
+                /*range_start=*/0,
```
johnny9, line 578: We should fill in these values for the descriptor. This can be a followup issue as it likely requires designing the wizard to include these choices and maybe a date picker

qml/walletqmlcontroller.cpp
```diff
@@ -486,6 +615,51 @@ bool WalletQmlController::walletPathExists(const QString& path) const
     return !normalized.isEmpty() && QFileInfo::exists(normalized);
 }
 
+bool WalletQmlController::walletNameExists(const QString& name) const
+{
+    const QString candidate = QDir::cleanPath(name);
+    if (candidate.isEmpty()) {
```
johnny9, line 707: i think you need to trim the string as 

QDir::cleanPath("   ")      -> "   "
QDir::cleanPath("  foo  ")  -> "  foo  "
QDir::cleanPath("./foo//bar") -> "foo/bar"
QDir::cleanPath("foo/../bar") -> "bar"

cleanPath can still get " " through this condition

johnny9 (commented):
Be sure to update unittest to cover needed changes.

qml/pages/wallet/CreateIntro.qml
```diff
@@ -102,7 +102,7 @@ Page {
             Layout.leftMargin: 20
             Layout.rightMargin: 20
             Layout.alignment: Qt.AlignCenter
-            text: qsTr("Start")
+            text: qsTr("Continue")
```
johnny9, line 105: The CreateIntro page can be removed now

qml/pages/wallet/CreateTypeSelector.qml
```diff
@@ -0,0 +1,108 @@
...
+
+        ColumnLayout {
+            id: columnLayout
+            width: Math.min(parent.width, 450)
+            anchors.horizontalCenter: parent.horizontalCenter
+            spacing: 10
+
+            Item { Layout.preferredHeight: 10 }
+
+            WalletTypeListItem {
```
johnny9, line 63: We should include external signer walets in this list if it is configured

qml/controls/CoreTextField.qml
pseudoramdom: That's embarrassing I didn't check the line above

johnny9: we're moving fast, its going to happen and we're going to find more once things settle

qml/pages/wallet/CreateIntro.qml
pseudoramdom: You mean this page is to be removed? Sure? I see it on [Figma](https://www.figma.com/design/ek8w3n3upbluw5UL2lGhRx/Bitcoin-Core-App-Design?node-id=7705-669020&t=H1DSZVkpr6RiFiNT-0) and the [core-app website](https://bitcoincore.app/milestones/1-4-create/)
Screenshot 2026-05-28 at 12 53 54 PM

qml/pages/wallet/CreateTypeSelector.qml
pseudoramdom: Added in [48d0575](https://github.com/bitcoin-core/gui-qml/pull/727/commits/48d057518da4d53cdf692a00c03b4d241f80f33c)

qml/walletqmlcontroller.cpp
pseudoramdom: This took a bit more time than I anticipated. Updated now.

qml/walletqmlcontroller.cpp
pseudoramdom: https://github.com/bitcoin-core/gui-qml/issues/735

qml/pages/wallet/CreateIntro.qml
johnny9: I meant the create wallet wizard page should just be the create type selector page

qml/walletqmlcontroller.cpp
```diff
@@ -389,6 +436,103 @@ bool WalletQmlController::createExternalSignerWallet(const QString& name)
...
+    const auto reset_deferred = qScopeGuard([this] {
+        m_deferred_wallet_name.clear();
+        m_deferred_wallet.reset();
+    });
+
+    const uint64_t creation_flags = wallet::WALLET_FLAG_DISABLE_PRIVATE_KEYS |
+                                    wallet::WALLET_FLAG_DESCRIPTORS |
+                                    wallet::WALLET_FLAG_BLANK_WALLET;
+
+    auto wallet_result = m_node.walletLoader().createWallet(
```
jarolrod: this is running synchronously on the qml controller thread.

The updated defer logic handles the model duplication problem, which is good. But the heavy wallet work is still happening inline here.

- `WalletLoader::createWallet()` calls into `CreateWallet()`, which can create or open wallet db storage, initialize wallet state, emit the load notification, add the wallet to the wallet context, run init processing, and update wallet settings.
- `AddWalletDescriptor()` persists descriptors and tops up descriptor script caches. Even if this is normally fast, it is still filesystem and wallet setup work being run from a QML invoked method.

That differs from the rest of the controller methods we have implemented:
- `createSingleSigWallet()`, `createExternalSignerWallet()`, `startWalletLoad()`, and `startWalletImport()` all post the wallet operation to `m_worker` and then marshal back to the controller thread for QML state updates. 

I think the correct model we should maintain is wallet & db work runs _outside_ of the UI thread; model adoption and signal emission stay on the QML thread.

qml/walletqmlcontroller.cpp
```diff
@@ -744,6 +940,14 @@ void WalletQmlController::startWalletImport(const QString& path)
...
 {
+    // Multi-step create flows (currently watch-only) stash the wallet here
+    // and finish setup before re-invoking with the defer-name cleared.
+    if (wallet && !m_deferred_wallet_name.isEmpty() &&
+        QString::fromStdString(wallet->getWalletName()) == m_deferred_wallet_name) {
+        m_deferred_wallet = std::move(wallet);
+        return;
+    }
+
     const WalletLoadAction load_action = m_pending_wallet_load_action;
```
jarolrod: This part intrigued me, so I dug into whether the pending wallet operation is actually tied to the wallet that just loaded. I think there is still a very real lifecycle bug here; that might be worthwhile to address in this PR.

The QML controller currently tracks “some wallet operation is pending” with `m_wallet_load_requested` and `m_pending_wallet_load_action`, but it does not track the target wallet name/path for that pending operation. `handleLoadWallet()` is registered as the global Core wallet-loaded callback, so it can be called for any wallet load notification, not only the wallet caused by the current QML action.

That matters because this function reads `m_pending_wallet_load_action`, clears it, and later emits `walletCreateSucceeded()`, `walletLoadSucceeded()`, or `walletImportSucceeded()` without first proving that the loaded wallet is the pending target. So an unrelated wallet notification can consume the active pending operation.

Comparing with Qt Widgets, it avoids this specific issue by keeping global wallet adoption separate from operation completion:

- `WalletController::handleLoadWallet()` only calls `getOrCreateWallet()` to adopt the loaded wallet model.
- `CreateWalletActivity`, `OpenWalletActivity`, and `RestoreWalletActivity` complete from the `WalletLoader::createWallet/loadWallet/restoreWallet` result returned by their own worker operation.
- The activity emits `created()` / `opened()` / `restored()` only if that specific operation set its own `m_wallet_model`.

So in Widgets, an unrelated wallet loaded notification may adopt another wallet model globally, but it does not complete the active create or open activity. In QML, `handleLoadWallet()` currently does **both** adoption and pending operation completion, which makes the global notification able to satisfy the wrong operation.

I wrote two local unit tests to check this:

- start creating `created_wallet`, then deliver a load notification for `unrelated_wallet`;
- start loading `target_wallet`, then deliver a load notification for `unrelated_wallet`.

Both fail on current HEAD. The unrelated notification emits the pending success signal:

```text
FAIL! initializedControllerDoesNotCompleteCreateForUnrelatedLoadNotification()
Actual create_spy.count(): 1
Expected: 0

FAIL! initializedControllerDoesNotCompleteLoadForUnrelatedLoadNotification()
Actual load_spy.count(): 1
Expected: 0
```

Here are the tests I wrote 

<details>

```diff
commit 0e7040ce0680ee7f88b4960489723ba4b6814008
Author: jarolrod <jarolrod@tutanota.com>
Date:   Fri May 29 16:17:05 2026 -0400

    test

diff --git a/test/test_walletqmlcontroller.cpp b/test/test_walletqmlcontroller.cpp
index 4917ccddc..50938302e 100644
--- a/test/test_walletqmlcontroller.cpp
+++ b/test/test_walletqmlcontroller.cpp
@@ -64,6 +64,10 @@ public:
         create_wallet_fn = [](const std::string&, const SecureString&, uint64_t, std::vector<bilingual_str>&) {
             return util::Error{Untranslated("Unexpected createWallet call")};
         };
+    std::function<util::Result<std::unique_ptr<interfaces::Wallet>>(const std::string&, std::vector<bilingual_str>&)>
+        load_wallet_fn = [](const std::string&, std::vector<bilingual_str>&) {
+            return util::Error{Untranslated("Unexpected loadWallet call")};
+        };
     std::function<util::Result<interfaces::WalletMigrationResult>(const std::string&, const SecureString&)>
         migrate_wallet_fn = [](const std::string&, const SecureString&) {
             return util::Error{Untranslated("Unexpected migrateWallet call")};
@@ -93,10 +97,10 @@ public:
         ++create_wallet_calls;
         return create_wallet_fn(name, passphrase, wallet_creation_flags, warnings);
     }
-    util::Result<std::unique_ptr<interfaces::Wallet>> loadWallet(const std::string&, std::vector<bilingual_str>&) override
+    util::Result<std::unique_ptr<interfaces::Wallet>> loadWallet(const std::string& name, std::vector<bilingual_str>& warnings) override
     {
         ++load_wallet_calls;
-        return util::Error{Untranslated("Unexpected loadWallet call")};
+        return load_wallet_fn(name, warnings);
     }
     std::string getWalletDir() override { return {}; }
     util::Result<std::unique_ptr<interfaces::Wallet>> restoreWallet(const fs::path&, const std::string&, std::vector<bilingual_str>&) override
@@ -294,6 +298,8 @@ private Q_SLOTS:
     void initializedControllerForwardsMigrationPassphrase();
     void initializedControllerForwardsUtf8CreatePassphrase();
     void initializedControllerEmitsWalletCreateSucceeded();
+    void initializedControllerDoesNotCompleteCreateForUnrelatedLoadNotification();
+    void initializedControllerDoesNotCompleteLoadForUnrelatedLoadNotification();
     void initializedControllerReportsCreateWarnings();
     void initializedControllerIgnoresDuplicateCreateWhileInProgress();
     void initializedControllerRequestsPassphraseBeforeEncryptedMigration();
@@ -672,6 +678,78 @@ void WalletQmlControllerTests::initializedControllerEmitsWalletCreateSucceeded()
     QCOMPARE(controller.walletCreateError(), QString{});
 }
 
+void WalletQmlControllerTests::initializedControllerDoesNotCompleteCreateForUnrelatedLoadNotification()
+{
+    using ::testing::StrictMock;
+
+    StrictMock<MockNode> node;
+    FakeWalletLoader loader;
+    ExpectControllerInitialization(node, loader);
+
+    WalletQmlController controller(node);
+    controller.initialize();
+    QVERIFY(loader.captured_load_wallet_fn);
+
+    FakeWallet::State created_state;
+    FakeWallet::State unrelated_state;
+    loader.create_wallet_fn = [&](const std::string& name,
+                                  const SecureString&,
+                                  uint64_t,
+                                  std::vector<bilingual_str>&) {
+        return util::Result<std::unique_ptr<interfaces::Wallet>>{
+            std::make_unique<FakeWallet>(name, &created_state)};
+    };
+
+    QSignalSpy create_spy(&controller, &WalletQmlController::walletCreateSucceeded);
+
+    controller.createSingleSigWallet("created_wallet", "");
+    QVERIFY(controller.walletLoadInProgress());
+
+    loader.captured_load_wallet_fn(std::make_unique<FakeWallet>("unrelated_wallet", &unrelated_state));
+
+    QCOMPARE(create_spy.count(), 0);
+    QVERIFY(controller.walletLoadInProgress());
+
+    loader.captured_load_wallet_fn(std::make_unique<FakeWallet>("created_wallet", &created_state));
+    QTRY_COMPARE_WITH_TIMEOUT(create_spy.count(), 1, 5000);
+    QVERIFY(!controller.walletLoadInProgress());
+}
+
+void WalletQmlControllerTests::initializedControllerDoesNotCompleteLoadForUnrelatedLoadNotification()
+{
+    using ::testing::StrictMock;
+
+    StrictMock<MockNode> node;
+    FakeWalletLoader loader;
+    ExpectControllerInitialization(node, loader);
+
+    WalletQmlController controller(node);
+    controller.initialize();
+    QVERIFY(loader.captured_load_wallet_fn);
+
+    FakeWallet::State loaded_state;
+    FakeWallet::State unrelated_state;
+    loader.load_wallet_fn = [&](const std::string& name,
+                                std::vector<bilingual_str>&) {
+        return util::Result<std::unique_ptr<interfaces::Wallet>>{
+            std::make_unique<FakeWallet>(name, &loaded_state)};
+    };
+
+    QSignalSpy load_spy(&controller, &WalletQmlController::walletLoadSucceeded);
+
+    controller.setSelectedWallet("target_wallet", "sqlite");
+    QVERIFY(controller.walletLoadInProgress());
+
+    loader.captured_load_wallet_fn(std::make_unique<FakeWallet>("unrelated_wallet", &unrelated_state));
+
+    QCOMPARE(load_spy.count(), 0);
+    QVERIFY(controller.walletLoadInProgress());
+
+    loader.captured_load_wallet_fn(std::make_unique<FakeWallet>("target_wallet", &loaded_state));
+    QTRY_COMPARE_WITH_TIMEOUT(load_spy.count(), 1, 5000);
+    QVERIFY(!controller.walletLoadInProgress());
+}
+
 void WalletQmlControllerTests::initializedControllerReportsCreateWarnings()
 {
     using ::testing::StrictMock;
```

</details>

qml/walletqmlcontroller.cpp
```diff
@@ -389,6 +436,103 @@ bool WalletQmlController::createExternalSignerWallet(const QString& name)
     return true;
 }
 
+void WalletQmlController::createWatchOnlyWallet(const QString &name, const QString &xpub)
+{
+    clearWalletLoadStatus();
```
jarolrod, line 527: Right now, it seems to me, this path can run before `initialize()`, and it can also run while another wallet load or create is already in progress. In the second case, `clearWalletLoadStatus()` clears the existing in-progress state, then the function continues into `WalletLoader::createWallet()`.

Maybe we add the same basic guards used by `createSingleSigWallet()` here: return if `m_wallet_load_in_progress` is true, and return with the “wallets still loading” error if `!m_initialized`?

Overall, I think we need more unit tests around all these load and create functions

johnny9: we've had other instances of calls that need to be protected with an initialization check

qml/walletqmlcontroller.cpp
pseudoramdom: Good catch. I’m going to rework this so watch-only creation follows the same threading model as the other wallet operations. Validation will stay on the QML thread, but `createWallet()` and descriptor import can run on `m_worker`

Probably create a helper method
```cpp
void createWalletAsync(
         QString wallet_name,
         SecureString passphrase,
         uint64_t flags,
         WalletLoadAction action,
         std::function<util::Result<void>(interfaces::Wallet&)> setup = {});
```

qml/walletqmlcontroller.cpp
johnny9: I think that means the fix should be:

- handleLoadWallet() should only addOrSelectWalletModel(...).
- createSingleSigWallet(), createWatchOnlyWallet(), startWalletLoad(), and startWalletImport() should emit success from their own worker result after addOrSelectWalletModel(...) returns a model.
- Watch-only needs one extra guard versus Qt widgets: because descriptor import happens after createWallet(), the NotifyWalletLoaded callback must defer that wallet before publishing it, otherwise QML can briefly expose a half-created wallet.

pseudoramdom: thanks both for the suggestion! I reworked this along the same lines as Qt Widgets. The global `handleLoadWallet()` path now only adopts/selects the wallet, and each create/load/import operation emits success from its own worker.
I also removed the pending `m_wallet_load_requested` / `m_pending_wallet_load_action` states.

pseudoramdom:
[317e42f](https://github.com/bitcoin-core/gui-qml/commit/317e42fc071b0cfe077e32c9830206227e7adcf0)
Rebased on top of `qt6` branch and added a commit to repurpose AlertPopup

qml/walletqmlcontroller.cpp
```diff
@@ -277,60 +297,196 @@ void WalletQmlController::removeWalletModel(WalletQmlModel* wallet_model)
...
+            });
+            return;
+        }
+
+        QString setup_error;
+        if (setup_wallet) {
+            setup_error = setup_wallet(**wallet);
+        }
+        if (!setup_error.isEmpty()) {
+            (*wallet)->remove();
```
jarolrod, line 414: Are we intentionally wanting to leave a newly created and blank wallet on disk here? even though the real wallet creation work failed?

I can see why `remove()` would avoid deleting wallets in general. But in this flow the UI reports watch-only creation as failed, so then retrying with the same name can be blocked because the wallet directory exists.

johnny9: I think we might need to define a flow to add keys to a wallet in the wallet settings. If we have this, we can re-order the watch-only onboarding to create the wallet first and then treat the xpub as adding keys to it. This better matches the actual interface and lets us have the create wallet stage be separate from the xpub failure so we don't have a left over wallet

johnny9: My suggestion is to leave these as the follow follow-up issues.

1. Watch-only wallet creation leaves a blank wallet on disc if adding keys fails
2. Create a path in wallet settings to add xpubs to a watch only wallet

Build 2 and then use that to close 1 by making it ok that the xpub path isn't filled in during onboard

jarolrod:
ACK a8527f13c67cf9fbee0e6b6da2b8af1bc32068bd

I agree this is ready to merge, we have documented several follow-ups to be done.

qml/walletqmlcontroller.cpp
johnny9: #737

johnny9 (approved):
ACK a8527f13c67cf9fbee0e6b6da2b8af1bc32068bd

Issue #582, opened 2026-05-06
Epic: Send Bitcoin


johnny9:
## Goal
Track v1.0 work for sending funds, fees, RBF, BIP21 URI intake, coin selection, watch-only send behavior, and PSBT operations.

## Release Scope
Preview fixes correctness bugs in existing send behavior: duplicate/dust validation and stale selected-coin state.

Beta covers primary release blockers for Send: watch-only unsigned PSBT creation, usable PSBT operations, RBF control, send metadata persistence, and the first documentation pass.

v1.0 adds release-scoped secondary send parity such as use-available-balance and final docs. Lower-priority PSBT breadth, advanced coin control, and expanded fee targets can follow after v1.0.

## Child Work
- #659 Send validation: reject duplicate recipients and dust outputs before review
- #660 Send metadata: persist labels and BIP21 message/order data
- #661 Send RBF: add opt-in replace-by-fee control and review status
- #662 Watch-only send: create unsigned PSBT instead of direct send
- #663 PSBT operations: add inspect, sign, broadcast, copy, and save screen
- #664 PSBT import: support partial, multisig, and non-wallet PSBT workflows
- #665 Coin control: clear selected coins after send and clear form
- #666 Coin control: add custom change address and detailed selection stats
- #667 Send amount: add use available balance action
- #668 Send fees: expand and persist fee target options
- #669 Docs: write Beta Send Bitcoin user guide
- #670 Docs: write final Send Bitcoin guide with screenshots

## Related Work
- #603 tracks desktop startup, single-instance, and file-open routing for payment URIs.
- External signer fallback behavior should be coordinated with #583.

## User Flows
- Send bitcoin to one recipient.
- Send bitcoin to multiple recipients.
- Validate address, amount, dust, and duplicate-recipient states.
- Use labels and BIP21 message/order metadata.
- Choose recommended or custom fees.
- Use subtract-fee-from-amount and send-available-balance behavior where supported.
- Enable or understand initial-send RBF.
- Select coins and understand selected-coin exclusivity.
- Send from a watch-only wallet by creating, saving, or copying an unsigned PSBT.
- Create, load, inspect, sign, broadcast, copy, and save PSBTs.
- Open a BIP21 payment URI from the app or desktop integration.

## Definition of Done
- Basic send, validation, fees, RBF, watch-only unsigned send, and PSBT flows required for v1.0 are implemented and tested.
- Watch-only unsigned PSBT save/copy is complete for release.
[...]

Issue #579, opened 2026-05-06
Epic: View Balances and Activity


johnny9:
## Goal
Track v1.0 work for wallet balances, overview, activity history, transaction details, export, and transaction recovery actions.

## Release Target Policy
- **Preview**: visible balances and Activity rows are not misleading. Amount signs and basic transaction display must be correct, and visible actions must not be dead ends. No documentation requirement.
- **Beta**: primary Activity flows are coherent, including transaction list/detail, sync warnings, speed-up behavior for supported wallets, scalable history loading, and baseline documentation.
- **v1.0**: final balance breakdown, privacy/mask, filter, label-editing, and recovery-action parity decisions are complete, final screenshots are captured, and any lower-priority Qt widget parity is explicitly deferred.
- **Post-v1.0**: mobile-specific Activity polish or lower-priority transaction-table parity that is useful but not release-blocking.

## User Flows
- View the current wallet balance and, where selected for release, available, pending, immature, total, and watch-only balances.
- Understand wallet sync and out-of-sync warnings before relying on balance or activity data.
- Use privacy/mask values behavior where selected for release.
- Browse recent activity and full transaction history.
- Filter transactions by date, type, search text, and amount where supported.
- Export transaction history.
- Open transaction details.
- Copy transaction information.
- Edit labels from activity where supported.
- Speed up an RBF-enabled transaction from Activity.
- Abandon or recover transactions where supported.

## Definition of Done
- Each child issue identifies a concrete implementation, documentation, or parity gap in the current QML app.
- Each child issue has a clear target: Preview, Beta, v1.0, or Post-v1.0.
- Preview-visible balance and activity amounts are correct and not misleading.
- The Beta user guide has a top-level `View Balances and Activity` page.
- The v1.0 guide includes prerequisites, steps, screenshots from the QML app, expected results, and important failure/recovery states for the final flow set.
- Remaining Qt widget parity gaps are tracked as child issues or intentionally deferred.

Issue #577, opened 2026-05-06
Epic: Create and Manage Wallets


johnny9:
## Goal
Track v1.0 work for wallet creation and wallet lifecycle management.

## Release Target Policy
- **Preview**: standard/encrypted wallet creation, wallet selection, wallet opening, and visible close controls are safe and coherent. Creation/open/close actions should not block the UI or fail silently. No documentation requirement.
- **Beta**: primary wallet lifecycle flows are complete, including watch-only/no-private-keys wallet creation, reliable wallet-open states, and baseline documentation.
- **v1.0**: final wallet-type and lifecycle parity decisions are made, final screenshots are captured, and any lower-priority Qt widget parity is explicitly deferred.
- **Post-v1.0**: optional wallet creation polish that is useful but not release-blocking.

## User Flows
- Create a standard wallet.
- Create an encrypted wallet.
- Create a no-private-keys/watch-only wallet.
- Create a blank wallet if retained for v1.0 scope.
- Create an external-signer wallet when external signer support is enabled.
- Open an existing wallet.
- Switch between loaded wallets.
- Close one wallet and, if retained for v1.0 scope, close all wallets.
- Understand the no-wallet empty state.

## Definition of Done
- Each child issue identifies a concrete implementation, documentation, or parity gap in the current QML app.
- Each child issue has a clear target: Preview, Beta, v1.0, or Post-v1.0.
- Preview-visible wallet creation, open, select, and close flows do not block the UI or silently lose failure states.
- Watch-only/no-private-keys wallet creation is complete enough for release.
- The Beta user guide has a top-level `Create and Manage Wallets` page.
- The v1.0 guide includes prerequisites, steps, screenshots from the QML app, expected results, and important failure/recovery states for the final flow set.
- Remaining Qt widget parity gaps are tracked as child issues or intentionally deferred.

PR #736, opened 2026-05-29
A bunch of fixes for Preview release target

### LINKED ISSUE #467: Amount validation in Send form is not taking into account the fee
### Report

Note that when entering full balance, `Review` is highlighted but not clickable.

As the GUI won't allow creating tx's with 0 fees, maybe `Review` should only be highlighted when amount < balance

Originally posted by @MarnixCroes in https://github.com/bitcoin-core/gui-qml/pull/462#pullrequestreview-2898268466

This requires dynamic fee estimation to be added to the Send form in order to calculate.

### LINKED ISSUE #624: Create backup page: make View file action functional
## Code Gap
The post-create backup page shows a `View file` action, but the QML button is not wired to any behavior. This leaves a visible dead control in the wallet creation backup flow.

## Work Needed
- Wire the `View file` action to open the created backup file or its containing folder using the selected desktop behavior.
- Show a user-visible failure state if the file/location cannot be opened.
- If opening the file is not desirable for release, replace the action with a clear supported action instead of leaving it dead.
- Keep the post-create backup page route forward coherent after backup creation.
- Add coverage for successful open, failed open, and missing backup path cases.

### LINKED ISSUE #636: Activity speed-up: unlock encrypted wallets before fee bump confirmation
## Code Gap
The Activity speed-up overlay can prepare and confirm a fee bump, but `SpeedUpOverlay.qml` still has a FIXME to unlock the wallet before confirming. `BumpTransactionModel::confirmFeeBump()` signs the replacement transaction directly, so encrypted locked wallets can fail the visible speed-up flow instead of prompting for the passphrase.

## Work Needed
- Detect when the selected wallet is encrypted and locked before fee-bump confirmation.
- Prompt for the wallet passphrase using the shared wallet passphrase UI.
- Confirm the fee bump after successful unlock and relock as appropriate.
- Surface wrong-passphrase and signing/commit failures clearly in the speed-up overlay.
- Add coverage for unencrypted speed-up, encrypted locked speed-up, wrong passphrase, and relock behavior.

### LINKED ISSUE #659: Send validation: reject duplicate recipients and dust outputs before review
## Gap
QML validates each recipient independently, but it does not reject duplicate destination addresses or dust outputs before transaction creation like Qt widgets do.

## Work
Add cross-recipient duplicate detection and dust validation for single and multiple-recipient send. Surface errors in the form before review and keep backend transaction creation as the final guard.

## Code References
- `qml/models/walletqmlmodel.cpp:215`
- `qml/models/sendrecipient.cpp:121`
- `bitcoin/src/qt/walletmodel.cpp:162`
- `bitcoin/src/qt/sendcoinsentry.cpp:138`

### LINKED ISSUE #665: Coin control: clear selected coins after send and clear form
## Gap
QML clears recipients after send or clear-form actions, but selected coin-control inputs remain in `m_coin_control`. Qt widgets unselect all coin-control inputs after successful send and clear.

## Work
Clear selected coins when the send form is cleared and after successful send. Make the visible coin-control state and backend coin-control state stay synchronized.

## Code References
- `qml/pages/main.qml:194`
- `qml/models/sendrecipientslistmodel.cpp:177`
- `bitcoin/src/qt/sendcoinsdialog.cpp:553`
- `bitcoin/src/qt/sendcoinsdialog.cpp:562`

### LINKED ISSUE #692: RBF: Unlock wallet before confirming (PR #548)
I realized that PR for RBF got merged before this. I had left a `//FIXME: Unlock wallet before confirming (PR #548)
` in `SpeedUpOverlay.qml`.
Do you want to implement that as part of this PR or I can track that separately.

_Originally posted by @pseudoramdom in https://github.com/bitcoin-core/gui-qml/issues/548#issuecomment-4437221848_
 

Implementation should be validated by extending qml_test_wallet_rbf.py to include an encrypted wallet speed up

### LINKED ISSUE #705: Disable send review when a visible recipient row is empty
## Problem

When the send form has multiple recipient rows, the `Review` button can remain enabled even if a later recipient row is still empty. Clicking `Review` then asks for the wallet passphrase, but after unlocking nothing appears to happen.

Video reference: https://github.com/user-attachments/assets/b66dd1ee-545d-4d17-9d42-d81b716f2b3a

## Steps to reproduce

1. Open the wallet send page for an encrypted wallet.
2. Add more than one recipient row.
3. Fill out the first recipient with valid send details.
4. Leave the second recipient row empty.
5. Click `Review`.
6. Enter the wallet passphrase when prompted.

## Actual behavior

The `Review` button is enabled even though the visible second recipient row is empty. After passphrase entry, the app does not proceed to the review flow and there is no clear feedback.

## Expected behavior

The `Review` button should stay disabled while any visible recipient row is empty or incomplete, or the send flow should otherwise handle empty extra rows before prompting for the passphrase.

## Notes

This was noticed while testing multi-recipient send behavior and should be tracked separately from the current PR.

### LINKED ISSUE #712: Focus current password field first when updating wallet password
## Problem
When opening Wallet Password Settings in update-password mode, the current-password field appears first, but keyboard focus is assigned to the new-password field.

The current password field is shown near the top of `WalletPasswordSettings.qml`, while `focus: true` is hard-coded on the new-password field. This means users can start typing into the wrong field when changing an existing wallet password.

## Expected behavior
When updating an existing wallet password, the current-password field should receive initial focus.

When setting a password for an unencrypted wallet, the new-password field should continue to receive initial focus.

## Notes

Relevant code:
- `qml/pages/wallet/WalletPasswordSettings.qml`: current-password field and new-password field focus behavior

### LINKED ISSUE #715: Send fees: avoid creating wallet addresses for fee previews
## Problem

QML send fee preview currently calls `WalletQmlModel::ensurePreviewChangeAddress()`, which generates a wallet destination with the label `qml-fee-preview`:

- `qml/models/walletqmlmodel.cpp`: `ensurePreviewChangeAddress()`
- `m_wallet->getNewDestination(m_wallet->getDefaultAddressType(), "qml-fee-preview")`

That destination is cached for the current `WalletQmlModel` lifetime, so it does not create a new address on every estimate. However, it still creates a real receive/address-book entry, and after app restart or wallet model recreation another preview address can be generated.

The fee preview path only needs to estimate fees. It should not consume wallet addresses or leave `qml-fee-preview` entries in wallet state/address-book surfaces.

## Expected behavior

Send fee preview should estimate transaction fees without creating persistent wallet addresses or address-book entries.

Repeated preview updates, app restarts, and wallet reloads should not accumulate preview-only receive addresses.

## Implementation notes

The current code pins a preview change destination to avoid a worse behavior: if `coin_control.destChange` is left unset, `CreateTransactionInternal()` reserves an internal change destination and keeps it on success, even for unsigned preview transactions. Since fee preview runs repeatedly for High/Default/Low/custom estimates, simply removing the pinned destination would likely burn internal change indexes.

Potential safer approaches:

- Use a valid dummy `CTxDestination` with the same output type for preview-only `coin_control.destChange`, so fee estimation has the right script shape without asking the wallet for a new address.
- Or add a wallet/interface-level preview API that can create fee estimates using an internal change script without keeping a reserved destination.

The Review page should remain the authoritative final preview because it prepares the real transaction candidate.

## Acceptance criteria

- Fee preview no longer calls `getNewDestination()` just to estimate fees.
- Opening or updating the Send fee selector does not create `qml-fee-preview` address-book entries.
- Repeated fee preview updates and app restarts do not consume new wallet destinations.
- Existing fee-preview behavior remains covered for standard targets, custom fee, selected coins, and subtract-fee-from-amount.

### LINKED ISSUE #721: Create functional test to validate wallet compatibility with bitcoin-qt
this being `int64_t` then being serialized and deserialized as such differs from how qt widgets does it, which it does `unsigned int` - the big difference here is writing and reading 8 bytes vs 4 bytes

This should mean that requests made in qt widgets can't really be read correctly in QML, so an existing GUI user moving to QML, may see their old requests disappear. 

We should make an effort to maintain cross compatibility. So, compatibility with Qt Widgets gui entries being read in QML. But also, compatibility (as much as possible) with QML entries begin read by Qt widgets (should a user ever need to move back or use another project that uses the qt widgets gui/or qt widgets gui style serialization and deser).

Worthwhile to add and maintain a cross-compat func test.

Also, something to keep in mind for future development.

_Originally posted by @jarolrod in https://github.com/bitcoin-core/gui-qml/pull/708#discussion_r3255600769_

### LINKED ISSUE #728: Fix padding on Navigation buttons.
Image

### LINKED ISSUE #730: Unsubscribe from core signals in NodeModel destructor
```suggestion
    unsubscribeFromCoreSignals();
    setMempoolInfoPollingActive(false);
```

_Originally posted by @jarolrod in https://github.com/bitcoin-core/gui-qml/pull/724#discussion_r3322138213_

### LINKED ISSUE #731: App can hang after SIGINT during wallet initialization
## Problem

Sending SIGINT/Ctrl+C while the QML app is still initializing can leave the application running instead of shutting down. This is easiest to reproduce when startup is loading an encrypted wallet.

## Observed behavior

- Start the app with wallet support enabled and an encrypted wallet configured to load on startup.
- Send Ctrl+C/SIGINT while initialization is still in progress, around wallet loading.
- Core receives the shutdown request, but the QML app can stay open/running instead of completing shutdown.

## Suspected cause

The QML startup path queues shutdown through the same init executor thread that is running `appInitMain()`. If Core returns `false` from initialization because shutdown was requested, the QML side currently treats it like a generic initialization error instead of entering the shutdown path. Wallet controller initialization can also run from the failed init result.

## Proposed fix direction

- Treat failed initialization caused by `node.shutdownRequested()` as a shutdown request.
- Make QML shutdown handling one-shot.
- Call `node->startShutdown()` from all QML shutdown entry points before queued `appShutdown()`.
- Only initialize/adopt wallets after a successful Core initialization result.

## Validation target

Add focused NodeModel coverage for interrupted initialization and one-shot shutdown, plus a smoke test with a disposable regtest datadir and encrypted wallet where SIGINT during `Loading wallet…` exits cleanly.

### LINKED ISSUE #739: QML coin control selected inputs should be exclusive
While auditing coin-control parity in QML, Jarol pointed out that selected inputs were not exclusive during transaction preparation.

`CCoinControl::m_allow_other_inputs` defaults to `true`, so selected inputs are required but Core may still add other unselected wallet inputs. Qt Widgets overrides this before send preparation with:

```cpp
coin_control.m_allow_other_inputs = !coin_control.HasSelected();
```

QML should match that behavior for both fee preview and final transaction preparation so manually selected coins mean only those selected coins are available unless the UI explicitly offers an allow-other-inputs option.

Covered by PR #736.

johnny9:
This PR groups Preview-targeted fixes around Send correctness, fee preview behavior, wallet shutdown/lifecycle handling, receive-request compatibility, and small wallet UI polish.

## Issues
Fixes #467 - Send amount validation accounts for the selected fee estimate before allowing review.
Fixes #624 - Create-backup `View file` opens the selected wallet location and reports failures.
Fixes #636 - Fee bump confirmation prompts to unlock encrypted wallets and relocks after signing.
Fixes #692 - RBF speed-up prompts to unlock encrypted wallets before confirming and is covered by the encrypted-wallet RBF functional test.
Fixes #659 - Send validation rejects duplicate recipients and dust outputs before review.
Fixes #665 - Coin control selections are cleared after send and when clearing recipients.
Fixes #739 - QML coin control treats manually selected coins as exclusive during fee preview and transaction preparation.
Fixes #705 - Send review stays disabled when any visible multi-recipient row is incomplete, with a clear validation error.
Fixes #712 - Update-password mode focuses the current-password field first.
Fixes #715 - Fee previews no longer create persistent `qml-fee-preview` wallet addresses.
Fixes #721 - Receive request serialization stays compatible with bitcoin-qt and is covered by a bitcoin-qt v31.0 regression workflow.
Fixes #728 - Navigation button content padding is restored.
Fixes #730 - `NodeModel` unsubscribes from core signals during destruction.
Fixes #731 - SIGINT/Ctrl+C during interrupted initialization exits cleanly instead of hanging.

Refs #582
Refs #577
Refs #579

### DIFF
--- a/.github/workflows/gui-functional-tests.yml
+++ b/.github/workflows/gui-functional-tests.yml
@@ -163,6 +163,7 @@ jobs:
           python3 test/functional/qml_test_wallet_qt_compat.py
           python3 test/functional/qml_test_wallet_rbf.py
           python3 test/functional/qml_test_wallet_restore.py
+          python3 test/functional/qml_test_uri_import.py
           python3 test/functional/qml_test_wallet_settings.py
           python3 test/functional/qml_test_watchonly_wallet.py
           python3 test/functional/qml_test_create_wallet_name.py
--- a/.github/workflows/gui-functional-tests.yml
+++ b/.github/workflows/gui-functional-tests.yml
@@ -151,6 +151,7 @@ jobs:
           python3 test/functional/qml_test_peers.py
           python3 test/functional/qml_test_proxy.py
           python3 test/functional/qml_test_debug_log.py
+          python3 test/functional/qml_test_node_runtime_dialogs.py
           python3 test/functional/qml_test_settings_display.py
           python3 test/functional/qml_test_receive.py
           python3 test/functional/qml_test_activity_filter_export.py
@@ -166,3 +167,4 @@ jobs:
           python3 test/functional/qml_test_watchonly_wallet.py
           python3 test/functional/qml_test_create_wallet_name.py
           python3 test/functional/qml_test_wallet_selector_load.py
+          python3 test/functional/qml_test_shutdown.py
--- a/.github/workflows/gui-functional-tests.yml
+++ b/.github/workflows/gui-functional-tests.yml
@@ -18,6 +18,7 @@ jobs:
     runs-on: ubuntu-24.04
     env:
       LEGACY_BITCOIND_VERSION: "28.0"
+      QT_COMPAT_VERSION: "31.0"
 
     steps:
       - name: Checkout
@@ -30,6 +31,7 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y \
             build-essential cmake curl pkgconf python3 \
+            xvfb xauth \
             libevent-dev libboost-dev libsqlite3-dev libgl-dev libqrencode-dev \
             qt6-qpa-plugins \
             qt6-base-dev qt6-tools-dev qt6-l10n-tools qt6-tools-dev-tools \
@@ -80,6 +82,47 @@ jobs:
           path: releases/v${{ env.LEGACY_BITCOIND_VERSION }}
           key: ${{ runner.os }}-legacy-bitcoind-v${{ env.LEGACY_BITCOIND_VERSION }}
 
+      - name: Restore bitcoin-qt compatibility cache
+        uses: actions/cache/restore@v4
+        id: bitcoin-qt-compat-cache
+        with:
+          path: releases/v${{ env.QT_COMPAT_VERSION }}
+          key: ${{ runner.os }}-bitcoin-qt-v${{ env.QT_COMPAT_VERSION }}
+
+      - name: Install bitcoin-qt compatibility binary
+        run: |
+          archive="bitcoin-${QT_COMPAT_VERSION}-x86_64-linux-gnu.tar.gz"
+          base_url="https://bitcoincore.org/bin/bitcoin-core-${QT_COMPAT_VERSION}"
+          compat_dir="${GITHUB_WORKSPACE}/releases/v${QT_COMPAT_VERSION}"
+
+          mkdir -p "${compat_dir}"
+
+          if [ ! -x "${compat_dir}/bin/bitcoin-qt" ]; then
+            workdir="$(mktemp -d)"
+            trap 'rm -rf "${workdir}"' EXIT
+
+            curl -fsSLo "${workdir}/${archive}" "${base_url}/${archive}"
+            curl -fsSLo "${workdir}/SHA256SUMS" "${base_url}/SHA256SUMS"
+            (
+              cd "${workdir}"
+              grep " ${archive}$" "SHA256SUMS" | sha256sum -c -
+            )
+
+            tar -xzf "${workdir}/${archive}" \
+              --strip-components=1 \
+              -C "${compat_dir}" \
+              "bitcoin-${QT_COMPAT_VERSION}/bin"
+          fi
+
+          "${compat_dir}/bin/bitcoin-qt" --version | head -n 1
+
+      - name: Save bitcoin-qt compatibility cache
+        uses: actions/cache/save@v4
+        if: github.event_name != 'pull_request' && steps.bitcoin-qt-compat-cache.outputs.cache-hit != 'true'
+        with:
+          path: releases/v${{ env.QT_COMPAT_VERSION }}
+          key: ${{ runner.os }}-bitcoin-qt-v${{ env.QT_COMPAT_VERSION }}
+
       - name: Configure
         run: |
           cmake -B build \
@@ -97,6 +140,8 @@ jobs:
           LIBGL_ALWAYS_SOFTWARE: "1"
         run: |
           export BITCOIND_LEGACY="${GITHUB_WORKSPACE}/releases/v${LEGACY_BITCOIND_VERSION}/bin/bitcoind"
+          export BITCOIN_QT_COMPAT="${GITHUB_WORKSPACE}/releases/v${QT_COMPAT_VERSION}/bin/bitcoin-qt"
+          export BITCOIN_QT_RUNNER="xvfb-run -a"
           python3 test/functional/qml_test_bridge_sanity.py
           python3 test/functional/qml_test_onboarding.py
           python3 test/functional/qml_test_disablewallet_boot.py
@@ -114,6 +159,7 @@ jobs:
           python3 test/functional/qml_test_addresses.py
           python3 test/functional/qml_test_send_review.py
           python3 test/functional/qml_test_wallet_migration.py
+          python3 test/functional/qml_test_wallet_qt_compat.py
           python3 test/functional/qml_test_wallet_rbf.py
           python3 test/functional/qml_test_wallet_restore.py
           python3 test/functional/qml_test_wallet_settings.py
--- a/qml/models/nodemodel.cpp
+++ b/qml/models/nodemodel.cpp
@@ -417,6 +417,11 @@ void NodeModel::startNodeInitializionThread()
 
 void NodeModel::requestShutdown()
 {
+    if (m_shutdown_requested) {
+        return;
+    }
+    m_shutdown_requested = true;
+    stopShutdownPolling();
     Q_EMIT requestedShutdown();
 }
 
@@ -424,7 +429,7 @@ void NodeModel::initializeResult(bool success, interfaces::BlockAndHeaderTipInfo
 {
     if (!success) {
         if (m_startup_failure_dialog_shown) {
-            Q_EMIT requestedShutdown();
+            requestShutdown();
             Q_EMIT nodeInitialized();
             return;
         }
@@ -436,6 +441,9 @@ void NodeModel::initializeResult(bool success, interfaces::BlockAndHeaderTipInfo
         }
         m_startup_warning_messages.clear();
         setStartupError(startup_error);
+        if (m_node.shutdownRequested()) {
+            requestShutdown();
+        }
     } else {
         m_startup_error_messages.clear();
         m_node_ready = true;
@@ -460,20 +468,28 @@ void NodeModel::handleRunawayException(const QString& message)
 
 void NodeModel::startShutdownPolling()
 {
+    if (m_shutdown_polling_timer_id != 0) {
+        return;
+    }
     m_shutdown_polling_timer_id = startTimer(200ms);
 }
 
 void NodeModel::stopShutdownPolling()
 {
+    if (m_shutdown_polling_timer_id == 0) {
+        return;
+    }
     killTimer(m_shutdown_polling_timer_id);
+    m_shutdown_polling_timer_id = 0;
 }
 
 void NodeModel::timerEvent(QTimerEvent* event)
 {
-    Q_UNUSED(event)
+    if (event->timerId() != m_shutdown_polling_timer_id) {
+        return;
+    }
     if (m_node.shutdownRequested()) {
-        stopShutdownPolling();
-        Q_EMIT requestedShutdown();
+        requestShutdown();
     }
 }
 
--- a/qml/models/nodemodel.cpp
+++ b/qml/models/nodemodel.cpp
@@ -116,6 +116,7 @@ NodeModel::NodeModel(interfaces::Node& node)
 
 NodeModel::~NodeModel()
 {
+    unsubscribeFromCoreSignals();
     setMempoolInfoPollingActive(false);
     if (m_mempool_info_thread) {
         m_mempool_info_thread->quit();
@@ -800,4 +801,17 @@ void NodeModel::ConnectToBannedListChangedSignal()
             Q_EMIT bannedListChanged();
         });
     });
+}
+
+void NodeModel::unsubscribeFromCoreSignals()
+{
+    if (m_handler_notify_block_tip) {
+        m_handler_notify_block_tip->disconnect();
+    }
+    if (m_handler_notify_num_peers_changed) {
+        m_handler_notify_num_peers_changed->disconnect();
+    }
+    if (m_handler_notify_banned_list_changed) {
+        m_handler_notify_banned_list_changed->disconnect();
+    }
 }
--- a/qml/models/nodemodel.cpp
+++ b/qml/models/nodemodel.cpp
@@ -808,8 +808,23 @@ void NodeModel::unsubscribeFromCoreSignals()
     if (m_handler_notify_block_tip) {
         m_handler_notify_block_tip->disconnect();
     }
+    if (m_handler_notify_header_tip) {
+        m_handler_notify_header_tip->disconnect();
+    }
     if (m_handler_notify_num_peers_changed) {
         m_handler_notify_num_peers_changed->disconnect();
+    }
+    if (m_handler_notify_network_active_changed) {
+        m_handler_notify_network_active_changed->disconnect();
+    }
+    if (m_handler_notify_alert_changed) {
+        m_handler_notify_alert_changed->disconnect();
+    }
+    if (m_handler_message_box) {
+        m_handler_message_box->disconnect();
+    }
+    if (m_handler_question) {
+        m_handler_question->disconnect();
     }
     if (m_handler_notify_banned_list_changed) {
         m_handler_notify_banned_list_changed->disconnect();
--- a/qml/models/sendrecipient.cpp
+++ b/qml/models/sendrecipient.cpp
@@ -147,7 +147,7 @@ void SendRecipient::validateAmount()
             setAmountError(tr("Amount exceeds maximum limit of 21,000,000 BTC"));
         } else if (!m_address->isEmpty() && m_addressError.isEmpty() &&
                    IsDust(CTxOut{m_amount->satoshi(), GetScriptForDestination(DecodeDestination(m_address->address().toStdString()))},
-                          CFeeRate{DUST_RELAY_TX_FEE})) {
+                          m_wallet ? m_wallet->dustRelayFee() : CFeeRate{DUST_RELAY_TX_FEE})) {
             setAmountError(tr("Amount is too small to send."));
         } else if (m_wallet && m_amount->satoshi() > m_wallet->balanceSatoshi()) {
             setAmountError(tr("Amount exceeds available balance"));
--- a/qml/models/sendrecipient.cpp
+++ b/qml/models/sendrecipient.cpp
@@ -9,6 +9,9 @@
 #include <qml/models/walletqmlmodel.h>
 
 #include <key_io.h>
+#include <policy/feerate.h>
+#include <policy/policy.h>
+#include <script/script.h>
 
 SendRecipient::SendRecipient(WalletQmlModel* wallet, QObject* parent)
     : QObject(parent), m_wallet(wallet), m_address(new BitcoinAddress(this)), m_amount(new BitcoinAmount(this))
@@ -132,7 +135,7 @@ void SendRecipient::validateAddress()
         setAddressError("");
     }
 
-    Q_EMIT isValidChanged();
+    validateAmount();
 }
 
 void SendRecipient::validateAmount()
@@ -142,6 +145,10 @@ void SendRecipient::validateAmount()
             setAmountError(tr("Amount must be greater than zero"));
         } else if (m_amount->satoshi() > MAX_MONEY) {
             setAmountError(tr("Amount exceeds maximum limit of 21,000,000 BTC"));
+        } else if (!m_address->isEmpty() && m_addressError.isEmpty() &&
+                   IsDust(CTxOut{m_amount->satoshi(), GetScriptForDestination(DecodeDestination(m_address->address().toStdString()))},
+                          CFeeRate{DUST_RELAY_TX_FEE})) {
+            setAmountError(tr("Amount is too small to send."));
         } else if (m_wallet && m_amount->satoshi() > m_wallet->balanceSatoshi()) {
             setAmountError(tr("Amount exceeds available balance"));
         } else {
--- a/qml/models/walletqmlmodel.cpp
+++ b/qml/models/walletqmlmodel.cpp
@@ -535,7 +535,10 @@ bool WalletQmlModel::sendAmountExhaustsBalance() const
         return false;
     }
 
-    const CAmount balance{balanceSatoshi()};
+    wallet::CCoinControl coin_control{m_coin_control};
+    ApplySelectedInputsPolicy(coin_control);
+
+    const CAmount balance{m_wallet->getAvailableBalance(coin_control)};
     const CAmount total_amount{m_send_recipients->totalAmountSatoshi()};
     if (total_amount > balance) {
         return true;
@@ -1529,7 +1532,9 @@ bool WalletQmlModel::prepareTransactionInternal(std::optional<SecureString> pass
     CAmount balance = m_wallet->getAvailableBalance(coin_control);
     if (balance < total) {
         relock_guard.relock();
-        setTransactionStatus(tr("The wallet does not have enough balance for this transaction."));
+        setTransactionStatus(coin_control.HasSelected()
+            ? tr("Selected inputs do not cover the amount plus fee")
+            : tr("The wallet does not have enough balance for this transaction."));
         return false;
     }
 
@@ -1705,13 +1710,21 @@ void WalletQmlModel::listLockedCoins(std::vector<COutPoint>& outputs)
 
 void WalletQmlModel::selectCoin(const COutPoint& output)
 {
+    const bool was_selected{m_coin_control.IsSelected(output)};
     m_coin_control.Select(output);
+    if (!was_selected) {
+        Q_EMIT sendAmountExhaustsBalanceChanged();
+    }
     scheduleFeeEstimates();
 }
 
 void WalletQmlModel::unselectCoin(const COutPoint& output)
 {
+    const bool was_selected{m_coin_control.IsSelected(output)};
     m_coin_control.UnSelect(output);
+    if (was_selected) {
+        Q_EMIT sendAmountExhaustsBalanceChanged();
+    }
     scheduleFeeEstimates();
 }
 
@@ -1734,6 +1747,7 @@ void WalletQmlModel::clearSelectedCoins()
     if (m_coins_list_model) {
         m_coins_list_model->refreshSelection();
     }
+    Q_EMIT sendAmountExhaustsBalanceChanged();
     scheduleFeeEstimates();
 }
 
--- a/qml/models/walletqmlmodel.cpp
+++ b/qml/models/walletqmlmodel.cpp
@@ -1305,10 +1305,37 @@ void WalletQmlModel::scheduleFeeEstimates()
     if (m_fee_estimation_timer == nullptr) {
         return;
     }
+    m_fee_estimation_timer->stop();
 
-    if (!m_wallet || !m_send_recipients) {
+    if (!m_wallet || !m_send_recipients || !BuildRecipients(*m_send_recipients).has_value()) {
         clearFeeEstimates();
         return;
+    }
+
+    ++m_fee_estimate_request_id;
+
+    bool estimates_changed{!m_fee_estimates.isEmpty()};
+    bool custom_estimate_changed{m_custom_fee_estimate.has_value()};
+    if (estimates_changed) {
+        m_fee_estimates.clear();
+    }
+    if (custom_estimate_changed) {
+        m_custom_fee_estimate.reset();
+    }
+    if (estimates_changed || custom_estimate_changed) {
+        Q_EMIT estimatedFeeChanged();
+        Q_EMIT sendAmountExhaustsBalanceChanged();
+    }
+
+    bool pending_changed{!m_fee_estimate_pending};
+    if (pending_changed) {
+        m_fee_estimate_pending = true;
+        Q_EMIT feeEstimatePendingChanged();
+    }
+
+    if (estimates_changed || custom_estimate_changed || pending_changed) {
+        ++m_fee_estimate_revision;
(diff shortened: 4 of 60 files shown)


qml/models/nodemodel.cpp
```diff
@@ -116,6 +116,7 @@ NodeModel::NodeModel(interfaces::Node& node)
 
 NodeModel::~NodeModel()
 {
+    unsubscribeFromCoreSignals();
```
jarolrod, line 119: To finish closing out issue https://github.com/bitcoin-core/gui-qml/issues/730, we need to finish the implementation of `unsubscribeFromCoreSignals()`.

The precedent in Qt Widgets is that all owned Core signal handlers are disconnected during teardown. For example, `ClientModel` clears all node event handlers, and `BitcoinGUI` disconnects both message/question handlers.

`NodeModel` currently registers eight Core handlers, but `unsubscribeFromCoreSignals()` only disconnects three of them. The remaining header, network, alert, messagebox, and question handlers also capture `this`, and the runtime dialog handlers can block. They should be disconnected before shutdown begins.

You could do something like this:

<details>

```diff

commit 3f27fc303ce02b590250529729507494ba79c7af
Author: jarolrod <jarolrod@tutanota.com>
Date:   Mon Jun 1 18:05:47 2026 -0400

    fixup

diff --git a/qml/models/nodemodel.cpp b/qml/models/nodemodel.cpp
index 281357c40..3a9ed5ba1 100644
--- a/qml/models/nodemodel.cpp
+++ b/qml/models/nodemodel.cpp
@@ -805,13 +805,19 @@ void NodeModel::ConnectToBannedListChangedSignal()
 
 void NodeModel::unsubscribeFromCoreSignals()
 {
-    if (m_handler_notify_block_tip) {
-        m_handler_notify_block_tip->disconnect();
-    }
-    if (m_handler_notify_num_peers_changed) {
-        m_handler_notify_num_peers_changed->disconnect();
-    }
-    if (m_handler_notify_banned_list_changed) {
-        m_handler_notify_banned_list_changed->disconnect();
-    }
+    const auto disconnect = [](std::unique_ptr<interfaces::Handler>& handler) {
+        if (handler) {
+            handler->disconnect();
+            handler.reset();
+        }
+    };
+
+    disconnect(m_handler_notify_block_tip);
+    disconnect(m_handler_notify_header_tip);
+    disconnect(m_handler_notify_num_peers_changed);
+    disconnect(m_handler_notify_network_active_changed);
+    disconnect(m_handler_notify_alert_changed);
+    disconnect(m_handler_message_box);
+    disconnect(m_handler_question);
+    disconnect(m_handler_notify_banned_list_changed);
 }
diff --git a/test/test_nodemodel.cpp b/test/test_nodemodel.cpp
index 2362c47b2..8d0996f48 100644
--- a/test/test_nodemodel.cpp
+++ b/test/test_nodemodel.cpp
@@ -415,16 +415,37 @@ void NodeModelTests::destructorUnsubscribesCoreSignalsBeforeStoppingPolling()
     NiceMock<MockNode> node;
     MempoolState mempool;
     std::atomic<int> disconnected_handlers{0};
+    constexpr int expected_disconnected_handlers{8};
 
     InstallMempoolGetters(node, mempool);
     ON_CALL(node, handleNotifyBlockTip(testing::_))
         .WillByDefault(Invoke([&](interfaces::Node::NotifyBlockTipFn) {
             return interfaces::MakeCleanupHandler([&] { ++disconnected_handlers; });
         }));
+    ON_CALL(node, handleNotifyHeaderTip(testing::_))
+        .WillByDefault(Invoke([&](interfaces::Node::NotifyHeaderTipFn) {
+            return interfaces::MakeCleanupHandler([&] { ++disconnected_handlers; });
+        }));
     ON_CALL(node, handleNotifyNumConnectionsChanged(testing::_))
         .WillByDefault(Invoke([&](interfaces::Node::NotifyNumConnectionsChangedFn) {
             return interfaces::MakeCleanupHandler([&] { ++disconnected_handlers; });
         }));
+    ON_CALL(node, handleNotifyNetworkActiveChanged(testing::_))
+        .WillByDefault(Invoke([&](interfaces::Node::NotifyNetworkActiveChangedFn) {
+            return interfaces::MakeCleanupHandler([&] { ++disconnected_handlers; });
+        }));
+    ON_CALL(node, handleNotifyAlertChanged(testing::_))
+        .WillByDefault(Invoke([&](interfaces::Node::NotifyAlertChangedFn) {
+            return interfaces::MakeCleanupHandler([&] { ++disconnected_handlers; });
+        }));
+    ON_CALL(node, handleMessageBox(testing::_))
+        .WillByDefault(Invoke([&](interfaces::Node::MessageBoxFn) {
+            return interfaces::MakeCleanupHandler([&] { ++disconnected_handlers; });
+        }));
+    ON_CALL(node, handleQuestion(testing::_))
+        .WillByDefault(Invoke([&](interfaces::Node::QuestionFn) {
+            return interfaces::MakeCleanupHandler([&] { ++disconnected_handlers; });
+        }));
     ON_CALL(node, handleBannedListChanged(testing::_))
         .WillByDefault(Invoke([&](interfaces::Node::BannedListChangedFn) {
             return interfaces::MakeCleanupHandler([&] { ++disconnected_handlers; });
@@ -441,14 +462,14 @@ void NodeModelTests::destructorUnsubscribesCoreSignalsBeforeStoppingPolling()
     QObject::connect(model.get(), &NodeModel::mempoolInfoPollingActiveChanged, [&](bool active) {
         if (!active) {
             checked_shutdown_order = true;
-            QCOMPARE(disconnected_handlers.load(), 3);
+            QCOMPARE(disconnected_handlers.load(), expected_disconnected_handlers);
         }
     });
 
     model.reset();
 
     QVERIFY(checked_shutdown_order);
-    QCOMPARE(disconnected_handlers.load(), 3);
+    QCOMPARE(disconnected_handlers.load(), expected_disconnected_handlers);
 }
 
 void NodeModelTests::nodeNotificationHandlersUpdateModelThroughQueuedSignals()
```

</details>

test/test_nodemodel.cpp
```diff
@@ -344,6 +347,67 @@ void NodeModelTests::banPeerDisconnectsAddressAfterSuccessfulBan()
     QVERIFY(model.banPeer(QStringLiteral("127.0.0.1"), 3600));
 }
 
+void NodeModelTests::requestShutdownEmitsOnlyOnce()
```
jarolrod, line 354: While these unit tests are good it's probably worthwhile to do a functional test so we can have `appInitMain()` running and have the GUI be in the `load_on_startup` path while the OS signal is delivered -> since that is where the hang occurred previously. 

We could add a functional test like so:


<details>

```diff
commit f2bbf564f75c82ad193e0963784da18c91d8b6c7
Author: jarolrod <jarolrod@tutanota.com>
Date:   Mon Jun 1 18:48:53 2026 -0400

    test startup interrupt

diff --git a/.github/workflows/gui-functional-tests.yml b/.github/workflows/gui-functional-tests.yml
index 774f772fd..ecce78862 100644
--- a/.github/workflows/gui-functional-tests.yml
+++ b/.github/workflows/gui-functional-tests.yml
@@ -155,6 +155,7 @@ jobs:
           python3 test/functional/qml_test_receive.py
           python3 test/functional/qml_test_activity_filter_export.py
           python3 test/functional/qml_test_password_wallet.py
+          python3 test/functional/qml_test_startup_interrupt.py
           python3 test/functional/qml_test_send_receive.py
           python3 test/functional/qml_test_addresses.py
           python3 test/functional/qml_test_send_review.py
diff --git a/test/functional/qml_test_startup_interrupt.py b/test/functional/qml_test_startup_interrupt.py
new file mode 100644
index 000000000..82e94f416
--- /dev/null
+++ b/test/functional/qml_test_startup_interrupt.py
@@ -0,0 +1,208 @@
+#!/usr/bin/env python3
+# Copyright (c) 2026 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Regression test for interrupted QML initialization during wallet loading."""
+
+import os
+import shutil
+import signal
+import subprocess
+import sys
+import tempfile
+import time
+
+from qml_test_harness import find_gui_binary, qsettings_sandbox_args
+from qml_wallet_test_lib import (
+    pick_unused_port,
+    find_bitcoind,
+    rpc_call,
+    wait_for_rpc,
+    write_datadir,
+)
+
+
+WALLET_PASSWORD = "correct horse battery staple"
+STARTUP_WALLET_NAMES = [f"startup_interrupt_wallet_{i}" for i in range(4)]
+WALLET_LOADING_LOG_MARKER = "Using wallet"
+
+
+def debug_log_path(datadir):
+    return os.path.join(datadir, "regtest", "debug.log")
+
+
+def process_output(process):
+    if not process:
+        return ""
+    stdout = process.stdout.read().decode("utf-8", errors="replace") if process.stdout else ""
+    stderr = process.stderr.read().decode("utf-8", errors="replace") if process.stderr else ""
+    if stdout and stderr:
+        return f"stdout:\n{stdout}\n\nstderr:\n{stderr}"
+    return stdout or stderr
+
+
+def debug_log_tail(datadir, lines=200):
+    log_path = debug_log_path(datadir)
+    if not os.path.isfile(log_path):
+        return ""
+    with open(log_path, "r", encoding="utf8", errors="replace") as log_file:
+        return "".join(log_file.readlines()[-lines:])
+
+
+def stop_node(process, rpc_port):
+    if process and process.poll() is None:
+        try:
+            rpc_call(rpc_port, "stop")
+        except Exception:
+            process.send_signal(signal.SIGTERM)
+        try:
+            process.wait(timeout=20)
+        except subprocess.TimeoutExpired:
+            process.kill()
+            process.wait()
+
+
+def prepare_encrypted_startup_wallets(datadir, rpc_port):
+    process = subprocess.Popen(
+        [find_bitcoind(), f"-datadir={datadir}"],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+    try:
+        wait_for_rpc(rpc_port)
+        for wallet_name in STARTUP_WALLET_NAMES:
+            rpc_call(
+                rpc_port,
+                "createwallet",
+                {
+                    "wallet_name": wallet_name,
+                    "passphrase": WALLET_PASSWORD,
+                    "load_on_startup": True,
+                },
+            )
+    finally:
+        stop_node(process, rpc_port)
+
+
+def wait_for_startup_wallet_loading(process, datadir, start_offset, timeout=20):
+    log_path = debug_log_path(datadir)
+    deadline = time.time() + timeout
+    last_error = None
+
+    while time.time() < deadline:
+        if process.poll() is not None:
+            raise AssertionError(
+                f"GUI exited before startup wallet loading was observed "
+                f"(code {process.returncode}).\n{process_output(process)}\n"
+                f"debug.log tail:\n{debug_log_tail(datadir)}"
+            )
+
+        try:
+            if os.path.isfile(log_path) and os.path.getsize(log_path) > start_offset:
+                with open(log_path, "r", encoding="utf8", errors="replace") as log_file:
+                    log_file.seek(start_offset)
+                    # bitcoin-core-app does not mirror early Core init messages to
+                    # debug.log, so use the wallet database log emitted while Core is
+                    # loading startup wallets.
+                    if WALLET_LOADING_LOG_MARKER in log_file.read():
+                        return
+        except OSError as err:
+            last_error = err
+
+        time.sleep(0.05)
+
+    suffix = f": {last_error}" if last_error else ""
+    raise AssertionError(
+        f"Timed out waiting for startup wallet loading marker{suffix}.\n"
+        f"debug.log tail:\n{debug_log_tail(datadir)}"
+    )
+
+
+def launch_gui(datadir, config_home, socket_path):
+    env = dict(os.environ)
+    env["QT_QPA_PLATFORM"] = "offscreen"
+    settings_args = qsettings_sandbox_args(env, config_home)
+    args = [
+        find_gui_binary(),
+        f"-datadir={datadir}",
+        f"-test-automation={socket_path}",
+    ] + settings_args + [
+        "-logtimemicros",
+        "-debug",
+        "-debugexclude=libevent",
+        "-debugexclude=leveldb",
+        "-nolisten",
+    ]
+    print(f"[startup_interrupt] starting GUI: {' '.join(args)}")
+    return subprocess.Popen(
+        args,
+        env=env,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+
+
+def run_test():
+    tmpdir = tempfile.mkdtemp(prefix="qml_startup_interrupt_")
+    datadir = os.path.join(tmpdir, "node")
+    config_home = os.path.join(tmpdir, "config")
+    socket_path = os.path.join(tmpdir, "test_bridge.sock")
+    rpc_port = pick_unused_port()
+    p2p_port = pick_unused_port()
+    gui_process = None
+
+    try:
+        write_datadir(datadir, rpc_port, p2p_port)
+        prepare_encrypted_startup_wallets(datadir, rpc_port)
+        print("[startup_interrupt] encrypted startup wallets prepared")
+
+        log_path = debug_log_path(datadir)
+        start_offset = os.path.getsize(log_path) if os.path.isfile(log_path) else 0
+
+        gui_process = launch_gui(datadir, config_home, socket_path)
+        wait_for_startup_wallet_loading(gui_process, datadir, start_offset)
+        print("[startup_interrupt] startup wallet loading observed; sending SIGINT")
+
+        gui_process.send_signal(signal.SIGINT)
+        try:
+            gui_process.wait(timeout=30)
+        except subprocess.TimeoutExpired as err:
+            gui_process.kill()
+            gui_process.wait()
+            raise AssertionError(
+                "GUI did not exit after SIGINT during startup wallet loading.\n"
+                f"GUI output:\n{process_output(gui_process)}\n"
+                f"debug.log tail:\n{debug_log_tail(datadir)}"
+            ) from err
+
+        if gui_process.returncode != 0:
+            raise AssertionError(
+                f"Expected clean exit after SIGINT, got code {gui_process.returncode}.\n"
+                f"GUI output:\n{process_output(gui_process)}\n"
+                f"debug.log tail:\n{debug_log_tail(datadir)}"
+            )
+
+        print("[startup_interrupt] GUI exited cleanly after interrupted startup")
+        return 0
+
+    except Exception as err:  # noqa: BLE001 - preserve functional test context
+        print(f"\nFAILED: {err}", file=sys.stderr)
+        import traceback
+        traceback.print_exc()
+        return 1
+    finally:
+        if gui_process and gui_process.poll() is None:
+            gui_process.send_signal(signal.SIGTERM)
+            try:
+                gui_process.wait(timeout=10)
+            except subprocess.TimeoutExpired:
+                gui_process.kill()
+                gui_process.wait()
+        if os.getenv("KEEP_QML_TEST_TMPDIR") == "1":
+            print(f"Preserving test directory: {tmpdir}")
+        else:
+            shutil.rmtree(tmpdir, ignore_errors=True)
+
+
+if __name__ == "__main__":
+    sys.exit(run_test())
```
</details>

qml/models/sendrecipient.cpp
```diff
@@ -142,6 +145,10 @@ void SendRecipient::validateAmount()
             setAmountError(tr("Amount must be greater than zero"));
         } else if (m_amount->satoshi() > MAX_MONEY) {
             setAmountError(tr("Amount exceeds maximum limit of 21,000,000 BTC"));
+        } else if (!m_address->isEmpty() && m_addressError.isEmpty() &&
+                   IsDust(CTxOut{m_amount->satoshi(), GetScriptForDestination(DecodeDestination(m_address->address().toStdString()))},
+                          CFeeRate{DUST_RELAY_TX_FEE})) {
```
jarolrod: I think its wrong to use this here

This should use the node’s **active** dust relay fee instead of the **static** default DUST_RELAY_TX_FEE.

DUST_RELAY_TX_FEE is only the default value. The node policy can be changed with -dustrelayfee, and Qt Widgets already validates dust with node.getDustRelayFee(). 

Wallet transaction creation also uses the live relay dust fee, so this seems wrong to do like this.

qml/models/walletqmlmodel.cpp
jarolrod, line 1541: while auditing the changes around coin control, I think there is still an adjacent coin-control parity issue here. 

It seems selected coins are still not *exclusive* during transaction preparation.

CCoinControl::m_allow_other_inputs defaults to true, which means Core will require the selected inputs but may still add additional unselected wallet inputs. Qt Widgets overrides this before preparing a send: `coin_control.m_allow_other_inputs = !coin_control.HasSelected();`

As QML isn't doing that, it can prepare a transaction that spends the user-selected coins plus other wallet coins. I haven't exactly tested, but that is the logical conclusion. Probably worth it to reevaluate the context around this, the wallet, and qt widgets; and maybe include any changes here or open an issue to be addressed as a follow-up.

test/test_nodemodel.cpp
johnny9: Yeah i built these already. I guess they got lost in the shuffle. Will add back the shutdown qml functional test.

qml/models/sendrecipient.cpp
johnny9: Will look into this.

qml/models/walletqmlmodel.cpp
johnny9: I'll see how big of an issue this is.

johnny9: added #739 for tracking and made an attempt

qml/models/nodemodel.cpp
johnny9: Should disconnect all handlers now

qml/models/walletqmlmodel.cpp
johnny9: I think this is correct and matches the intended behavior of the user interface

qml/models/walletqmlmodel.cpp
```diff
@@ -377,21 +480,68 @@ qint64 WalletQmlModel::balanceSatoshi() const
...
+    if (AnyRecipientSubtractsFeeFromAmount(*m_send_recipients)) {
+        return false;
+    }
+    if (const auto fee = selectedFeeEstimate()) {
+        return AmountPlusFeeExceedsBalance(total_amount, *fee, balance);
+    }
+
+    // Without an estimate, a non fee-included send cannot safely spend the full
+    // balance because prepareTransaction() will still need to add a fee.
+    return total_amount >= balance;
```
pseudoramdom, line 554: We've got some weird UX going on.
If I select a UTXO with amount less than the one I'm sending, the "Review" button is enabled. But we get stopped after entering the passphrase.

We should probably emit `Q_EMIT sendAmountExhaustsBalanceChanged()` on coin selection

https://github.com/user-attachments/assets/5534e3ef-4876-4ea9-be2c-c41e098ff063

johnny9: I'll take a look at this. It is a gap with the calculations when considering coin control and the added coin control restriction to not include other coins. I changed coin control to not pull in other coins if the amount isnt covered and I didnt update the rules for deciding if we have enough of the amount covered

jarolrod: The root cause looks like the pre-review check is still using the full wallet balance, while the actual prepare path now uses coincontrol aware available balance, from my previous review.

Emitting `sendAmountExhaustsBalanceChanged()` on coin selection is needed so the button refreshes immediately, but it won’t fix the mismatch by itself. 

I think this path should use the same effective balance as `prepareTransactionInternal()`, then apply the same `CCoinControl` policy and call `getAvailableBalance(coin_control)`, ideally through a small shared helper so Review and prepare can’t drift in the future.

jarolrod: There is another adjacent state issue here around this code, thought not as important as this one.

When send inputs change, this only schedules a future preview. The existing fee estimate is not marked stale, and `feeEstimatePending` is not set until `requestFeeEstimatesNow()` runs after the debounce. 

Even after that, the preview still runs asynchronously. During that whole period, `sendAmountExhaustsBalance()` can still answer using either no estimate or an estimate from the previous send inputs.

That means the Review enabled state can temporarily be based on fee data that does not correspond to the current reality; of course keep in mind we're talking about 250ms, so its not the worst thing.

But architecturally, and this is where my comment focuses on, it means the model does not know whether the selected fee estimate is current, which makes the pre-review check quite weak, and its worthwhile to revisit the architecture,

pseudoramdom:
> Fixes https://github.com/bitcoin-core/gui-qml/issues/692 - RBF speed-up prompts to unlock encrypted wallets before confirming and is covered by the encrypted-wallet RBF functional test.

The text in red while doing RBF is redundant. 
Screenshot 2026-06-02 at 12 43 18 PM

test/functional/qml_test_wallet_rbf.py
```diff
@@ -198,6 +201,18 @@ def run_test():
...
+        gui.click("speedUpPassphraseConfirmButton")
+        gui.wait_for_property(
+            "speedUpPassphraseErrorText",
+            "text",
+            "The wallet password you entered was incorrect.",
+            timeout_ms=10000,
+        )
+        gui.set_text("speedUpPassphraseField", WALLET_PASSWORD)
+        gui.click("speedUpPassphraseConfirmButton")
+        gui.wait_for_property("speedUpPassphrasePopup", "opened", False, timeout_ms=10000)
```
jarolrod, line 222: maybe also assert the wallet is locked again after the successful bump? 

After the passphrase popup closes and the bump succeeds, checking `getwalletinfo()["unlocked_until"] == 0` would catch a regression where the gui signs but leaves the wallet unlocked.

.github/workflows/gui-functional-tests.yml
```diff
@@ -114,6 +159,7 @@ jobs:
           python3 test/functional/qml_test_addresses.py
           python3 test/functional/qml_test_send_review.py
           python3 test/functional/qml_test_wallet_migration.py
+          python3 test/functional/qml_test_wallet_qt_compat.py
```
jarolrod, line 163: don't we need to add the new shutdown test here?

qml/models/walletqmlmodel.cpp
johnny9: Stale pending reviews should be ignored now.

johnny9:
> > Fixes #692 - RBF speed-up prompts to unlock encrypted wallets before confirming and is covered by the encrypted-wallet RBF functional test.
> 
> The text in red while doing RBF is redundant. 

Set the default text to ""

pseudoramdom:
Screenshot 2026-06-04 at 11 28 18 AM

nit: The payment message and the import result seem to be not up to spec.

This is what I see in Figma. 
Screenshot 2026-06-04 at 11 30 40 AM

"The payment request imported..." should probably be a toast banner.

I can fix both in my upcoming PR for PSBT where I'm fixing the Send page layout.

johnny9:
> 
> "The payment request imported..." should probably be a toast banner.
> 
> I can fix both in my upcoming PR for PSBT where I'm fixing the Send page layout.

Ill leave it to you then if you already have some ideas or considerations with relation to PSBT.

qml/pages/wallet/Send.qml
```diff
@@ -87,8 +116,251 @@ PageStack {
...
+            if (result.hasLabel) {
+                root.recipient.label = result.label
+            }
+            // Lower the guard only after all writes are done. Field-change
+            // Connections must not fire checkClipboard() before the Fill
+            // handler has had a chance to set m_filledUri.
+            m_applyingUri = false
+            paymentRequestMessage = result.hasMessage ? result.uriMessage : ""
+            paymentRequestStatus = qsTr("Payment request imported from %1.").arg(source)
+            paymentRequestIsError = false
```
pseudoramdom, line 214: We should probably call `root.scheduleFeeEstimates()` so that we don't see a stale fee after importing the URI

qml/pages/wallet/Send.qml
```diff
@@ -87,8 +116,251 @@ PageStack {
...
+        }
```
pseudoramdom, line 215: If I import a URI without label/message to a form that has label & message, should we clear them out?

pseudoramdom (approved):
ACK fdb30e81effcd76d5fae0a201814384ca2849cf6

Issue #600, opened 2026-05-06
Settings: add reset GUI settings and open bitcoin.conf flows


johnny9:
## Code Gap
Qt Options includes reset/config-file flows such as resetting GUI settings and opening `bitcoin.conf`. The QML Settings app does not currently provide equivalent release-ready flows.

## Work Needed
- Add a QML-facing API for reset GUI settings behavior.
- Preserve or intentionally adapt Qt's backup/reset behavior.
- Add a QML-facing API/action to open `bitcoin.conf` where supported.
- Surface failures when the config file cannot be opened.
- Add tests or documented manual coverage for reset and config-file open behavior.

jarolrod:
I’m working on a settings/onboarding QML branch that partially addresses this issue. This can be considered picked up for now; once the PR is ready, I’ll revisit whether it fully addresses this issue or whether follow-up work should remain tracked here.

Issue #594, opened 2026-05-06
Proxy settings: add advanced Qt proxy parity


johnny9:
## Code Gap
QML proxy settings use combined address fields and basic proxy/Tor enablement. Qt exposes deeper proxy UI behavior that QML does not currently match.

## Work Needed
- Add separate proxy IP and port fields if desired for Qt parity.
- Add IPv4, IPv6, and Tor reachability indicators.
- Preserve previous disabled proxy values where Qt does.
- Add tests for any advanced behavior selected for implementation.

jarolrod:
I’m working on a settings/onboarding QML branch that partially addresses this issue. This can be considered picked up for now; once the PR is ready, I’ll revisit whether it fully addresses this issue or whether follow-up work should remain tracked here.

PR #742, opened 2026-06-02
Settings & Onboard Rework

### LINKED ISSUE #299: Add in an error message for the storage settings screen
Add in error message for storage settings screen.

Group 141

### LINKED ISSUE #373: Storage settings: Block storage limit initially selectable while disabled
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behaviour

[qml.webm](https://github.com/bitcoin-core/gui-qml/assets/93143998/1ccf3f09-8b15-4657-93f6-e47411148e03)

### Expected behaviour

It shouldn't be selectable if disabled, not only after `onCheckedChanged`.
As second part of video

### Steps to reproduce

see video

### Relevant log output

_No response_

### How did you obtain Bitcoin Core

Compiled from source

### What version of Bitcoin Core are you using?

2f08efce0bbac7a350f74bc1c8ccbbf39ab029aa

### Operating system and version

linux

### Machine specifications

_No response_

### LINKED ISSUE #405: Add error message for Storage Location Onboarding Page
### Please describe the feature you'd like to see added.

Base on the feedback from [mouxdesign](https://github.com/mouxdesign) and @GBKS in PR #397 add an error message if the storage location doesn't have enough space to store the necessary data.

### Is your feature related to a problem, if so please describe it.

_No response_

### Describe the solution you'd like

<details open>
<summary>Solution Proposed by @GBKS </summary>

Screenshot 2024-06-07 084048

### Describe any alternatives you've considered

_No response_

### Please leave any additional context

_No response_

### LINKED ISSUE #406: Detect Storage Amount Available In OnboardingStorageAmount.qml page
### Motivation

Based on the feedback by @GBKS in PR #397: 
> Can we detect how much space is available on each location and show it? And can we estimate the required space needed (the mock-up uses 500GB as a placeholder, which is meant to be replaced with something accurate).

### Possible solution

add a variable (i.e. optionsModel.diskSpace) that checks storage amount in a specific location (i.e. local drive, mounted drive or directory)

### Useful Skills

qml c++

### Guidance for new contributors

Want to work on this issue?

For guidance on contributing, please read [CONTRIBUTING.md](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md) before opening your pull request.

### LINKED ISSUE #419: notification that restart is needed for change to take effect
### Please describe the feature you'd like to see added.

When changing a setting, like enabling RPC, the app needs to be restarted for it to come into effect.
Thus having a notification that restart is needed should be there to let the user know.

qt gui has this.

### Is your feature related to a problem, if so please describe it.

_No response_

### Describe the solution you'd like

some text pop up telling that a restart is needed.

(maybe later an actual restart button could be added too, so it's not only the text but clicking the button will restart)

### Describe any alternatives you've considered

_No response_

### Please leave any additional context

_No response_

### LINKED ISSUE #431: Proxy needs to support unix sockets once this repo is refreshed from upstream
### Please describe the feature you'd like to see added.

In the UI where it says "Proxy IP address" it should say "Proxy location"...

### Is your feature related to a problem, if so please describe it.

_No response_

### Describe the solution you'd like

_No response_

### Describe any alternatives you've considered

_No response_

### Please leave any additional context

_No response_

### LINKED ISSUE #438: IP Address input value control - move validation and persistence out from onTextChanged into onEditingFinished but needs UI change (maybe adding a new "done" button)
Now, every time a user types into the ip address input value, the value is validated and persisted (if #433 gets merged), this is because of the nature of the app design, there's no where to "accept" the changes or land after editing a value. Originally the validation was triggered in the editing finished but it seemed more correct to [move](https://github.com/bitcoin-core/gui-qml/pull/391#issuecomment-2021432971) it into the text change due to the above reason. But when we added the persistence logic (#433), didn't feel right anymore, so @GBKS suggested to add a "done" button when we revert back to the correct event again:

image

### LINKED ISSUE #471: Unit choice should be global
- In the prototype, I set up an application-wide setting for the unit format. So if you toggle it in the form for one recipient, it also toggles it for all other recipients, the balance in the top left, any other screen you visit, etc. As a mental model, that seems easier track than having individual components have their own setting.

_Originally posted by @GBKS in https://github.com/bitcoin-core/gui-qml/issues/450#issuecomment-2948807685_

### LINKED ISSUE #511: Feature Parity 07: Complete Default/Tor Proxy Settings

### LINKED ISSUE #587: Display settings: add mBTC/uBTC units and Qt-compatible display-unit migration
## Code Gap
QML Display settings only expose BTC and sat in `qml/pages/settings/SettingsDisplayUnit.qml`, while Qt supports BTC, mBTC, uBTC, and sat.

QML also persists the setting with `SettingsKeys::DISPLAY_UNIT` (`displayUnit`), while Qt uses `DisplayBitcoinUnit` in `bitcoin/src/qt/optionsmodel.cpp`.

## Work Needed
- Add mBTC and uBTC options to `SettingsDisplayUnit.qml`.
- Extend QML amount formatting to handle all selected units consistently across Send, Receive, Activity, Wallet Badge, and payment detail views.
- Decide and implement migration/compatibility between QML `displayUnit` and Qt `DisplayBitcoinUnit`.
- Update display-unit tests for all supported units and migration behavior.

### LINKED ISSUE #588: Connection settings: add restart-required state for listen, NAT-PMP, and RPC changes
## Code Gap
`qml/components/ConnectionSettings.qml` lets users change `listen`, `natpmp`, and `server`, and `OptionsQmlModel` writes those settings after onboarding. Unlike proxy settings, the page does not show a dirty/restart-required state after changing these values.

Qt marks these changes as restart-required in `bitcoin/src/qt/optionsmodel.cpp` and surfaces that in the Options dialog.

## Work Needed
- Track initial/current values for `listen`, `natpmp`, and `server` in `OptionsQmlModel`.
- Expose a dirty/restart-required property for connection settings.
- Show restart-required messaging in `SettingsConnection.qml` after post-onboarding changes.
- Clear the dirty state when users revert to initial values.
- Add tests for persistence, dirty-state transitions, and restart messaging.

### LINKED ISSUE #593: Storage/developer settings: add restart-required state for prune, dbcache, and script-thread changes
## Code Gap
`qml/components/StorageSettings.qml` and `qml/components/DeveloperOptions.qml` let users edit settings that take effect on restart, including prune/prune size, database cache, and script verification threads.

`OptionsQmlModel` persists these values after onboarding, but the QML pages do not expose a dirty/restart-required state like Qt's Options dialog does.

## Work Needed
- Track initial/current values for prune, prune size, dbcache, and script verification threads in `OptionsQmlModel`.
- Expose dirty/restart-required properties for storage/developer settings.
- Show restart-required messaging in the relevant Settings pages after post-onboarding changes.
- Clear dirty state when users revert to initial values.
- Add tests for validation, persistence, dirty-state transitions, and restart messaging.

### LINKED ISSUE #596: Onboarding storage: replace placeholder custom data directory wiring
## Code Gap
`qml/components/StorageLocations.qml` exposes a custom storage directory option during onboarding. It calls `OptionsQmlModel::setCustomDataDirArgs()`, but `qml/models/options_model.cpp` still has a TODO and only updates QML model state:

```cpp
// TODO: add actual custom data wiring
qDebug() << "PlaceHolder: Created data directory: " << path;
```

This means the UI can imply that a custom data directory is configured even though the underlying node startup arguments/settings are not fully wired.

## Work Needed
- Replace the placeholder implementation with real data-directory configuration before node startup.
- Validate the selected directory and surface errors in the onboarding UI.
- Ensure the selected custom directory is used when the node starts.
- Persist/display the selected data directory consistently after onboarding.
- Add tests for default directory, custom directory, cancel, invalid path, and restart/startup behavior.

### LINKED ISSUE #598: Display settings: add third-party transaction URL setting
## Code Gap
Qt Options exposes a third-party transaction URL setting used to open transaction details in an external block explorer. The QML Settings app does not expose or persist an equivalent setting.

## Work Needed
- Add a QML-facing model property for the third-party transaction URL setting.
- Add a Display or Settings row to edit the URL.
- Wire transaction detail views to use the configured URL where applicable.
- Validate/sanitize the configured URL pattern.
- Add coverage for persistence and transaction-link behavior.

### LINKED ISSUE #599: Display settings: add overview money font option
## Code Gap
Qt Options exposes a money-font/display preference for overview balances. The QML Settings app does not expose an equivalent setting.

## Work Needed
- Decide which QML balance/amount surfaces should honor a money font setting.
- Add the setting to the QML options model and Display settings UI.
- Apply the selected font consistently to relevant amount displays.
- Persist the setting and cover it with tests.

### LINKED ISSUE #704: Onboard existing Bitcoin Core datadir without overwriting user settings
## Problem

QML onboarding needs to distinguish between:

1. whether the GUI already has a datadir choice, and
2. whether the active datadir/profile has already completed QML onboarding.

An existing Bitcoin Core user may launch the QML app for the first time using an already-initialized Core datadir. In that case, QML onboarding should still be shown if `qml_onboarded` has not yet been set for that profile, but the app must not treat that datadir as a fresh setup or overwrite existing user settings.

This is separate from #596, which covers custom datadir wiring for fresh onboarding.

### Datadir resolution

Resolve the active datadir in this order:

1. CLI `-datadir`
2. GUI-saved datadir in global QSettings (matching Qt's `strDataDir` behavior)
3. default datadir

The GUI-saved datadir should be global, not network-specific, like Qt. It should be persisted at the end of onboarding.

### When to show the storage directory page

Whether `OnboardingStorageLocation` is shown should depend on whether the datadir has already been chosen by CLI or by a previous GUI onboarding run.

Show `OnboardingStorageLocation` only when:

- no CLI `-datadir` is set, and
- no previously saved GUI datadir choice exists in QSettings

Skip `OnboardingStorageLocation` when:

- CLI `-datadir` is set, or
- QSettings already contains a GUI-saved datadir from a prior completed onboarding

`-resetguisettings` should mimic Qt behavior by re-showing datadir selection and resetting the relevant GUI/onboarding state.

Support for `-choosedatadir` should be tracked separately in a new issue.

### `qml_onboarded` remains datadir-specific

`qml_onboarded` should remain tied to the resolved active datadir/profile via that profile's `settings.json`, not to global QSettings.

After resolving the active datadir:

- if `qml_onboarded=true`, skip onboarding
- if `qml_onboarded` is missing/false, show onboarding

### Detecting an existing Core profile

Whether to show the remaining settings choices should not be based on `settings.json` existence alone.

Treat the resolved datadir as an existing Core profile if any same-network Core state already exists there, including for example:

- network-specific `settings.json`
- `bitcoin.conf`
- existing same-network block index / chainstate / block data
- prior prune state such as `prunedblockfiles`

If none of those are present, treat the datadir as a fresh profile.

### Page behavior when `qml_onboarded` is missing

If the resolved datadir is an existing Core profile:

- show only:
  - `OnboardingCover` ("Bitcoin Core App")
  - `OnboardingStrengthen` ("Strengthen bitcoin")
  - `OnboardingBlockclock` ("The block clock")
- skip:
  - `OnboardingStorageLocation` ("Storage location")
  - `OnboardingStorageAmount` ("Storage")
  - `OnboardingConnection` ("Starting initial download")
- do not treat the datadir as a fresh profile
- do not call `resetSettings()` by default
- do not overwrite existing persistent storage/connection/prune settings
- on completion, write only `qml_onboarded=true`

If the resolved datadir is a fresh profile:

- show the full onboarding flow
- on completion, persist the chosen datadir to global QSettings
- write the onboarding-selected settings plus `qml_onboarded=true`

## Non-goals / related work

- Fresh-profile custom datadir setup remains in #596.
- Storage onboarding UX improvements remain separate in #405 and #406.
- `-choosedatadir` support should be tracked in a new issue.

## Definition of Done

- Datadir resolution follows:
  1. CLI `-datadir`
  2. GUI-saved global QSettings datadir
  3. default datadir
- `OnboardingStorageLocation` is shown only when no CLI datadir and no GUI-saved datadir choice exist.
- `qml_onboarded` is evaluated for the resolved active datadir/profile, not as a global GUI flag.
- Existing profile + missing `qml_onboarded`:
  - onboarding is shown
  - only `OnboardingCover`, `OnboardingStrengthen`, and `OnboardingBlockclock` are shown
  - `OnboardingStorageLocation`, `OnboardingStorageAmount`, and `OnboardingConnection` are skipped
  - existing persistent settings are preserved
- Fresh profile + missing `qml_onboarded`:
  - full onboarding is shown
  - chosen datadir is persisted to global QSettings at the end of onboarding
- `-resetguisettings` re-shows datadir selection and resets the relevant GUI/onboarding state without defaulting to destructive resets of an existing Core profile.
- Functional coverage validates combinations of:
  - CLI `-datadir` set vs unset
  - GUI QSettings datadir set vs unset
  - resolved datadir looking fresh vs existing
  - `qml_onboarded` present vs absent
  - `-resetguisettings` behavior

## Notes

PR #549 introduced the useful concept of a `qml_onboarded` marker. That marker should mean QML onboarding has completed for a specific profile. It should not be replaced with a broad rule that skips onboarding whenever any Core settings or config exist, because existing Bitcoin Core users are exactly the users who still need first-run QML onboarding without losing their prior configuration.

jarolrod:
This PR wires the QML settings and onboarding flows to real Core backed settings instead of placeholder UI state. It adds persistent settings models, restart-required state, proxy configuration, display unit and money font support, third-party transaction URL support, onboarding storage validation, custom datadir handling, and test coverage for the new behavior.

- Adds model-backed onboarding settings for storage location, storage amount, connection options, and pre-init datadir handling.
- Adds storage validation for custom datadirs, available disk space, and prune-size constraints.
- Adds restart-required state for settings that require restart, while applying live settings such as NAT-PMP immediately.
- Completes default and Tor proxy settings with validation, persistence, disabled-value preservation, and restart notice behavior.
- Adds global display unit handling, Qt-compatible display-unit migration, and amount formatting across wallet views.
- Adds display settings for money font and third-party transaction URL templates.
- Adds `-resetguisettings` support and reset behavior for GUI-owned settings.
- Adds C++, QML, and functional test coverage for the new settings/onboarding behavior.

Closes #299
Closes #596  
Closes #406  
Closes #373  
Closes #593  
Closes #431  
Closes #471  
Closes #511  
Closes #588  
Closes #598  
Closes #599  
Closes #587
Closes #405
Closes #419
Closes #438
Closes #704

This PR also makes progress on #594, #600 but those should stay open for remaining UX or lifecycle follow up work.


test/CMakeLists.txt
```diff
@@ -56,6 +56,7 @@ target_include_directories(bitcoinqml_unit_tests
 target_link_libraries(bitcoinqml_unit_tests
   PRIVATE
     bitcoinqml
+    bitcoinqt
```
johnny9: No bitcoinqt dependancies should be present

test/CMakeLists.txt
johnny9, line 44: All of the new cpp modules should have unit test coverage

johnny9 (commented):
[ 90%] Building CXX object CMakeFiles/bitcoinqml.dir/bitcoin/src/init/bitcoin-qt.cpp.o
In file included from /usr/include/x86_64-linux-gnu/qt6/QtCore/qobject.h:18,
                 from /usr/include/x86_64-linux-gnu/qt6/QtCore/QObject:1,
                 from /home/johnny/github/gui-qml/qml/appmode.h:8,
                 from /home/johnny/github/gui-qml/qml/bitcoin.cpp:19:
/usr/include/x86_64-linux-gnu/qt6/QtCore/qmetatype.h: In instantiation of ‘struct QtPrivate::IsPointerToTypeDerivedFromQObject<QQuickCloseEvent*>’:
/usr/include/x86_64-linux-gnu/qt6/QtCore/qmetatype.h:1102:60:   required from ‘struct QMetaTypeId<QQuickCloseEvent*>’
/usr/include/x86_64-linux-gnu/qt6/QtCore/qmetatype.h:1122:38:   required from ‘struct QMetaTypeId2<QQuickCloseEvent*>’
/usr/include/x86_64-linux-gnu/qt6/QtCore/qobject_impl.h:32:41:   required from ‘struct QtPrivate::TypesAreDeclaredMetaType<QtPrivate::List<QQuickCloseEvent*> >’
/usr/include/x86_64-linux-gnu/qt6/QtCore/qobject.h:306:86:   required from ‘static typename std::enable_if<((QtPrivate::FunctionPointer<Func2>::ArgumentCount == -1) && (! is_convertible_v<Func2, const char*>)), QMetaObject::Connection>::type QObject::connect(const typename QtPrivate::FunctionPointer<Func>::Object*, Func1, const QObject*, Func2, Qt::ConnectionType) [with Func1 = void (QQuickWindow::*)(QQuickCloseEvent*); Func2 = {anonymous}::RunPreInitOnboarding(PreInitOnboardingContext&, const std::vector<std::__cxx11::basic_string<char> >&, bool)::<lambda()>; typename std::enable_if<((QtPrivate::FunctionPointer<Func2>::ArgumentCount == -1) && (! is_convertible_v<Func2, const char*>)), QMetaObject::Connection>::type = QMetaObject::Connection; typename QtPrivate::FunctionPointer<Func>::Object = QQuickWindow]’
/home/johnny/github/gui-qml/qml/bitcoin.cpp:389:21:   required from here
/usr/include/x86_64-linux-gnu/qt6/QtCore/qmetatype.h:842:23: error: invalid application of ‘sizeof’ to incomplete type ‘QQuickCloseEvent’
  842 |         static_assert(sizeof(T), "Type argument of Q_PROPERTY or Q_DECLARE_METATYPE(T*) must be fully defined");
      |                       ^~~~~~~~~
gmake[2]: *** [CMakeFiles/bitcoinqml.dir/build.make:319: CMakeFiles/bitcoinqml.dir/qml/bitcoin.cpp.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....

build failure on ubuntu 24 LTS

qml/pages/node/NetworkTraffic.qml
```diff
@@ -5,7 +5,7 @@
 import QtQuick 2.15
 import QtQuick.Controls 2.15
 import QtQuick.Layouts 1.15
-import Qt.labs.settings 1.0
+import QtCore
```
johnny9: This change isn't compatible with Ubuntu 24.04 provided Qt 6.4.2. Lets move this change into a new PR that we merge in later once Ubuntu 26.04.1 LTS is released in August.

qml/bitcoin.cpp
```diff
@@ -220,6 +229,182 @@ void ApplyTestSettingsDir()
...
+#endif
+
+    context.window = qobject_cast<QQuickWindow*>(context.engine->rootObjects().first());
+    if (!context.window) {
+        return PreInitOnboardingStatus::FAILED;
+    }
+
+    QEventLoop loop;
+    QObject::connect(context.engine->rootObjects().first(), SIGNAL(finished()), &loop, SLOT(quit()));
+    QObject::connect(context.window, &QQuickWindow::closing, &loop, [&] {
```
johnny9: This typed `QQuickWindow::closing` connection does not build on Qt 6.4 because the signal argument type, `QQuickCloseEvent`, is only forward-declared by the public `QQuickWindow` header there. Since this handler only quits the local event loop and does not inspect the close event, please use the string-based signal/slot form:

QObject::connect(context.window, SIGNAL(closing(QQuickCloseEvent*)), &loop, SLOT(quit()));

That keeps this compatible with the repo’s Qt 6.2+ range.

test/functional/qml_test_proxy.py
```diff
@@ -78,10 +68,8 @@ def test_default_proxy_toggle(gui):
     assert checked, "Expected proxyEnableSwitch to be checked after click"
     print("  Default proxy toggled ON: OK")
 
-    # proxySettingsDirty stays false during onboarding.
-    dirty = gui.get_property("settingsProxy", "proxySettingsDirty")
-    assert not dirty, "Expected proxySettingsDirty=False during onboarding"
-    print("  proxySettingsDirty=False during onboarding: OK")
+    gui.wait_for_property("settingsProxy", "proxySettingsDirty", True, timeout_ms=2000)
```
johnny9: This assertion is good coverage, but under Qt 6.4 the current test bridge click fallback can make the switch look checked without exercising the same QML handler path as a real user click. I hit this locally as `proxyEnableSwitch.checked == true` while `settingsProxy.proxySettingsDirty` stayed false.

The bridge helper itself is not in this diff hunk, so anchoring this here. I think the fix is to prefer a real mouse press/release before falling back to raw `toggle()` / `clicked()` invocation:

```diff
diff --git a/qml/test/testbridge.cpp b/qml/test/testbridge.cpp
--- a/qml/test/testbridge.cpp
+++ b/qml/test/testbridge.cpp
@@ -55,6 +55,27 @@ QByteArray clickObject(QObject* obj)
     if (int idx = meta->indexOfMethod("trigger()"); idx >= 0) {
         if (meta->method(idx).invoke(obj, Qt::DirectConnection)) return okResponse();
     }
+
+    // Qt < 6.8 does not expose QQuickAbstractButton::click(). Prefer a real
+    // mouse press/release before falling back to raw signal emission so QML
+    // handlers such as Switch::onToggled run exactly as they do for users.
+    if (auto* item = qobject_cast<QQuickItem*>(obj)) {
+        QQuickWindow* window = item->window();
+        if (window && item->isVisible() && item->isEnabled() && item->width() > 0 && item->height() > 0) {
+            const QPointF center = item->mapToScene(
+                QPointF(item->width() / 2.0, item->height() / 2.0));
+            const QPoint pos = center.toPoint();
+
+            QMouseEvent press(QEvent::MouseButtonPress, pos, window->mapToGlobal(pos),
+                              Qt::LeftButton, Qt::LeftButton, Qt::NoModifier);
+            QMouseEvent release(QEvent::MouseButtonRelease, pos, window->mapToGlobal(pos),
+                                Qt::LeftButton, Qt::NoButton, Qt::NoModifier);
+
+            QCoreApplication::sendEvent(window, &press);
+            QCoreApplication::sendEvent(window, &release);
+            return okResponse();
+        }
+    }
 
     // Fallback for Qt < 6.8 (no QQuickAbstractButton::click()): invoke both
     // toggle() (if checkable) and the clicked() signal. We can't tell which
@@ -72,25 +93,6 @@ QByteArray clickObject(QObject* obj)
     }
     if (invoked_any) return okResponse();
 
-    // Last resort for items without click()/trigger()/toggle()/clicked()
-    // (bare Item + MouseArea, etc.): synthesize a real mouse press + release
-    // at the item's center.
-    if (auto* item = qobject_cast<QQuickItem*>(obj)) {
-        QQuickWindow* window = item->window();
-        if (window && item->isVisible() && item->width() > 0 && item->height() > 0) {
-            const QPointF center = item->mapToScene(
-                QPointF(item->width() / 2.0, item->height() / 2.0));
-            const QPoint pos = center.toPoint();
-
-            QMouseEvent press(QEvent::MouseButtonPress, pos, window->mapToGlobal(pos),
-                              Qt::LeftButton, Qt::LeftButton, Qt::NoModifier);
-            QMouseEvent release(QEvent::MouseButtonRelease, pos, window->mapToGlobal(pos),
-                                Qt::LeftButton, Qt::NoButton, Qt::NoModifier);
-
-            QCoreApplication::sendEvent(window, &press);
-            QCoreApplication::sendEvent(window, &release);
-            return okResponse();
-        }
-    }
-
     QJsonObject resp;
```

qml/pages/MainWindow.qml
```diff
@@ -5,21 +5,58 @@
...
+    readonly property bool waitForPostOnboardingWalletRoute: preInitOnboardingRanForUi && desktopWalletMode
+    property bool postOnboardingWalletRouteResolved: false
+    property bool postOnboardingWalletWizardShown: false
+
+    function showPostOnboardingWalletWizard() {
+        if (!appWindow.waitForPostOnboardingWalletRoute || appWindow.postOnboardingWalletWizardShown) {
+            return
+        }
+        appWindow.postOnboardingWalletWizardShown = true
+        main.push(createWalletWizard, {
```
johnny9: This new post-onboarding wallet route opens the existing add-wallet intro in `CreateWalletWizard.Context.Onboarding`, but that page still shows the Back button. In this context Back sends the user back out of the create-wallet flow after pre-init onboarding has already committed startup settings and handed off to the main app.



```diff
diff --git a/qml/pages/wallet/CreateWalletWizard.qml b/qml/pages/wallet/CreateWalletWizard.qml
--- a/qml/pages/wallet/CreateWalletWizard.qml
+++ b/qml/pages/wallet/CreateWalletWizard.qml
@@ -87,13 +87,13 @@ PageStack {
 
             header: NavigationBar2 {
                 id: navbar
-                leftItem: NavButton {
+                leftItem: root.launchContext === CreateWalletWizard.Context.Onboarding ? null : NavButton {
                     objectName: "createWalletWizardBackButton"
                     iconSource: "image://images/caret-left"
                     text: qsTr("Back")
                     onClicked: root.finished()
                 }
                 rightItem: NavButton {
```

qml/datadir.cpp
```diff
@@ -0,0 +1,298 @@
...
+}
+
+bool ShouldShowDataDirChooser(const ArgsManager& args)
+{
+    if (HasExplicitDataDirArg(args)) return false;
+
+    QSettings settings;
+    const QString data_dir = ReadGuiDataDir();
+    const QFileInfo data_dir_info(data_dir);
+    return !data_dir_info.exists() ||
```
johnny9: This only checks whether the saved custom datadir still exists and is a directory. If the directory still exists but becomes unwritable, startup skips the chooser and still applies it as `-datadir`; `ApplyGuiDataDirSetting()` also ignores `EnsureDataDir()` failure below.

Please revalidate the saved custom datadir before deciding to skip the chooser, and stop applying it if it can no longer be ensured:

```diff
diff --git a/qml/datadir.cpp b/qml/datadir.cpp
--- a/qml/datadir.cpp
+++ b/qml/datadir.cpp
@@ -263,9 +263,11 @@ bool ShouldShowDataDirChooser(const ArgsManager& args)
 
     QSettings settings;
     const QString data_dir = ReadGuiDataDir();
+    const QString validation_error = IsDefaultDataDir(data_dir) ? QString{} : ValidateCustomDataDir(data_dir);
     const QFileInfo data_dir_info(data_dir);
-    return !data_dir_info.exists() ||
+    return !validation_error.isEmpty() ||
+           !data_dir_info.exists() ||
            !data_dir_info.isDir() ||
            args.GetBoolArg("-choosedatadir", false) ||
            args.GetBoolArg("-resetguisettings", false) ||
@@ -290,8 +292,7 @@ bool ApplyGuiDataDirSetting(ArgsManager& args)
     const QString data_dir = ReadGuiDataDir();
     if (IsDefaultDataDir(data_dir)) return false;
 
-    QString ignored_error;
-    EnsureDataDir(data_dir, &ignored_error);
+    if (!EnsureDataDir(data_dir)) return false;
     return ApplyDataDirArg(args, data_dir);
 }
```

qml/models/options_model.cpp
```diff
@@ -87,233 +81,323 @@ bool TokenLooksLikePath(const QString& token)
...
+    if (restartRequired() != before.restart) Q_EMIT restartRequiredChanged();
+}
+
+void OptionsQmlModel::applyRuntimeCoreChange(const QmlCoreSettings::Change& change, const DirtySnapshot& before)
+{
+    if (!change.accepted || !QmlCoreSettings::ValuesChanged(change)) return;
+    m_core_settings.writeToNode(m_node, m_args, change.setting_name);
+    refreshCoreSettingStatuses();
+    QmlCoreSettings::EmitCoreSettingSignals(*this, change);
+    if (change.setting_name == QStringLiteral("natpmp")) {
```
johnny9, line 251: This live-applies NAT-PMP even though the settings UI/test framing treats connection changes as restart-required. It also means toggling NAT-PMP does not participate in `connectionSettingsDirty()`, so the restart notice can disappear even though the saved startup setting changed.

If NAT-PMP is meant to match the other connection startup settings here, please make it dirty/restart-required and remove the live map-port call:

```diff
diff --git a/qml/models/options_model.cpp b/qml/models/options_model.cpp
--- a/qml/models/options_model.cpp
+++ b/qml/models/options_model.cpp
@@ -157,7 +157,9 @@ bool OptionsQmlModel::connectionSettingsDirty() const
 {
     const QmlCoreSettings::Values& values = m_core_settings.values();
-    return values.listen != m_initial_core_values.listen || values.server != m_initial_core_values.server;
+    return values.listen != m_initial_core_values.listen ||
+           values.natpmp != m_initial_core_values.natpmp ||
+           values.server != m_initial_core_values.server;
 }
@@ -234,14 +236,6 @@ void OptionsQmlModel::applyRuntimeCoreChange(const QmlCoreSettings::Change& chan
     m_core_settings.writeToNode(m_node, m_args, change.setting_name);
     refreshCoreSettingStatuses();
     QmlCoreSettings::EmitCoreSettingSignals(*this, change);
-    if (change.setting_name == QStringLiteral("natpmp")) {
-        // Disabling NAT-PMP joins the mapport thread and can briefly block.
-        // Defer the live apply so the switch state and animation are not
-        // held behind that work.
-        QTimer::singleShot(200, this, [this, natpmp = change.after.natpmp] {
-            m_node.mapPort(natpmp);
-        });
-    }
     emitDirtySignals(before);
 }
```

If live NAT-PMP is intentional, then the issue/test naming should be adjusted to say NAT-PMP is deliberately live-applied rather than restart-required.

qml/models/options_model.cpp
```diff
@@ -478,57 +564,59 @@ void OptionsQmlModel::setDisplayUnit(int new_display_unit)
...
-QString OptionsQmlModel::displayUnitLabelForAmount(qint64 satoshi) const
+QVariantList OptionsQmlModel::thirdPartyTransactionLinks(const QString& txid) const
 {
-    if (m_display_unit != 1) return QStringLiteral("₿");
-    return (qAbs(satoshi) == 1) ? QStringLiteral("sat") : QStringLiteral("sats");
+    QVariantList links;
+    const QStringList urls = m_third_party_transaction_urls.split(QLatin1Char('|'), Qt::SkipEmptyParts);
+    for (QString url : urls) {
+        url = url.trimmed();
+        const QUrl parsed{url, QUrl::StrictMode};
```
johnny9, line 599: The UI says `%s` is the placeholder for the transaction hash, but entries without `%s` still produce a link. Those links can never point at the selected transaction. Please skip malformed entries here so only URLs that actually include the txid are exposed.

```suggestion
        url = url.trimmed();
        if (!url.contains(QStringLiteral("%s"))) continue;
        const QUrl parsed{url, QUrl::StrictMode};
```

qml/pages/wallet/WalletBadge.qml
```diff
@@ -161,7 +161,7 @@ Button {
                 CoreText {
                     id: balanceText
                     visible: root.showBalance
-                    text: root.balance + " " + (optionsModel.displayUnit === 1 ? (root.balanceSatoshi === 1 ? qsTr("sat") : qsTr("sats")) : "₿")
+                    text: root.balance + " " + optionsModel.displayUnitLabelForAmount(root.balanceSatoshi)
                     color: Theme.color.neutral7
                     font.pixelSize: 15
                     bold: true
```
johnny9, line 170: This overview balance still uses the default font, so the new money font setting is only applied in some wallet surfaces. Please apply `optionsModel.moneyFont` here as well.

```suggestion
                    text: root.balance + " " + optionsModel.displayUnitLabelForAmount(root.balanceSatoshi)
                    color: Theme.color.neutral7
                    font.family: optionsModel.moneyFont.family
                    font.weight: optionsModel.moneyFont.weight
                    font.pixelSize: 15
                    bold: true
```

qml/pages/onboarding/OnboardingStorageAmount.qml
```diff
@@ -26,26 +33,38 @@ Page {
...
+                        assumedBlockchainSize: root.assumedBlockchainSize
+                        assumedChainstateSize: root.assumedChainstateSize
+                        customStorage: root.customStorage
+                        customStorageAmount: root.customStorageAmount
                         Layout.maximumWidth: 450
                         Layout.alignment: Qt.AlignCenter
+                        onStorageSelectionChanged: {
+                            root.customStorage = customStorage
+                            root.customStorageAmount = customStorageAmount
+                        }
```
johnny9: Selecting the custom storage option logs a QML deprecation warning because this handler relies on implicit signal-parameter injection:

```text
Parameter "customStorage" is not declared. Injection of parameters into signal handlers is deprecated.
```

Please use a JavaScript function with formal parameters instead.

```suggestion
                        onStorageSelectionChanged: function(customStorage, customStorageAmount) {
                            root.customStorage = customStorage
                            root.customStorageAmount = customStorageAmount
                        }
```

johnny9 (commented):
Left comments

qml/components/StorageSettings.qml
```diff
@@ -11,20 +11,19 @@ ColumnLayout {
     id: root
     property bool customStorage: false
     property int customStorageAmount
+    property bool showRestartNotice: false
     spacing: 4
+    SettingsRestartNotice {
```
pseudoramdom, line 37: 35795a12129fe71427c2a9b565846b94abe60192
We should probably use `InfoBanner` here and not create one-off views for this.

qml/models/options_model.cpp
```diff
@@ -109,63 +158,162 @@ OptionsQmlModel::OptionsQmlModel(interfaces::Node& node, bool is_onboarded)
...
         }
         Q_EMIT listenChanged(new_listen);
+        emitDirtySignals(old_connection, storageSettingsDirty(), developerSettingsDirty(), proxySettingsDirty(), walletSettingsDirty(), old_restart);
     }
 }
 
 void OptionsQmlModel::setMaxMempoolSizeMB(int new_max_mempool_size_mb)
 {
     if (new_max_mempool_size_mb != m_max_mempool_size_mb) {
+        const DirtySnapshot before = dirtySnapshot();
```
pseudoramdom, line 320: There's no `struct DirtySnapshot` in  32afebfcd04a45a2968b7913c0e948dc11e07365. This commit won't compile I guess.
You might want to pull that from e92097a6ef0df5c8fa2e85b0344bc23c74657435?

qml/models/options_model.cpp
```diff
@@ -183,120 +332,156 @@ void OptionsQmlModel::setNatpmp(bool new_natpmp)
...
-            }
+        const bool old_proxy = proxySettingsDirty();
+        const bool old_restart = restartRequired();
+        if (enabled && m_proxy_address.isEmpty()) {
+            m_proxy_address = defaultProxyAddress();
+            Q_EMIT proxyAddressChanged(m_proxy_address);
         }
-        if (proxySettingsDirty() != was_dirty) {
-            Q_EMIT proxySettingsDirtyChanged();
+        if (enabled && !validateProxyLocation(m_proxy_address).isEmpty()) {
```
pseudoramdom: 32afebfcd04a45a2968b7913c0e948dc11e07365
Should we check this before we update the `m_proxy_address`

qml/models/options_model.h
```diff
@@ -188,8 +228,14 @@ public Q_SLOTS:
     QString m_language;
     QStringList m_available_languages;
     int m_display_unit{0};
+    QString m_third_party_transaction_urls;
+    QString m_money_font_choice;
 
     common::SettingsValue pruneSetting() const;
+    common::SettingsValue proxySetting(bool enabled, const QString& address) const;
+    void resetDirtySnapshots();
+    void emitDirtySignals(bool old_connection, bool old_storage, bool old_developer, bool old_proxy, bool old_wallet, bool old_restart);
```
pseudoramdom: Would this be better?
`void emitDirtySignals(const DirtySnapshot& before)`

qml/components/ProxySettings.qml
```diff
@@ -51,26 +43,31 @@ ColumnLayout {
...
-        header: ipAndPortHeader
-        errorText: invalidIpError
-        state: !defaultProxyEnable.loadedItem.checked ? "DISABLED" : "FILLED"
-        showErrorText: !defaultProxy.loadedItem.validInput && defaultProxyEnable.loadedItem.checked
-        actionItem: IPAddressValueInput {
+        header: proxyLocationHeader
+        errorText: loadedItem ? loadedItem.validationError : ""
+        state: optionsModel.proxyEnabled ? "FILLED" : "DISABLED"
+        showErrorText: loadedItem && !loadedItem.validInput && optionsModel.proxyEnabled
+        actionItem: ProxyLocationInput {
```
pseudoramdom, line 76: 016c79e75aa52e77f055945af4bfa4629e4de8c9
Can we repurpose `IPAddressValueInput.qml`?

qml/models/options_model.cpp
```diff
@@ -605,6 +610,15 @@ bool OptionsQmlModel::selectCustomDataDir(const QString& path)
         return true;
     }
 
+    try {
+        const fs::path data_dir_path = QStringToPath(local_path);
+        if (TryCreateDirectories(data_dir_path)) {
+            TryCreateDirectories(data_dir_path / "wallets");
```
pseudoramdom: 0ec1b80f8156b1c029402241d269aad10384e3d6
We deliberately persist the location even thought the user might not have completed the onboarding, yeah?

qml/components/StorageLocations.qml
```diff
@@ -35,21 +40,22 @@ ColumnLayout {
         ButtonGroup.group: group
         text: qsTr("Custom")
         description: qsTr("Choose the directory and storage device.")
-        customDir: customDirOption.checked ? fileDialog.currentFolder.toString() : ""
+        customDir: checked && optionsModel.dataDir !== optionsModel.getDefaultDataDirString ? optionsModel.dataDir : ""
         checked: optionsModel.dataDir !== optionsModel.getDefaultDataDirString
         onClicked: fileDialog.open()
     }
     FileDialog {
```
pseudoramdom: Should this use `FolderDialog` instead?

qml/bitcoin.cpp
```diff
@@ -220,6 +225,28 @@ void ApplyTestSettingsDir()
...
+    if (!data_dir.isEmpty() && data_dir != default_dir) {
+        const fs::path data_dir_path = fs::u8path(data_dir.toStdString());
+        try {
+            TryCreateDirectories(data_dir_path);
+            TryCreateDirectories(data_dir_path / "wallets");
+        } catch (const fs::filesystem_error&) {
+            // Defer the user-visible error to InitConfig(), matching other
+            // startup datadir validation failures.
+        }
+        gArgs.SoftSetArg("-datadir", fs::PathToString(data_dir_path));
```
pseudoramdom: Should this be inside the `try`? 
If the saved path is unwritable, invalid, or no longer usable, startup proceeds with a bad `-datadir`

pseudoramdom (commented):
Partial review till 0ec1b80f8156b1c029402241d269aad10384e3d6 `qml: persist custom data directory selection`

qml/pages/wallet/ActivityDetails.qml
```diff
@@ -140,6 +142,26 @@ Page {
...
 
+            ColumnLayout {
+                id: thirdPartyLinks
+                objectName: "activityDetailsThirdPartyLinks"
+                visible: root.txid.length > 0 && optionsModel.thirdPartyTransactionLinks(root.txid).length > 0
+                Layout.fillWidth: true
+                Layout.bottomMargin: visible ? 20 : 0
+                spacing: 8
+
+                Repeater {
```
epicleafies, line 155: **Blocker:** A `Repeater` as a direct child of a `ColumnLayout` whose model populates after construction (`root.txid` arrives via the StackView push) hits the Qt 6.4 Quick Layouts use-after-free: `QQuickLayout::effectiveSizePolicy_helper` runs on a freed delegate. It SIGSEGVs only under a real windowing backend, so the offscreen suite stays green. PR #746 fixes the same crash class (originally #750) in `AlertPopup.qml`: the action buttons live in a `Row` positioner inside the layout, with the delegate using `width` instead of `Layout.fillWidth`. The same wrap applies here.

test/functional/qml_test_preinit_onboarding.py
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env python3
```
epicleafies, line 1: **Blocker:** This test and `qml_test_resetguisettings.py` are absent from `.github/workflows/gui-functional-tests.yml`, so the pre-init onboarding and reset-GUI-settings flows run in no CI job. Add both to the functional-test step. They refuse an external `--socket-path` and self-launch, so they go in without the shared-socket args.

qml/components/StorageOptions.qml
```diff
@@ -12,51 +12,91 @@ import "../controls"
...
+    }
     OptionButton {
+        objectName: "storageReduceOption"
         Layout.fillWidth: true
         ButtonGroup.group: group
         text: qsTr("Reduce storage")
-        description: qsTr("Uses about %1GB. For simple wallet use.").arg(chainModel.assumedChainstateSize + 2)
-        recommended: true
-        checked: !root.customStorage && optionsModel.prune
+        description: qsTr("Uses about %1GB. For simple wallet use.").arg(root.reduceRequiredGB) + root.storageSuffix()
```
epicleafies: **Suggestion:** `qsTr("Uses about %1GB. For simple wallet use.").arg(...) + root.storageSuffix()` concatenates two separately-translated strings (`storageSuffix()` is `qsTr(" %1GB available.")`). Word order varies between languages, so the glued sentence mistranslates. Merge them into one `qsTr` with a second placeholder for the available-GB value. Same at line 77.

qml/models/onboardingoptionsmodel.h
```diff
@@ -0,0 +1,180 @@
...
+    bool m_can_listen_ipc;
+    QString m_data_dir;
+    QString m_custom_datadir_string;
+    CoreSettingsModel m_core_settings;
+    QString m_preview_error;
+    int m_assumed_blockchain_size{0};
+    int m_assumed_chainstate_size{0};
+    uint64_t m_storage_request_id{0};
+    uint64_t m_storage_available_bytes{0};
+    uint64_t m_storage_capacity_bytes{0};
```
epicleafies: **Nit:** `m_storage_capacity_bytes` is assigned in `applyStorageCheckResult` (`onboardingoptionsmodel.cpp:256`) but never read. Either it is dead and the member plus assignment can go, or it is meant to drive UI and is unwired.

epicleafies (commented):
Concept ACK. I wasn't able to build or test as the branch does not compile on Qt 6.4.2. Will re-review once it is buildable.

Two things stand out for the 6.4 baseline: the import QtCore swap is repo-wide, not just NetworkTraffic.qml (reply on that thread), and a new Repeater-in-ColumnLayout in ActivityDetails repeats the Quick Layouts crash class. The rest is one i18n suggestion, a nit, and replies extending existing threads. I traced the onboarding startup path: the InitConfig / parameter-interaction double-pass is idempotent by construction, so that is not a concern.

qml/pages/node/NetworkTraffic.qml
epicleafies: The same `Qt.labs.settings 1.0` to `import QtCore` swap is in 11 other files, so this is repo-wide: `controls/Theme.qml`, `pages/main.qml`, `components/BlockClock.qml`, `components/BlockClockDisplayMode.qml`, `components/ExternalPopup.qml`, `components/ThemeSettings.qml`, `pages/node/Peers.qml`, `pages/wallet/Activity.qml`, `pages/wallet/Send.qml`, `pages/wallet/RequestPayment.qml`, `pages/wallet/ActivityDetails.qml`. The `QtCore` QML module is 6.5+ (absent on a 6.4.2 install), and QML resolves imports at load, so each of these (including the main window) fails to load on 6.4. `BlockClockDisplayMode.qml` and `ExternalPopup.qml` no longer use `Settings` at all, so there the import can be dropped rather than reverted.

test/CMakeLists.txt
epicleafies: On which modules specifically: `onboarding_storage.cpp` (`Evaluate` / `ShouldRecommendPrune`) has no direct C++ test, and the QML test only exercises the mock's reimplementation of that math, so a bug in the real `Evaluate()` would slip through. `legacy_settings_migration.cpp` is the branchiest new file (per-chain `AppNameForChain`, the legacy proxy `%2`-suffix parsing), and those paths look unexercised. A table-driven test that runs `MigrateCoreSettings(Persist)` twice and asserts each key maps correctly, the second run is a no-op, pre-existing settings are not overwritten, defaults are dropped, and a disabled-then-re-enabled proxy round-trips would cover the highest-risk behavior.

qml/datadir.cpp
epicleafies: Same validation path, one more gap: `NormalizeLocalPath` does not absolutize (it returns the input unchanged when it is not a `file:` URL), so a relative saved datadir would be applied as `-datadir` and resolved against the process CWD. Returning `QDir(local).absolutePath()` from `NormalizeLocalPath` closes it (checking against the `__ANDROID__` `content://` handling first).

test/functional/qml_test_proxy.py
epicleafies: Same handler-bypass hazard elsewhere in the bridge: `cmdInvoke` in `testbridge.cpp` resolves any method or signal by name and emits it with `DirectConnection`, so a test can fire `clicked()` directly and pass without the QML handler running. A header note that `invoke` of a signal bypasses handlers would steer tests away from it.

qml/models/options_model.cpp
epicleafies: Security note on the same loop: `thirdPartyTransactionLinks` only checks `host.isEmpty()`, not the scheme, so a configured `file://host/...` URL reaches `Qt.openUrlExternally` (through the ExternalPopup confirm). Allowlisting `http`/`https` here keeps non-web schemes out at the source. The exploitation path is weak (the URL is user-configured and the popup shows it first), so this is hardening, not a blocker; Core opens these with no scheme check or confirmation at all, so it would improve on Core. Separately, the URL is shown in `ExternalPopup` via a `CoreText` defaulting to `Text.AutoText`, so a URL containing markup renders as rich text on the popup; `Text.PlainText` on that display removes that.

qml/components/StorageOptions.qml
jarolrod: its fine for now

qml/models/onboardingoptionsmodel.h
jarolrod: fine for now

test/CMakeLists.txt
jarolrod: The current bitcoinqml_unit_tests link list contains
bitcoinqml, wallet/core libraries, GMock, and Qt test libraries, but no
bitcoinqt. The QML options-model compatibility test no longer includes
<qt/bitcoinunits.h> and instead uses a small local legacy display-unit
enum to verify the same DisplayBitcoinUnit integer mapping. This keeps the
test coverage without introducing a dependency on the legacy Qt Widgets GUI.

test/CMakeLists.txt
jarolrod: should be all addressed in the current stack

qml/pages/node/NetworkTraffic.qml
jarolrod: addressed. The branch no longer uses QML import QtCore for
Settings. The affected QML files were moved back to
import Qt.labs.settings 1.0, which is the Qt 6.4-compatible import path

qml/bitcoin.cpp
jarolrod: addressed. The typed connection was replaced with the string-based
signal/slot form:
QObject::connect(context.window, SIGNAL(closing(QQuickCloseEvent*)), &loop, SLOT(quit())).
The handler only needs to quit the pre-init onboarding event loop, so this
keeps the same behavior without requiring Qt 6.4 to instantiate the
incomplete QQuickCloseEvent type.

test/functional/qml_test_proxy.py
jarolrod: addressed. qml/test/testbridge.cpp now tries real
press/release mouse events on visible, enabled QQuickItems before falling
back to raw signal/method invocation. This keeps functional tests aligned
with the same QML handler path users exercise.

qml/pages/MainWindow.qml
jarolrod: addressed. The onboarding create-wallet flow no longer exposes
the intro Back action after pre-init onboarding has committed startup
settings. Functional wallet helpers were also updated to tolerate the
resulting create-wallet entry shape.

qml/datadir.cpp
jarolrod: addressed. ShouldShowDataDirChooser() now revalidates a saved
non-default GUI datadir with ValidateCustomDataDir() before skipping the
chooser. ApplyGuiDataDirSetting() now stops without soft-setting
-datadir if EnsureDataDir() fails. Unit tests cover both an unusable
saved datadir and an unwritable saved datadir.

qml/models/options_model.cpp
jarolrod: addressed by clarifying and testing the intended behavior.
NAT-PMP remains live-applied because this matches Qt Widgets behavior:
accepting the settings change persists natpmp and calls mapPort() without
requiring restart. The code now documents that NAT-PMP is intentionally
live-applied, and the unit test verifies enable and disable both update the
setting, call mapPort(), and do not mark connection settings dirty or
restart-required.

qml/models/options_model.cpp
jarolrod: addressed. thirdPartyTransactionLinks() now skips configured
URLs that do not contain %s before exposing generated links. The unit test
includes a valid URL missing %s and verifies only the placeholder URL is
returned.

Will evaluate the security comment.

qml/pages/wallet/WalletBadge.qml
jarolrod: addressed. WalletBadge now applies
optionsModel.moneyFont.family and optionsModel.moneyFont.weight to the
balance text, and tst_desktopwallets.qml verifies the binding.

qml/pages/onboarding/OnboardingStorageAmount.qml
jarolrod: addressed. The onStorageSelectionChanged handler now uses a
JavaScript function with formal customStorage and customStorageAmount
parameters.

qml/components/StorageSettings.qml
jarolrod: addressed. SettingsRestartNotice.qml is now a small semantic
wrapper around the shared InfoBanner component instead of custom
Rectangle/Icon/CoreText banner markup.

qml/models/options_model.cpp
jarolrod: addressed by reflowing the stack. DirtySnapshot,
dirtySnapshot(), and emitDirtySignals(const DirtySnapshot& before) are
introduced in the same commit that first uses them, so the intermediate
commit now compiles.

qml/models/options_model.cpp
jarolrod: addressed. Proxy location changes now flow through
QmlCoreSettings::Session::changeProxyLocation(), which trims and validates
the location before mutating m_values.proxy_address. Invalid values return
an unaccepted change and leave the previous address intact.

qml/models/options_model.h
jarolrod: addressed. The final options model uses
emitDirtySignals(const DirtySnapshot& before) and keeps the dirty-state
comparison grouped in DirtySnapshot.

qml/components/ProxySettings.qml
jarolrod: addressed. The branch no longer carries both components. The
existing input was refactored/renamed into
qml/controls/ProxyLocationInput.qml, and the old
IPAddressValueInput.qml file and QRC entry are gone. The new name matches
Core proxy semantics better because valid proxy locations can include
hostnames, host-with-port values, IPv6 forms, and supported unix: paths,
not only IP addresses.

qml/models/options_model.cpp
jarolrod: addressed by the model split. Runtime OptionsQmlModel persists
runtime datadir settings immediately, which is expected for the settings
page. Pre-init onboarding uses OnboardingOptionsModel; its
selectCustomDataDir() only updates onboarding model state and preview
state. The path is ensured/applied during onboarding completion through
QmlOnboardingSettings::ApplyToArgs().

qml/components/StorageLocations.qml
jarolrod: addressed. StorageLocations.qml now uses FolderDialog with a
stable customDataDirFolderDialog object name. QML tests verify the folder
dialog is present and exposes selectedFolder.

qml/bitcoin.cpp
jarolrod: addressed together with the datadir helper review. Startup now
asks ShouldShowDataDirChooser() before applying saved GUI datadir state,
and ApplyGuiDataDirSetting() refuses to apply a saved non-default path if
EnsureDataDir() fails. That prevents an unwritable, invalid, or otherwise
unusable saved path from being silently soft-set as -datadir.

qml/pages/MainWindow.qml
```diff
@@ -35,17 +69,7 @@ ApplicationWindow {
...
-                onboardingWizard
-            } else {
-                if (AppMode.walletEnabled && AppMode.isDesktop) {
-                    desktopWallets
-                } else {
-                    node
-                }
-            }
-        }
+        initialItem: appWindow.desktopWalletMode ? desktopWallets : node
```
epicleafies: Blocker: On Qt 6.4 (the minimum supported version), a first-run user launching headed SIGSEGVs at startup before reaching any interactive page. This PR moves onboarding into the pre-init window and lands the main window on desktopWallets for a no-wallet user, whose StackLayout eagerly instantiates the Send page and its AlertPopups at first layout, so the latent crash now fires at startup for every new user.

This is already fixed by #746 (commit b53ddbbf81 "qml: Fix Qt 6.4 use-after-free in AlertPopup actions", which wraps the actions in a Row positioner), but that fix has not landed on upstream/qt6 yet. So nothing needs to change in this PR right now, but it should rebase off 746 once it is merged.

test/functional/qml_test_preinit_onboarding.py
epicleafies: Still unresolved at 682d2042, neither test is in the workflow. I ran both on this head and both pass, so they only need wiring in.

qml/pages/wallet/ActivityDetails.qml
epicleafies: Still unresolved at 682d2042, both Repeaters (153 and 200) are unchanged. Following the AlertPopup pointer in this comment: AlertPopup.qml on this branch still carries the broken RowLayout + Repeater, not the #746 Row-positioner fix, and it actually SIGSEGVs at first-run startup now (see the new top-level finding). So this is the same live use-after-free, not a theoretical one. Same fix here: wrap the delegates in a Column positioner with explicit widths.

epicleafies (commented):
Re-review at 682d2042. The previous compile blocker is fixed: the branch now builds clean on Qt 6.4.2 and the full offscreen suite is green (unit, QML, 29 functional pass / 2 skip / 0 fail). One new blocker, plus two findings from my earlier review that are still open (followed up on their threads). Couldn't manually test it because of the new blocker.

New blocker: on Qt 6.4, a first-run user launching headed SIGSEGVs at startup before any interactive page. It is fixed in 746, if that gets merged in first can rebase onto it.

Still open from my earlier review (details on each thread): the two new functional tests are not in the CI workflow and the ActivityDetails Repeater-in-Layout is the same now-confirmed crash class.

johnny9 (commented):
commits 662f07b455  through 682d204307 don't compile and require a small patch.

Patch to squash into 662f07b455 (qml: add display settings parity):

```
  diff --git a/qml/bitcoin.cpp b/qml/bitcoin.cpp
  --- a/qml/bitcoin.cpp
  +++ b/qml/bitcoin.cpp
  @@ -58,7 +58,6 @@
   #include <qml/networkstyle.h>
   #include <qml/util.h>
   #include <qml/walletqmlcontroller.h>
  -#include <qt/bitcoinunits.h>
   #ifdef ENABLE_TEST_AUTOMATION
   #include <qml/test/testbridge.h>
   #endif
  @@ -425,7 +424,6 @@ int QmlGuiMain(int argc, char* argv[])

       Q_INIT_RESOURCE(bitcoin_qml);
       qRegisterMetaType<interfaces::BlockAndHeaderTipInfo>("interfaces::BlockAndHeaderTipInfo");
  -    qRegisterMetaType<BitcoinUnit>("BitcoinUnit");

       QGuiApplication::styleHints()->setTabFocusBehavior(Qt::TabFocusAllControls);
       QGuiApplication app(argc, argv);
```

BitcoinUnit doesn't appear to be needed for qml for these changes.

hebasto:
Please rebase.

qml/models/core_settings_model.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+CoreSettingsModel::CoreSettingsModel(QmlCoreSettings::Values values, QVariantMap statuses, QObject* parent)
+    : QObject{parent}
+    , m_session{std::move(values), std::move(statuses)}
+{
+}
+
+QObject* CoreSettingsModel::entry(const QString& key)
+{
+    return ensureEntry(key);
```
pseudoramdom, line 147: In 32a9578bbc qml: expose core settings as entry models

Should this reject unknown keys before calling `ensureEntry()`

qml/components/StorageSettings.qml
```diff
@@ -9,9 +9,30 @@ import "../controls"
 
 ColumnLayout {
     id: root
+    property var settingsModel: optionsModel
+    property var coreSettingsModel: settingsModel.coreSettings
     property bool customStorage: false
     property int customStorageAmount
     property bool showRestartNotice: false
+    readonly property var pruneSetting: coreSettingsModel.entry("prune")
+    readonly property bool hasStorageResult: root.settingsModel && root.settingsModel["storageAvailableText"] !== undefined && root.settingsModel.storageAvailableText.length > 0 && !root.settingsModel.storageCheckPending
```
pseudoramdom, line 18: Should we bundle all the check as one property in the model?

pseudoramdom (commented):
Reviewed till db326f2a04 (qml: show core setting source status in rows)

pseudoramdom:
Concept ACK f029c3e1fa93ffe962e09a8ca08981992c84a2ff
(Code review in progress)

---
Manually tested with isolated datadirs and GUI settings dirs:
- Fresh wallet-enabled onboarding flow
- Restart after completed onboarding
- Fresh node-only onboarding flow with `-disablewallet`
- Runtime startup with `-qml_onboarded=1`
- `-resetguisettings` flow on an onboarded runtime profile
- CLI-sourced startup settings flow. Tried with `-listen`, `-natpmp`, `-server`, `-proxy`
- Config-file-sourced startup settings flow
- Prune startup setting flow

Let me know if I missed any scenario to test.

pseudoramdom:
I can still repro #704. 
I ran with an existing regtest datadir copy containing wallets and no `qml_onboarded` marker. On first launch, onboarding is shown, but it runs through the full fresh-profile wizard instead of the abbreviated existing-profile flow described in the issue. 
After completing onboarding, the existing wallets from the datadir are not surfaced in the app.

qml/datadir.cpp
```diff
@@ -0,0 +1,303 @@
...
+    const QString validation_error = ValidateCustomDataDir(local_path);
+    if (!validation_error.isEmpty()) {
+        if (error) *error = validation_error;
+        return false;
+    }
+
+    try {
+        const fs::path data_dir_path = QStringToPath(local_path);
+        TryCreateDirectories(data_dir_path);
+        TryCreateDirectories(data_dir_path / "wallets");
```
johnny9: This is not safe for an existing profile. `EnsureDataDir()` always creates `<datadir>/wallets` here. Core uses that directory's existence to switch wallet discovery away from the datadir root, so a valid top-level wallet can disappear from the UI after onboarding even though the wallet files are still on disk. Please preserve the existing wallet discovery behavior instead of unconditionally creating `wallets/` for an existing datadir.

qml/onboarding_settings.cpp
```diff
@@ -0,0 +1,400 @@
...
+    } catch (const std::exception& e) {
+        result.error = QString::fromStdString(e.what());
+        return result;
+    }
+
+    const bool explicit_datadir = HasExplicitDataDirArg(preview_args);
+    const bool custom_datadir = !explicit_datadir && !QmlDataDir::IsDefaultDataDir(data_dir);
+    const bool custom_datadir_exists = custom_datadir && QFileInfo::exists(data_dir);
+    if (custom_datadir_exists) {
+        QmlDataDir::ApplyDataDirArg(preview_args, data_dir);
```
johnny9, line 407: This breaks Core's config-loading invariant. Preview soft-sets the resolved datadir before `ReadConfigFiles()`, and the same ordering is repeated in `ApplyToArgs()` below. If the default `bitcoin.conf` contains `datadir=`, the next config read comes from the new datadir instead of the original config source, so the rest of the config can be silently dropped and the wrong network can start. Please keep the original config source stable when `datadir` is changed from config.

qml/models/onboardingoptionsmodel.cpp
```diff
@@ -0,0 +1,351 @@
...
+    if (!m_storage_result_valid || !m_storage_error_text.isEmpty()) return;
+
+    const QVariantMap prune_status = m_core_settings.statuses().value(QStringLiteral("prune")).toMap();
+    if (!prune_status.value(QStringLiteral("canEdit"), true).toBool()) return;
+    if (prune_status.value(QStringLiteral("source")).toString() != QStringLiteral("default")) return;
+
+    const bool user_touched_prune = m_core_settings.isTouched(QStringLiteral("prune")) && !m_prune_auto_recommended;
+    if (user_touched_prune) return;
+
+    const bool should_prune = QmlOnboardingStorage::ShouldRecommendPrune(storageState());
```
johnny9, line 298: This recommendation path is not safe for an existing profile. It compares free space against a fresh-chain requirement and then marks `prune` touched, so an existing full node can be turned into a pruned node just by completing onboarding, even without the user explicitly choosing pruning. Existing blocks/chainstate need to count toward capacity here, and prune should not be auto-persisted for an existing node.

qml/bitcoin.cpp
```diff
@@ -314,6 +458,53 @@ int QmlGuiMain(int argc, char* argv[])
...
+            InitError(Untranslated(reset_error.toStdString()));
+            return EXIT_FAILURE;
+        }
+    }
+
+    LoadFontResource(":/fonts/bitcoincoresans/regular");
+    LoadFontResource(":/fonts/bitcoincoresans/semibold");
+    LoadFontResource(":/fonts/robotomono/regular");
+
+    AppMode app_mode = SetupAppMode();
```
johnny9, line 480: `disablewallet=1` from `bitcoin.conf` is still broken here. `AppMode` is resolved before config parsing, so a config-only `disablewallet` is missed. The wallet controller later sees the final setting and never initializes, while the main window keeps waiting for wallet startup forever. Please resolve wallet mode after `InitConfig()` so config-sourced `disablewallet` routes to the node UI instead of a permanent loading state.

johnny9 (changes requested):
Request changes. Left 4 blocking comments for existing-profile safety and config handling.

qml/models/core_settings_model.cpp
jarolrod: I looked into this again. I don’t think this needs a code change in this PR.

`CoreSettingsModel::entry()` is only called from repo-owned QML with static keys. There is no user-provided or dynamic path that can pass arbitrary keys into this API.

If an unsupported key were introduced by a future QML change, it would also fail inertly rather than persist anything: the entry returns empty/default values, setters do not match any mutation path, and session writes reject names outside the supported session setting set.

So I agree this could be a useful defensive guardrail for catching future developer mistakes, but I don’t think it is a correctness issue in the current branch. I’m leaving it asis for now.

uqlidi (changes requested):

qml/models/core_settings_model.cpp
```diff
@@ -0,0 +1,301 @@
...
+    } else if (m_key == QStringLiteral("server")) {
+        m_model.changeServer(enabled);
+    } else if (m_key == QStringLiteral("prune")) {
+        m_model.changePrune(enabled);
+    } else if (m_key == QStringLiteral("proxy")) {
+        m_model.changeProxyEnabled(enabled);
+    } else if (m_key == QStringLiteral("onion")) {
+        m_model.changeTorEnabled(enabled);
+    }
+}
```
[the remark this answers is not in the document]
jarolrod: I agree this is cleaner, but I want to have a discussion on the model itself at a later time. I have a few better options to implement on their own, we need to get this across the line now

qml/models/options_model.cpp
jarolrod: Addressed in the current branch.

`thirdPartyTransactionLinks()` now skips entries that do not contain `%s`, rejects non-HTTP(S) schemes, and only exposes generated links with a host. The external-link confirmation popup also renders the configured URL as `Text.PlainText`, so user-configured URL text is not interpreted as rich text.

Coverage was added in `OptionsModelTests::thirdPartyTransactionLinksParseValidUrls()` for valid HTTP(S) URLs, missing `%s`, malformed URLs, and non-web schemes.

qml/pages/wallet/ActivityDetails.qml
jarolrod: Addressed in the current branch.

Both dynamic `ActivityDetails` repeater sections now live inside plain `Column` positioners instead of being direct layout-managed repeater delegates. Delegates are sized explicitly with the section width, matching the Qt 6.4-safe pattern used for the same crash class elsewhere.

I also added QML component coverage for the third-party-link section so we verify the section lays out with non-zero content and collapses when no links are configured.

test/functional/qml_test_preinit_onboarding.py
jarolrod: Addressed.

`.github/workflows/gui-functional-tests.yml` now runs both standalone tests:

- `test/functional/qml_test_preinit_onboarding.py`
- `test/functional/qml_test_resetguisettings.py`

These are invoked without shared socket arguments, matching the way the tests self-launch the app.

qml/pages/MainWindow.qml
jarolrod: Addressed by bringing the Qt 6.4-safe popup action layout into this branch.

`AlertPopup.qml` now uses a plain `Row` positioner for repeated action buttons instead of handing repeater delegates directly to a Quick Layout. This avoids the Qt 6.4 use-after-free class described here.

qml/components/StorageSettings.qml
jarolrod: leaving as is for now, but we can revisit soon

qml/datadir.cpp
jarolrod: Addressed in the current branch.

Onboarding now preserves existing wallet discovery behavior. It only creates the network `wallets/` subdirectory when the selected data directory or network data directory is newly created. Existing profiles that store wallets directly under the network datadir no longer get switched into `wallets/` discovery by onboarding.

There is unit coverage for new-vs-existing datadir behavior and functional coverage for preserving an existing root wallet through pre-init onboarding.

qml/onboarding_settings.cpp
jarolrod: Addressed.

The onboarding flow now tracks the source of the active datadir. Datadirs resolved from `bitcoin.conf` are treated as config-owned, not GUI-owned. Preview/apply only soft-set and persist `-datadir` when the path comes from GUI settings, legacy GUI settings, or an explicit user selection in onboarding.

This preserves Core’s config-loading source: a `datadir=` value from bitcoin.conf is not copied into QSettings and is not reapplied as a GUI-owned override.

qml/models/onboardingoptionsmodel.cpp
jarolrod: Addressed.

Existing profiles now bypass the automatic prune recommendation path. For an existing profile, onboarding reflects the effective prune state from config/settings/CLI and does not mark prune touched just because the fresh-chain storage recommendation would prefer pruning.

Storage validation for existing profiles also uses the existing-profile storage policy rather than fresh-chain download requirements.

qml/bitcoin.cpp
jarolrod: Addressed.

The app now refreshes `AppMode.walletEnabled` after Core config/settings have been loaded, so `disablewallet=1` from `bitcoin.conf` or `settings.json` is reflected in initial QML routing.

The functional test covers:
- command-line `-disablewallet`
- `disablewallet=1` from bitcoin.conf
- `disablewallet=true` from settings.json
- command-line `-disablewallet=0` overriding bitcoin.conf

qml/bitcoin.cpp
```diff
@@ -314,6 +463,53 @@ int QmlGuiMain(int argc, char* argv[])
...
+        break;
+    case PreInitOnboardingStatus::CANCELED:
+        return EXIT_SUCCESS;
+    case PreInitOnboardingStatus::FAILED:
+        return EXIT_FAILURE;
+    case PreInitOnboardingStatus::NOT_SHOWN:
+        break;
+    }
+
     if (auto error = common::InitConfig(
```
johnny9, line 513: 1. App starts with clean gArgs.
  2. There is no -conf, so Core’s default config location is used:
     ~/.bitcoin/bitcoin.conf

  3. That default config contains:

  datadir=/tmp/B

  4. Pre-init onboarding runs before the normal main-window startup.
  5. During onboarding, ApplyToArgs(gArgs) reads that default config and applies datadir=/tmp/B into the live global args.
  6. Onboarding finishes.
  7. Then normal startup continues and calls InitConfig(gArgs) again.
  8. But now gArgs already has datadir=/tmp/B, so the default config lookup is no longer ~/.bitcoin/bitcoin.conf; it shifts to /tmp/B/bitcoin.conf.
  9. If /tmp/B/bitcoin.conf does not exist, or does not contain the same setup, startup no longer sees the config it used during onboarding.
  10. The app fails/returns during startup before creating the main window.

qml/bitcoin.cpp
```diff
@@ -314,6 +463,53 @@ int QmlGuiMain(int argc, char* argv[])
     ApplyTestSettingsDir();
 #endif
 
+    app.setQuitOnLastWindowClosed(false);
+    setupChainQSettings(&app, QString::fromStdString(gArgs.GetChainTypeString()).toUpper());
+    if (gArgs.GetBoolArg("-resetguisettings", false)) {
```
johnny9, line 468: `-resetguisettings` is applied too early.

  Right now the app clears GUI settings before `bitcoin.conf` has been read and before the active datadir/profile is known. If the default `bitcoin.conf` redirects to `B/
  regtest`, startup still resets the default/main profile first just by opening onboarding. Finishing onboarding later resets `B`, but canceling leaves only the wrong profile
  modified.

  Please defer the reset until after the active profile has been resolved, and add regression coverage for both canceling and completing onboarding when the default config
  redirects to another datadir.

qml/bitcoin.cpp
johnny9: A first glance this doesn't seem to cause any issues but I think its worth double checking because its odd that the application doesn't come up after finishing onboarding.

epicleafies:

Aborted (core dumped)

Repro:

Suggested fix, two parts:

johnny9 (approved):
ACK ff0292f6467cff2d4bc24e5681659d6c4723fa5b

  I re-reviewed the latest updates and confirmed the previous blockers have been addressed. The existing-profile onboarding, config-owned datadir handling, reset/
  settings behavior, disablewallet startup routing, Activity details display unit, Qt 6.4 layout/startup regressions, and functional-test registration all look covered now.

  I opened follow-up issues for remaining startup/reset edge cases that should not block this PR:
  - #773: Pre-init onboarding can reread config after mutating startup args
  - #774: -resetguisettings can reset the wrong profile before datadir resolution
  - #775: Unable to load unreadable custom datadir

johnny9:
> On Linux, if the saved GUI-selected custom datadir still exists but is no longer readable, the app aborts at launch with an unhandled exception instead of re-showing the datadir chooser:
> 

Created #775 to track this. Lets not block on this bug right now.

PR #746, opened 2026-06-04
Desktop tray icon and rpc console

### LINKED ISSUE #508: RPC Command Console QML Surface

### LINKED ISSUE #513: Desktop Tray and Window Behavior Settings

### LINKED ISSUE #685: Settings: use desktop sidebar layout instead of drill-down navigation
## Design Source
- Source design issue: https://github.com/BitcoinDesign/Bitcoin-Core-App/issues/163
- Figma reference: https://www.figma.com/design/fmpjcEftcnyXudd4cbce20/Bitcoin-Core-App-Design?node-id=15152-10698&t=a1mVyYUeUdxd6nfM-4

## Problem
The current Settings drill-down list is suitable for mobile, but it does not use desktop horizontal space well. As Settings grows to include more subsections, desktop users need a persistent navigation structure that reduces transitions and keeps context visible.

## Implementation Scope
- Add a desktop/wide-window Settings shell that uses a persistent sidebar or tabbed sidebar navigation instead of the mobile drill-down list.
- Group related Settings pages in the sidebar so the menu remains scannable as subsections grow.
- Reuse existing Settings page components where possible while allowing separate desktop and mobile layout shells.
- Keep the existing mobile/narrow Settings drill-down layout working.
- Collapse desktop-only nested drill-downs into the main content pane where practical, using inline controls such as selectors instead of extra navigation levels.
- Keep the sidebar in a stable position while selecting topics or resizing the window.
- Use a bounded content column as the default desktop treatment, with wider handling only where a page genuinely needs it.

## Acceptance Criteria
1. Desktop/wide Settings opens with persistent sidebar navigation and a content pane.
2. Selecting a Settings topic changes the content pane without moving the sidebar or pushing the user through the mobile drill-down stack.
3. Related Settings pages are grouped in the sidebar.
4. Mobile/narrow Settings keeps the existing drill-down interaction.
5. Existing Settings read/write and persistence behavior continues to work.
6. Layout behavior is covered by focused QML tests or documented manual visual coverage where automated coverage is not practical.

## Out of Scope
- Moving overloaded Settings content, such as network graphs, debug log, or wallet overview surfaces, into other application areas.
- Redesigning individual Settings pages beyond the changes needed to fit the desktop sidebar content pane.

### LINKED ISSUE #688: Move console out of settings
Add a console icon to the navigation bar to quickly access the command view and move the page out of the settings list. 

Image

### LINKED ISSUE #747: Move peers out of settings
Add a peers icon to the navigation bar to quickly access the command view and move the page out of the settings list.

Image

epicleafies:
Combines the tray icon PR, the rpc console PR and the settings redesign, closes #538 and closes #540 

For the tray icon, fixes #513:
Added window behavior as a new option in settings. 
Added window behavior page, allows the user to configure if the tray is visible, behavior on minimization and behavior on closing.
Tray icon color matches the network the app is on.
Tray icon gives option to show or quit when clicked on.
Added unit, qml and functional tests.

For the rpc console, fixes #508:
Created page for rpc command console.
Created option under settings for console.
Preserved parsing logic currently in use along with autocomplete.
Added unit and functional tests.

For the settings redesign, fixes #685, fixes #688, fixes #747:
The settings menu now uses use a desktop sidebar layout instead of drill-down navigation.
Introduced a reusable SettingsHeader component, replacing NavigationBar2 + state machine boilerplate across all settings pages.
Removed redundant Dark Mode toggle from Developer Options, already exists in Display -> Theme.
Rpc console and peers now have their own icons on the main navbar and have been moved out of the settings page. 
Added unit, qml and functional tests.
Updated functional tests to support rpc console and peers being out of the settings page.

Fixes from #750:
Fixes send page nav button crash on Qt 6.4.2.
Added qml_test_psbt.py to CI workflow file.

Also cleaned up a few minor style/design issues.

Platform support. The tray-icon window toggle uses an obscured-check (QGuiApplication::topLevelAt()) ported from the Qt Widgets GUI's GUIUtil::isObscured(). On Wayland the compositor forbids querying other windows, so obscured-detection is unavailable, showMinimized() does not report a Minimized visibility, and the window position is not preserved across a tray hide/show because the compositor, not the client, controls placement. The position limitation applies to any Wayland-compositor session, including X11 clients via Xwayland; geometry is restored correctly on native X11 (Xorg), macOS, and Windows. These match the limitations the existing Qt Widgets GUI has on Wayland and meet issue #513's DoD, which does not require obscured-detection parity. Tracking: bitcoin/bitcoin#19950 and #748.

Qt Quick has no system-tray API, so DesktopTrayIconController wraps QSystemTrayIcon (Qt Widgets) in C++, with a native QMenu context menu (Linux SNI/AppIndicator hosts require a menu set via setContextMenu()). This reuses the Qt Widgets dependency the GUI already links.

image

image


epicleafies:
Note main navbar can't fit anymore elements, adding the rpc console and peers icon fill it up at max width

qml/pages/settings/SettingsAbout.qml
```diff
@@ -25,41 +26,12 @@ InformationPage {
...
-    Component {
-        id: header
-        Header {
-            headerBold: true
-            headerSize: 18
-            header: qsTr("About")
-        }
+    showNavBar: false
+    header: SettingsHeader {
+        title: root.onboarding ? "" : qsTr("About")
```
johnny9, line 31: I think these headers aren't needed if we have the context already on the left sidebar.

johnny9 (commented):
```
FAILED: wait_for_page('gotoAboutSetting') failed: Timed out waiting for page: gotoAboutSetting
Traceback (most recent call last):
  File "/home/runner/work/gui-qml/gui-qml/test/functional/qml_test_blocksonly_settings.py", line 26, in run_tests
    gui.wait_for_page("gotoAboutSetting", timeout_ms=10000)
  File "/home/runner/work/gui-qml/gui-qml/test/functional/qml_driver.py", line 189, in wait_for_page
    raise QmlDriverError(
qml_driver.QmlDriverError: wait_for_page('gotoAboutSetting') failed: Timed out waiting for page: gotoAboutSetting
Starting GUI: /home/runner/work/gui-qml/gui-qml/build/bin/bitcoin-core-app -datadir=/tmp/qml_test_bridge_cbg9940j/node0 -test-automation=/tmp/qml_test_bridge_cbg9940j/test_bridge.sock -test-settings-dir=/tmp/qml_test_bridge_cbg9940j/config -resetguisettings -logtimemicros -debug -debugexclude=libevent -debugexclude=leveldb -nolisten -disablewallet -blocksonly -disablewallet -blocksonly
QmlDriver connected to test bridge.
```

Node only mode functional tests are failing. `-disablewallet` cli arg will launch it in that mode.

qml/pages/settings/SettingsAbout.qml
epicleafies: I can take them out if you want but without a full page redesign it doesn't look good. If I did do it I'd probably align the page with top of menu. 
image

image

pseudoramdom: I'd prefer to keep the header title.

qml/res/icons/node-1-connection.png
johnny9, line 1: The peers icon here is designed to be 5 icons with each one representing an increasing amount of connections to the node. 

Image

johnny9 (approved):
There should be 5 peer/node icons

qml/res/icons/node-1-connection.png
johnny9: The original source of these icon is the bitcoin wallet ui kit https://www.figma.com/design/iaIQLsKeyK6CDRilTPeHie/Bitcoin-Wallet-UI-Kit---Design-System--Community-?node-id=4499-57506&t=Kq8aiiPsn58BlnQ1-4

pseudoramdom: It's available on https://bitcoinicons.com

epicleafies: Okay. I'll update to use all dynamically.

pseudoramdom:
The command hints are super slick and useful! But when I hit return on any of the command, I'd expect the selected command to go through , but instead what ever is in my textfield went.
https://github.com/user-attachments/assets/f24d2c71-65c2-498f-b813-3a8f21a13f6d

pseudoramdom:
Re: Settings layout, I believe Christoph had a simpler version [here](https://github.com/BitcoinDesign/Bitcoin-Core-App/issues/163#issuecomment-4368468139)
Mainly
- highlight over selected item instead of the accent
- Grouping of items
- left aligned

pseudoramdom:
The icon sizes are off on macOS. Worth fixing it here if it's quick
Screenshot 2026-06-08 at 12 12 31 PM

qml/models/rpcconsolemodel.cpp
```diff
@@ -0,0 +1,439 @@
...
+                                      "               getblockhash,0\n\n"
+                                      "Named results can be queried with a non-quoted key string in brackets.\n"
+                                      "   example:    getblock(getblockhash(0) 1)[tx]\n\n"
+                                      "Results without keys can be queried with an integer in brackets.\n"
+                                      "   example:    getblock(getblockhash(0),1)[tx][0]\n\n"));
+                return;
+            }
+
+            std::string result;
+            if (!RpcCommandExecutor::RPCExecuteCommandLine(m_node, result, executableCommand)) {
```
jarolrod: Are we intentionally choosing to drop the wallet context here?

The lower-level executor already supports wallet scoping: `RpcCommandExecutor::RPCParseCommandLine()` accepts `wallet_name` and builds the `/wallet/<percent-encoded-name>` URI before calling `node.executeRpc()`.

qml/models/rpcconsolemodel.cpp
```diff
@@ -0,0 +1,439 @@
...
+        }
+    }
+    m_history_idx = -1;
+    m_pending_text.clear();
+
+    // Append the request line immediately (main thread) so QML sees it before
+    // the async reply arrives. Mirrors Qt GUI RPCConsole behaviour.
+    appendFormattedRow(time, CMD_REQUEST, filteredCmd);
+
+    setExecuting(true);
```
jarolrod, line 399: The button itself is disabled while `rpcConsoleModel.executing` is true, but `submitCommand()` itself still accepts commands. That matters because Enter calls the QML `submitCommand()` helper directly, so keyboard submission can still reach the invokable while another RPC is running.

Qt Widgets handles this in `RPCConsole::on_lineEdit_returnPressed()` by allowing `stop` as a special case, then returning early when `m_is_executing` is true. 

The QML worker still executes queued commands serially, so this is not parallel RPC execution, but the `executing` state can become inaccurate: the first result calls `setExecuting(false)` even if another command is already queued or running behind it.

jarolrod: The suggestion here is to model Qt Widgets with how it prevents calls from all forms, button and keyboard press

qml/models/desktopwindowbehaviormodel.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <qml/models/desktopwindowbehaviormodel.h>
+
+#include <QSettings>
+
+void DesktopWindowBehaviorModel::loadFromSettings()
+{
+    QSettings settings;
+    // fHideTrayIcon stores the inverse of showTrayIcon (default: show = true → hide = false)
+    m_show_tray_icon = !settings.value(KEY_HIDE_TRAY_ICON, false).toBool();
+    m_minimize_to_tray = settings.value(KEY_MINIMIZE_TO_TRAY, false).toBool();
```
jarolrod: Maybe we should mirror the Qt Widgets load time clamp here?

Widgets has treated persisted settings as mutable external input in a few places:
- malformed proxy QSettings were reset to avoid Options dialog crashes ([ce2418fa4c](https://github.com/bitcoin/bitcoin/pull/11448))
- invalid DisplayBitcoinUnit values are rewritten to BTC ([75832fdc37](https://github.com/bitcoin-core/gui/pull/556))
- bad settings.json values get recovery handling ([1ee6d0b01a](https://github.com/bitcoin-core/gui/pull/379), [5b1aae12ca](https://github.com/bitcoin/bitcoin/pull/24498))
- coupled settings like listen/listenonion are reconciled during init ([70f2c579b9](https://github.com/bitcoin-core/gui/pull/568), [3b82608dd1](https://github.com/bitcoin-core/gui/pull/568))

The tray setting has the same precedent: since the original hide-tray feature ([8b0e497028](https://github.com/bitcoin/bitcoin/pull/8006)) and still after the positive “Show tray icon” rename ([17174f8328](https://github.com/bitcoin-core/gui/pull/115)), Widgets loads minimizeToTray as settings.value("fMinimizeToTray").toBool() && m_show_tray_icon.

An invalid state is not reachable through normal user clicks, but the gui is implemented to be defensive against an old/manually altered/stale QSettings value so that it cannot produce runtime `showTrayIcon=false && minimizeToTray=true`.

jarolrod: The needed changes would look something like 

<details>

```diff
diff --git a/qml/models/desktopwindowbehaviormodel.cpp b/qml/models/desktopwindowbehaviormodel.cpp
index 5e8d2ed049..b8c74ae04e 100644
--- a/qml/models/desktopwindowbehaviormodel.cpp
+++ b/qml/models/desktopwindowbehaviormodel.cpp
@@ -11,7 +11,11 @@ void DesktopWindowBehaviorModel::loadFromSettings()
     QSettings settings;
     // fHideTrayIcon stores the inverse of showTrayIcon (default: show = true → hide = false)
     m_show_tray_icon = !settings.value(KEY_HIDE_TRAY_ICON, false).toBool();
-    m_minimize_to_tray = settings.value(KEY_MINIMIZE_TO_TRAY, false).toBool();
+    const bool raw_minimize_to_tray = settings.value(KEY_MINIMIZE_TO_TRAY, false).toBool();
+    m_minimize_to_tray = raw_minimize_to_tray && m_show_tray_icon;
+    if (raw_minimize_to_tray && !m_show_tray_icon) {
+        settings.setValue(KEY_MINIMIZE_TO_TRAY, false);
+    }
     m_minimize_on_close = settings.value(KEY_MINIMIZE_ON_CLOSE, false).toBool();
 
     if (!m_desktop_platform) {
diff --git a/test/test_desktopwindowbehaviormodel.cpp b/test/test_desktopwindowbehaviormodel.cpp
index a25a07ad89..664093eac7 100644
--- a/test/test_desktopwindowbehaviormodel.cpp
+++ b/test/test_desktopwindowbehaviormodel.cpp
@@ -33,6 +33,7 @@ private Q_SLOTS:
     void shouldMinimizeWindowOnClose_requiresDesktopAndMinimizeOnClose();
 
     // QSettings persistence
+    void loadSettings_clampsMinimizeToTrayWhenTrayHidden();
     void settings_persistAcrossInstances();
 
 private:
@@ -184,6 +185,23 @@ void DesktopWindowBehaviorModelTests::shouldMinimizeWindowOnClose_requiresDeskto
     }
 }
 
+void DesktopWindowBehaviorModelTests::loadSettings_clampsMinimizeToTrayWhenTrayHidden()
+{
+    QSettings settings;
+    settings.setValue("fHideTrayIcon", true);
+    settings.setValue("fMinimizeToTray", true);
+    settings.sync();
+
+    DesktopWindowBehaviorModel model(/*desktop_platform=*/true);
+
+    QCOMPARE(model.showTrayIcon(), false);
+    QCOMPARE(model.minimizeToTray(), false);
+    QCOMPARE(model.shouldHideToTrayOnMinimize(), false);
+
+    QSettings updated_settings;
+    QCOMPARE(updated_settings.value("fMinimizeToTray", true).toBool(), false);
+}
+
 void DesktopWindowBehaviorModelTests::settings_persistAcrossInstances()
 {
     // minimizeOnClose persists when tray icon stays enabled
```

</details>

Alternatively I'm fine with this being a follow-up, or if this is merged before settings, I can include it in my settings PR.

qml/models/rpcconsolemodel.cpp
jarolrod: a fix for this could look like: 

<details>

```diff
diff --git a/qml/models/rpcconsolemodel.cpp b/qml/models/rpcconsolemodel.cpp
index 6fa98e01a5..48937fdfae 100644
--- a/qml/models/rpcconsolemodel.cpp
+++ b/qml/models/rpcconsolemodel.cpp
@@ -203,7 +203,7 @@ public:
         : QObject(parent), m_node(node) {}
 
 public Q_SLOTS:
-    void execute(const QString& command)
+    void execute(const QString& command, const QString& wallet_name)
     {
         QString time = QDateTime::currentDateTime().toString("hh:mm:ss");
         try {
@@ -232,7 +232,7 @@ public Q_SLOTS:
             }
 
             std::string result;
-            if (!RpcCommandExecutor::RPCExecuteCommandLine(m_node, result, executableCommand)) {
+            if (!RpcCommandExecutor::RPCExecuteCommandLine(m_node, result, executableCommand, nullptr, wallet_name)) {
                 Q_EMIT resultReady(time, RpcConsoleModel::CMD_ERROR,
                                    tr("Parse error: unbalanced ' or \""));
                 return;
@@ -325,7 +325,7 @@ void RpcConsoleModel::appendFormattedRow(const QString& time, int category, cons
                              WrapColor(prefix + body, color));
 }
 
-void RpcConsoleModel::submitCommand(const QString& command)
+void RpcConsoleModel::submitCommand(const QString& command, const QString& wallet_name)
 {
     if (command.trimmed().isEmpty()) return;
 
@@ -359,6 +359,15 @@ void RpcConsoleModel::submitCommand(const QString& command)
     m_history_idx = -1;
     m_pending_text.clear();
 
+    if (m_last_wallet_name != wallet_name) {
+        if (!wallet_name.isEmpty()) {
+            appendFormattedRow(time, CMD_REQUEST, tr("Executing command using \"%1\" wallet").arg(wallet_name));
+        } else {
+            appendFormattedRow(time, CMD_REQUEST, tr("Executing command without any wallet"));
+        }
+        m_last_wallet_name = wallet_name;
+    }
+
     // Append the request line immediately (main thread) so QML sees it before
     // the async reply arrives. Mirrors Qt GUI RPCConsole behaviour.
     appendFormattedRow(time, CMD_REQUEST, filteredCmd);
@@ -367,7 +376,8 @@ void RpcConsoleModel::submitCommand(const QString& command)
 
     // Dispatch to worker thread.
     QMetaObject::invokeMethod(m_worker, "execute", Qt::QueuedConnection,
-                              Q_ARG(QString, command));
+                              Q_ARG(QString, command),
+                              Q_ARG(QString, wallet_name));
 }
 
 QString RpcConsoleModel::browseHistory(int direction, const QString& currentText)
diff --git a/qml/models/rpcconsolemodel.h b/qml/models/rpcconsolemodel.h
index 8ff8a11e53..840042c440 100644
--- a/qml/models/rpcconsolemodel.h
+++ b/qml/models/rpcconsolemodel.h
@@ -73,7 +73,7 @@ private:
  *        emitted rows keep their baked-in colours, which is intentional).
  *
  * Invokable from QML:
- *  - submitCommand(command)
+ *  - submitCommand(command, walletName)
  *  - browseHistory(direction, currentText) → QString
  *  - resetHistoryNavigation()
  *  - clear()
@@ -105,7 +105,7 @@ public:
     QStringList availableCommands() const { return m_available_commands; }
     QAbstractListModel* outputModel() { return &m_output_model; }
 
-    Q_INVOKABLE void submitCommand(const QString& command);
+    Q_INVOKABLE void submitCommand(const QString& command, const QString& wallet_name = {});
 
     /**
      * Navigate command history.
@@ -151,6 +151,7 @@ private:
     QStringList m_history;
     int m_history_idx{-1};
     QString m_pending_text; // text being edited before browsing history
+    QString m_last_wallet_name;
 
     QThread m_worker_thread;
     RpcConsoleWorker* m_worker{nullptr};
diff --git a/qml/pages/node/CommandConsole.qml b/qml/pages/node/CommandConsole.qml
index 0806885676..d99dc353ac 100644
--- a/qml/pages/node/CommandConsole.qml
+++ b/qml/pages/node/CommandConsole.qml
@@ -41,6 +41,8 @@ Page {
     // Exposed for E2E execution-complete detection.
     property bool executing: rpcConsoleModel.executing
 
+    property string walletName: ""
+
     // Guard flag: true while Up/Down navigation is setting inputField.text
     // programmatically, so onTextChanged does not reset the history cursor.
     // Kept in a QtObject to avoid leaking this internal flag into the public API.
@@ -404,7 +406,7 @@ Page {
     function submitCommand() {
         var cmd = inputField.text.trim()
         if (cmd.length === 0) return
-        rpcConsoleModel.submitCommand(cmd)
+        rpcConsoleModel.submitCommand(cmd, root.walletName)
         inputField.text = ""
         filteredCommands = []
     }
diff --git a/qml/pages/wallet/DesktopWallets.qml b/qml/pages/wallet/DesktopWallets.qml
index 7dd931d68e..2c31a731ec 100644
--- a/qml/pages/wallet/DesktopWallets.qml
+++ b/qml/pages/wallet/DesktopWallets.qml
@@ -339,6 +339,9 @@ Page {
         }
         CommandConsole {
             showHeader: false
+            walletName: walletController.isWalletLoaded && walletController.selectedWallet
+                ? walletController.selectedWallet.name
+                : ""
         }
         NodeSettings {
             id: nodeSettings
```

</details>

qml/models/rpcconsolemodel.cpp
jarolrod: a fix for the above can look like:

<details>

```diff
diff --git a/qml/models/rpcconsolemodel.cpp b/qml/models/rpcconsolemodel.cpp
index 6fa98e01a5..fcea7627a3 100644
--- a/qml/models/rpcconsolemodel.cpp
+++ b/qml/models/rpcconsolemodel.cpp
@@ -325,9 +325,10 @@ void RpcConsoleModel::appendFormattedRow(const QString& time, int category, cons
                              WrapColor(prefix + body, color));
 }
 
-void RpcConsoleModel::submitCommand(const QString& command)
+bool RpcConsoleModel::submitCommand(const QString& command)
 {
-    if (command.trimmed().isEmpty()) return;
+    const QString trimmed_command = command.trimmed();
+    if (trimmed_command.isEmpty()) return false;
 
     // Compute the filtered version to store in history (passwords redacted).
     // Mirrors Qt5's rpcconsole.cpp:994-1004: the fExecute=false parse call must be
@@ -337,18 +338,27 @@ void RpcConsoleModel::submitCommand(const QString& command)
     std::string dummy;
     QString time = QDateTime::currentDateTime().toString("hh:mm:ss");
     try {
-        if (!RpcCommandExecutor::RPCParseCommandLine(nullptr, dummy, command.toStdString() + "\n",
+        if (!RpcCommandExecutor::RPCParseCommandLine(nullptr, dummy, trimmed_command.toStdString() + "\n",
                                                      false, &filtered)) {
             appendFormattedRow(time, CMD_ERROR, tr("Parse error: unbalanced ' or \""));
-            return;
+            return true;
         }
     } catch (const std::runtime_error& e) {
         appendFormattedRow(time, CMD_ERROR,
                            tr("Error: %1").arg(QString::fromStdString(e.what())));
-        return;
+        return true;
     }
     QString filteredCmd = QString::fromStdString(filtered).trimmed();
 
+    // Match Qt Widgets: allow "stop" even while another command is executing,
+    // but do not enqueue other commands until the current reply has arrived.
+    if (trimmed_command == QLatin1String("stop")) {
+        std::string result;
+        RpcCommandExecutor::RPCExecuteCommandLine(m_node, result, trimmed_command.toStdString());
+        return true;
+    }
+    if (m_executing) return false;
+
     // Add to history (deduplicate consecutive identical entries).
     if (m_history.isEmpty() || m_history.last() != filteredCmd) {
         m_history.append(filteredCmd);
@@ -367,7 +377,8 @@ void RpcConsoleModel::submitCommand(const QString& command)
 
     // Dispatch to worker thread.
     QMetaObject::invokeMethod(m_worker, "execute", Qt::QueuedConnection,
-                              Q_ARG(QString, command));
+                              Q_ARG(QString, trimmed_command));
+    return true;
 }
 
 QString RpcConsoleModel::browseHistory(int direction, const QString& currentText)
diff --git a/qml/models/rpcconsolemodel.h b/qml/models/rpcconsolemodel.h
index 8ff8a11e53..e9787b9056 100644
--- a/qml/models/rpcconsolemodel.h
+++ b/qml/models/rpcconsolemodel.h
@@ -73,7 +73,7 @@ private:
  *        emitted rows keep their baked-in colours, which is intentional).
  *
  * Invokable from QML:
- *  - submitCommand(command)
+ *  - submitCommand(command) -> bool
  *  - browseHistory(direction, currentText) → QString
  *  - resetHistoryNavigation()
  *  - clear()
@@ -105,7 +105,12 @@ public:
     QStringList availableCommands() const { return m_available_commands; }
     QAbstractListModel* outputModel() { return &m_output_model; }
 
-    Q_INVOKABLE void submitCommand(const QString& command);
+    /**
+     * Submit a command for execution.
+     * @return true if the command was accepted, false if it was ignored because
+     *         another command is already executing.
+     */
+    Q_INVOKABLE bool submitCommand(const QString& command);
 
     /**
      * Navigate command history.
diff --git a/qml/pages/node/CommandConsole.qml b/qml/pages/node/CommandConsole.qml
index 0806885676..e9c780d385 100644
--- a/qml/pages/node/CommandConsole.qml
+++ b/qml/pages/node/CommandConsole.qml
@@ -310,7 +310,8 @@ Page {
                 Layout.preferredHeight: 32
                 Layout.alignment: Qt.AlignVCenter
                 hoverEnabled: true
-                enabled: !rpcConsoleModel.executing && inputField.text.trim().length > 0
+                enabled: inputField.text.trim().length > 0 &&
+                         (!rpcConsoleModel.executing || inputField.text.trim() === "stop")
                 onClicked: submitCommand()
 
                 background: Rectangle {
@@ -404,8 +405,9 @@ Page {
     function submitCommand() {
         var cmd = inputField.text.trim()
         if (cmd.length === 0) return
-        rpcConsoleModel.submitCommand(cmd)
-        inputField.text = ""
-        filteredCommands = []
+        if (rpcConsoleModel.submitCommand(cmd)) {
+            inputField.text = ""
+            filteredCommands = []
+        }
     }
 }
```

</details>

jarolrod: To reproduce manaully, you can enter the following, which is long running command:

```
waitfornewblock 50000
```

On the current branch you can enter as many of them as you want and they will execute one after another, and return when they return. With the suggested patch you cannot enter another rcp command until that one returns, which is the Qt Widgets gui behavior.

qml/pages/node/CommandConsole.qml
```diff
@@ -0,0 +1,473 @@
...
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import org.bitcoincore.qt 1.0
+import "../../controls"
+import "../../components"
+
+Page {
```
jarolrod, line 12: This component is currently functional, and does the right things, so we can merge.

Longer term, though, I think the RPC console would benefit from being split into a few focused pieces.

Right now this one QML file owns page layout, warning/header text, output rendering, command input, keyboard handling, history navigation, autocomplete filtering, popup behavior, focus policy, and wallet context submission. 

I think this warrants refactoring the console around a clearer ownership model (placeholder names below, we can figure out naming later):

```text
RpcConsoleModel / RpcConsoleController
  - RPC execution
  - busy state
  - command history
  - wallet scoping
  - command acceptance policy
  - available commands / completions
  - output model

CommandConsole.qml
  - page layout
  - warning/header area
  - places output view and command line

RpcCommandLine.qml
  - TextField + submit button
  - Enter/Tab/Up/Down handling
  - history navigation
  - accepted/refused submit UX

RpcCompletionPopup.qml
  - suggestion list
  - highlighted row behavior
  - click/keyboard completion behavior

ConsoleOutputView.qml / MonospaceOutputView
  - output rendering
  - scroll behavior
  - selection/copy behavior
```

The important design direction is that command acceptance should remain controller owned. 

QML can ask `canSubmit(command)` or consume a richer `SubmitResult`, but it shouldn't need to duplicate special cases like "stop is allowed while another command is executing." 

That keeps direct invokable calls, keyboard submission, button submission, and tests all going through the same policy.

While also giving us the benefit of a clean architecture split, and easier testability, and finer grain testability of each component independently.




Architecture diagram:

```mermaid
flowchart TD
    Page["CommandConsole.qml<br/>page layout only"]
    Header["Header / warning area"]
    Output["ConsoleOutputView.qml<br/>output rendering + scroll"]
    Line["RpcCommandLine.qml<br/>input, submit button, key handling"]
    Popup["RpcCompletionPopup.qml<br/>suggestions + highlighted item"]

    Controller["RpcConsoleModel / Controller<br/>acceptance policy, history, completions,<br/>wallet scoping, output model"]
    Worker["RpcConsoleWorker<br/>background RPC execution"]
    Node["interfaces::Node<br/>executeRpc / listRpcCommands"]

    Page --> Header
    Page --> Output
    Page --> Line
    Line --> Popup

    Line -->|"canSubmit / submitCommand"| Controller
    Popup -->|"selected completion"| Line
    Output -->|"binds outputModel"| Controller
    Controller --> Worker
    Worker --> Node
    Worker -->|"resultReady"| Controller

File: qml/models/desktoptrayiconcontroller.cpp [hunk 6]
```diff
@@ -0,0 +1,190 @@
...
+    : QObject(parent)
+    , m_tray_icon(new QSystemTrayIcon(this))
+{
+    // Linux SNI/AppIndicator hosts require a native QMenu registered via
+    // setContextMenu() — the menu is exported over D-Bus and displayed by
+    // the DE natively. Without this, right-click does nothing on GNOME/KDE.
+    auto* menu = new QMenu();
+    m_show_action = menu->addAction(tr("Show"));
+    connect(m_show_action, &QAction::triggered, this, [this] {
+        if (m_window_visible) {
```
jarolrod, line 48: I think in a follow-up we can make the tray controller use one shared show/hide/restore model.

The tray click path and the native menu path currently make different decisions:

- `toggleWindow()` checks the real window state: hidden, minimized, and obscured. 
- The menu action only checks `m_window_visible`. 

This means that a covered but visible window can be treated differently depending on whether the user clicks the tray icon or uses the OS native menu; tray click would try to show/raise it, while the menu can label/action it as “Hide”.

I don't think this needs to block this PR, but the follow up should probably centralize the decision in one helper, for example:

```cpp
bool DesktopTrayIconController::shouldShowMainWindow() const
{
    if (!m_main_window) return false;
    if (!m_main_window->isVisible()) return true;
    if (m_main_window->windowStates() & Qt::WindowMinimized) return true;
    return isObscured(m_main_window);
}
```

Then use that same logic for:

- tray icon activation
- native menu action
- native menu label text

qml/models/debuglogmodel.cpp
jarolrod, line 277: I noticed here that there is probably a related followup we would want to do:

This `hasMoreLines` calculation cannot reliably detect older entries. `ReadRawLines()` returns at most `loadLimit` rows, and then this path checks `filtered.size() > m_load_limit`. For a normal large log whose newest `loadLimit` rows are non-blank, `filtered.size()` will be exactly `m_load_limit`, not greater than it, even if the file has many older entries. So `hasMoreLines` becomes false and the button stays hidden.

jarolrod: Building on the above, I think the debug log viewer would benefit from a refactor that treats this as a small log viewing subsystem rather than one list model that happens to read a file.

From first principles, the debug log viewer has four separate responsibilities:

1. read the file correctly and efficiently
2. track pagination and file identity
3. expose rows to QML with precise model notifications
4. render/search/update visible UI

Right now those responsibilities are mixed together. That makes a few correctness issues hard to avoid.

The current model mostly knows “I loaded N rows from the tail,” then tries to infer everything else from that. That makes `Load more` a limit bump plus reread, append refresh a search for the old top row by message text, and rotation/truncation detection dependent on whether old content can still be found.

I think the cleaner followup is to make this cursor-based: the reader returns rows plus `oldestCursor`, `newestCursor`, `hasOlderEntries`, and file generation. Then initial load, `Load more`, appends, and reloads become direct operations instead of things that we try to best effort guesstimate.

That also gives cleaner model notifications: insert new rows at the top, insert older rows at the bottom, emit `dataChanged` for relative-time-only updates, and reserve full resets for true replacement cases like filtering or rotation.

I think a good architecture to move towards is one that implements the following split:

```text
DebugLogReader
  Owns file facts. It should be a pure file-IO worker.
  It can return structured results from:
  - readTail(limit)
  - readOlder(before_cursor, limit)
  - readAppended(after_cursor)
  - readReload(limit)

DebugLogStore
  Owns loaded-window state:
  - loaded rows
  - newest cursor
  - oldest cursor
  - hasOlderEntries
  - file generation

DebugLogModel
  Owns QAbstractListModel behavior:
  - exposes rows and roles to QML
  - emits beginInsertRows for appended/older rows
  - emits dataChanged for role-only updates
  - uses beginResetModel only for full replacement cases

SettingsDebugLog.qml / log view
  Owns presentation:
  - search controls
  - load-more button
  - visible list rendering
  - ideally virtualized row creation
```


```mermaid
flowchart TD
    QML["SettingsDebugLog.qml<br/>controls + page state"]
    View["Virtualized Log View<br/>renders visible rows only"]
    Model["DebugLogModel<br/>QAbstractListModel adapter"]
    Store["DebugLogStore<br/>loaded rows + cursors + file generation"]
    Reader["DebugLogReader<br/>pure async file IO"]
    Watcher["LogFileWatcher<br/>file/directory events"]
    File["debug.log"]

    QML --> View
    View --> Model
    QML --> Model

    Model --> Store
    Model --> Reader
    Watcher --> Model

    Reader --> File
    Watcher --> File
    Reader --> Store
```

The important shift here is that the model stops guessing and starts tracking where it actually is in the file. The reader gives the model real cursors and file information.

This would make the behavior easier to reason about and test, while fixing the current pagination ambiguity, stale relative time labels, duplicate content refresh risk, and log lifecycle edge cases.

jarolrod:
ACK 8f4a8379742ceece1b200292d740ad95d8097fca

These changes are good to merge now. Important follow-ups are documented in review.

johnny9 (approved):
ACK 8f4a8379742ceece1b200292d740ad95d8097fca

hebasto:
Needs rebase.

epicleafies:
@pseudoramdom I fixed the 2 blockers from #750 here (send crash and adding to CI flow). Rest of review issues are less urgent.

pseudoramdom:
nit: Should we give a padding to the left of the menu? Feels like there's no breathing room there compared to the rest of the app.
Screenshot 2026-06-24 at 3 20 16 PM

johnny9 (commented):
There are 2 commits in the PR that don't compile. Please resolve this before we ACK/merge. 

8ac27bdc15 build: Add Qt6::Widgets and optional DBus support for system tray

48f9f2355b qt: Add DesktopWindowBehaviorModel

Both fail at CMake configure because test/CMakeLists.txt references test_desktoptrayiconcontroller.cpp before that file exists. The fix is straightforward: move the test_desktoptrayiconcontroller.cpp addition to test/CMakeLists.txt into the commit that adds that file, or add the file in the earlier build-system commit.

Everything else looks good enough to ACK/merge. I have added an issue to track a bug that Jarol also mentioned relating to the relativeTime updates.

johnny9 (commented):
The second commit 8d3d0818fd  doesn't link properly

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 0a8291af95..97dd827f04 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,18 +105,20 @@ target_link_libraries(bitcoinqml
     Qt6::Quick
     Qt6::QuickControls2
     Qt6::Network
     Qt6::Widgets
+  PRIVATE
+    bitcoin_cli
 )
```

johnny9 (approved):
ACK 6ff51b854d12e4852ba2216302f43efed15000e8

Re-reviewed the latest push and confirmed the commit-order build issue is fixed

Manually tested on Ubuntu and macOS. Remaining non-blocking issues are tracked for follow-up.

Issue #745, opened 2026-06-03
Consolidate one-off menu implementations with a reusable component


pseudoramdom:
The codebase has accumulated several one-off context-menu implementations with inconsistent visual and behavioral conventions:

- `SendOptionsPopup`, `ReceiveOptionsPopup`, `PaymentDetailOptionsPopup` each roll a `Popup` (or `OptionPopup`) wrapping `EllipsisMenuActionItem` / `EllipsisMenuToggleItem` / `EllipsisMenuButtonItem` with
  hand-tuned widths, paddings, and `Item`/`Rectangle` separators.
 - `Activity.qml` defines an inline `FilterMenuItem: ItemDelegate` (radio-style row with checkmark) used nowhere else.
- `FeeSelection.qml` builds its own `Popup` with a custom `Repeater` of `ItemDelegate`s, a manual divider, and a toggle `ItemDelegate`.

The row heights vary (33px vs 36px vs 44px), the EllipsisMenu* API leaks layout knobs (`leftIconSource`, `contentTopPadding`, `bgRadius`), and there is no shared abstraction for a "picker section with checkmark" — so each new menu reinvents one

<img width="260" height="285" alt="Image" src="https://github.com/user-attachments/assets/057103e6-b83c-43b7-b59d-9388e338054e" />
<img width="328" height="314" alt="Image" src="https://github.com/user-attachments/assets/af6a9163-339c-427f-9715-a64be3f028fa" />
<img width="314" height="242" alt="Image" src="https://github.com/user-attachments/assets/154bfe74-38d9-4ce6-8a8b-3a0a6c07e3b4" />
<img width="326" height="230" alt="Image" src="https://github.com/user-attachments/assets/d5333696-2737-4a39-86ed-8df0861e4cb7" />
<img width="303" height="296" alt="Image" src="https://github.com/user-attachments/assets/2c37d895-f3b5-4ec4-ba95-7eada86119c9" />
<img width="331" height="222" alt="Image" src="https://github.com/user-attachments/assets/74d5794d-486b-44fe-a48d-3e463e79c206" />

pseudoramdom:
I'll take this

pseudoramdom:
This was implemented in #750 . We can close this.

Issue #664, opened 2026-05-06
PSBT import: support partial, multisig, and non-wallet PSBT workflows


johnny9:
## Gap
QML currently rejects many PSBTs that Qt can inspect or partially sign, including non-wallet inputs, multisig inputs, incomplete signatures, non-address outputs, and broader recipient sets.

## Work
Relax imported-PSBT restrictions where safe, support partial signing/inspection workflows, and keep clear messaging for unsupported cases. This can follow v1.0 after the basic release PSBT flow is complete.

## Code References
- `qml/models/walletqmlmodel.cpp:1413`
- `bitcoin/src/qt/psbtoperationsdialog.cpp:59`

pseudoramdom:
I'm taking this.

Issue #663, opened 2026-05-06
PSBT operations: add inspect, sign, broadcast, copy, and save screen


johnny9:
## Gap
The backend has imported-PSBT state and operations, but the Send UI only imports a file and either converts fully signable wallet-owned PSBTs into review or shows unsupported. There is no reachable PSBT operations screen or load-from-clipboard flow.

## Work
Add a PSBT operations UI that can load from file and clipboard, inspect summaries, sign what the wallet can sign, broadcast when complete, copy the PSBT, save it to disk, and surface operation status.

## Code References
- `qml/pages/wallet/Send.qml:337`
- `qml/models/walletqmlmodel.h:76`
- `qml/models/walletqmlmodel.cpp:1541`
- `bitcoin/src/qt/psbtoperationsdialog.cpp:55`

pseudoramdom:
I'll take this.

pseudoramdom:
@johnny9 we can close this.

Issue #522, opened 2026-02-28
Feature Parity 18: PSBT Import/Export Flow


johnny9:
## Summary
Implement PSBT export from send flow and PSBT import/sign/finalize pathways.

## Complexity
- `XL`

## Dependencies
- `16`
- `02`

## Global Constraints
- All implementation lives under `gui-qml/`.
- No GUI runtime dependency on `bitcoin/src/qt` components. Reused logic must be moved/ported into `gui-qml/`.
- Do not change `interfaces::Wallet` or `interfaces::Node` in this version.
- Issue completion must satisfy `DefinitionOfDone.md`.

## Scope
- Add create/export PSBT actions (clipboard + file).
- Add PSBT import path (file + clipboard) and decode/validation UX.
- Support sign/finalize/broadcast paths using wallet interface methods.
- Integrate with external signer readiness states where relevant.

## Acceptance Criteria (DoD-Aligned)
1. User can export unsigned PSBT from transaction review.
2. User can import PSBT, inspect status, sign if possible, and finalize/broadcast when complete.
3. Large/invalid PSBT inputs fail gracefully with actionable messages.

## Required Tests
- C++ unit tests: Unit tests for PSBT parsing/sign/finalize adapters and error handling.
- QML QTest: QTests for import/export dialog states and action enablement.
- Python E2E (`qml/test/testbridge.h`): E2E flow covering PSBT export, re-import, sign/finalize, and broadcast.

## Implementation Notes
- Do not reuse `qt/psbtoperationsdialog.*`; implement gui-qml-native model + UI.

PR #750, opened 2026-06-08
Implement PSBT import, export, inspect, broadcast operations

### LINKED ISSUE #662: Watch-only send: create unsigned PSBT instead of direct send
## Gap
Private-key-disabled/watch-only wallets should create an unsigned PSBT, but QML review only offers Send and then tries to sign/finalize through the normal send path.

## Work
Detect private-key-disabled wallets during send review, switch the action to create an unsigned PSBT, and provide copy/save behavior plus user-facing success and failure states.

## Code References
- `qml/pages/wallet/SendReview.qml:123`
- `qml/models/walletqmlmodel.cpp:1274`
- `bitcoin/src/qt/sendcoinsdialog.cpp:215`
- `bitcoin/src/qt/sendcoinsdialog.cpp:503`

pseudoramdom:
This PR brings the ability to import, export and view PSBT files.

### PSBT operations
- Add ability to import PSBT file via the send options menu. This opens a file dialog to import the PSBT file.
  <details><summary>Screenshot</summary>Screenshot 2026-06-07 at 8 57 37 PM</details> 
- Export a transaction as PSBT when in Review transaction page
   <details><summary>Screenshot</summary>Screenshot 2026-06-07 at 9 00 00 PM</details> 
- Add support to view PSBT files that a wallet can't sign, including multi-sig PSBTs
   <details><summary>Screenshot</summary>RegularScreenshot 2026-06-07 at 7 59 57 PM MultisigScreenshot 2026-06-07 at 8 22 05 PM</details> 
- Broadcast a transaction if it has required signatures
   <details><summary>screenshot</summary>Screenshot 2026-06-07 at 7 59 44 PM</details> 
### Other UI fixes
- Introduce a new ContextMenu component that can list options with button with icons, toggle,  separator. It replaces `EllipsisMenu` family of controls. Migrate existing menus to use ContextMenu across the app.
  <details><summary>screenshot</summary>Screenshot 2026-06-07 at 9 08 05 PMScreenshot 2026-06-07 at 9 08 13 PMScreenshot 2026-06-07 at 9 08 23 PMScreenshot 2026-06-07 at 9 08 39 PM</details> 
- Refine Send and receive page layout
  <details><summary>screenshot</summary>Screenshot 2026-06-07 at 5 11 37 PMScreenshot 2026-06-07 at 9 14 20 PM</details> 
- Review Transaction for watch-only wallets
  <details><summary>screenshot</summary>Screenshot 2026-06-07 at 7 58 38 PM</details> 

Fixes #662 , #663 , #664, #522, #745


pseudoramdom:
Supersedes https://github.com/bitcoin-core/gui-qml/pull/557

qml/models/psbtqmlmodel.cpp
```diff
@@ -0,0 +1,391 @@
...
+}
+
+QString PsbtQmlModel::LoadPsbtFromFile(const QString& path, PartiallySignedTransaction& psbt)
+{
+    QFile file(LocalFilePath(path));
+    if (!file.open(QIODevice::ReadOnly)) {
+        return tr("Could not open PSBT file: %1").arg(file.errorString());
+    }
+
+    const QByteArray bytes{file.readAll()};
```
jarolrod, line 115: we probably need to add the core file size guard here before readAll()

This reads the entire selected file into memory before any validation, and the decode paths also make additional full size copies. 

Core defines MAX_FILE_SIZE_PSBT for this, and Qt Widgets rejects oversized PSBT files before loading them.

A very large selected file could freeze or OOM the gui before decode fails.

qml/models/walletqmlmodel.cpp
```diff
@@ -1686,6 +1856,228 @@ bool WalletQmlModel::sendTransactionInternal()
...
+        draft_recipients.push_back({address, getAddressLabel(address), output.nValue});
+        recipient_total += output.nValue;
+    }
+
+    if (draft_recipients.empty()) {
+        reason = tr("The PSBT does not have any recipient outputs to review.");
+        return false;
+    }
+    if (draft_recipients.size() > 25) {
+        reason = tr("The PSBT has more recipients than this send flow supports.");
```
jarolrod, line 2032: how come our qml send flow can only support 25 max recipients?

just documenting that the core limit is weight instead of recipients.

I see with send we are imposing a 25 limit as well. 

Maybe some documentation on these design choices.

qml/models/walletqmlmodel.cpp
```diff
@@ -1686,6 +1856,228 @@ bool WalletQmlModel::sendTransactionInternal()
...
+        QString address;
+        QString label;
+        CAmount amount;
+    };
+    std::vector<DraftRecipient> draft_recipients;
+    CAmount recipient_total{0};
+    for (const CTxOut& output : analysis_psbt.tx->vout) {
+        CTxDestination destination;
+        if (!ExtractDestination(output.scriptPubKey, destination)) {
+            reason = tr("Only PSBTs with standard address outputs are supported right now.");
```
jarolrod, line 2016: We would reject an OP_RETURN output here, and those are quite popular

I'm not sure still why we are imposing extra restrictions on the qml side

qml/models/walletqmlmodel.cpp
```diff
@@ -1604,49 +1639,57 @@ void WalletQmlModel::approveExternalSignerTransaction()
...
+                return;
             }
-            Q_EMIT externalSignerApprovalFailed(message, signer_not_found);
-            return;
         }
 
-        complete = FinalizeAndExtractPSBT(psbtx, mtx);
-        if (!complete) {
+        CMutableTransaction signed_tx;
+        if (!FinalizeAndExtractPSBT(psbtx, signed_tx)) {
```
jarolrod, line 1685: I think this drops a valid external signer result in the incomplete case. Core’s external signer notes that `FinalizePSBT()` “won’t work in a multisig setup”.

Widgets handle this differently: after external signing, `signWithExternalSigner()` returns success even if `FinalizeAndExtractPSBT()` leaves `complete=false`, and `SendCoinsDialog` then calls `presentPSBT(psbtx)` so the user can copy/save the partially signed PSBT for the next signer. 

Here we emit `INCOMPLETE` and return before updating `m_current_psbt`, so the partial signature is discarded; saving afterwards reconstructs an unsigned PSBT.

We should probably for now preserve this behavior. I believe you are working on better multisig support, so maybe you intend this to be intentional for now?

qml/models/psbtqmlmodel.cpp
pseudoramdom: This is a good catch. I should've caught this when I ported over the code (from #557)
I'll mirror the qt-widgets check to use MAX_FILE_SIZE_PSBT

qml/models/walletqmlmodel.cpp
pseudoramdom: Will document it.
Something @epicleafies also caught while reviewing https://github.com/bitcoin-core/gui-qml/pull/557/changes#r3370700953

jarolrod: the thing is that this is a new gui restriction on being able to broadcast a psbt; aside from how many recipients we allow during creation. So documentation on why we are adding that new restriction here would be valuable.

A user with many recipients will have to use the rpc, instead of our UI flow

pseudoramdom: Looks like this was inherited from https://github.com/bitcoin-core/gui-qml/pull/450/changes/5fd444135a0f118bab0681036875610d637c960e change.
The limit seems arbitrary to me. I'm okay to drop this limit on GUI. 
Unless @johnny9 has more insight on the limit?

qml/models/walletqmlmodel.cpp
pseudoramdom: Ah! I previously hit a bug while testing wherein we'd sign automatically before exporting the PSBT. I might've regressed this there.
I changed the flow to preserve the externally signed PSBT when signing succeeds but the PSBT remains incomplete.

test/qml/tst_address_components.qml
```diff
@@ -84,23 +93,29 @@ TestCase {
         })
         const noteButton = findObject(row, "addressRowNoteButton")
         verify(noteButton !== null)
-        noteButton.clicked()
+        mouseClick(noteButton, noteButton.width / 2, noteButton.height / 2)
```
jarolrod, line 103: it looks like this is making the linux ci task fail because we switched to a real mouseClick, but the row is not created under a visible item. The click never reaches the note button, so editRequested stays false. 

It should be enough to make the test case visible and put the row under a visible host item.

like so:

<details>

```diff
diff --git a/test/qml/tst_address_components.qml b/test/qml/tst_address_components.qml
index 6ba2baeb9a..c2365759c5 100644
--- a/test/qml/tst_address_components.qml
+++ b/test/qml/tst_address_components.qml
@@ -13,6 +13,13 @@ TestCase {
     when: windowShown
     width: 720
     height: 480
+    visible: true
+
+    Item {
+        id: host
+        width: parent.width
+        height: parent.height
+    }
 
     function findObject(root, objectName) {
         if (!root) {
@@ -80,7 +87,7 @@ TestCase {
     }
 
     function test_row_uses_display_amount_and_emits_actions() {
-        const row = createTemporaryObject(rowComponent, this)
+        const row = createTemporaryObject(rowComponent, host)
         verify(row !== null)
 
         const amountText = findObject(row, "addressRowAmountText")
```

</details>

Maybe there is a more elegant fix.

qml/models/walletqmlmodel.cpp
```diff
@@ -1686,6 +1863,228 @@ bool WalletQmlModel::sendTransactionInternal()
...
+    if (m_current_transaction) {
+        delete m_current_transaction;
+    }
+    m_current_transaction = new WalletQmlModelTransaction(m_send_recipients, this);
+    m_current_transaction->setWtx(MakeTransactionRef(*analysis_psbt.tx));
+    if (analysis.fee) {
+        m_current_transaction->setTransactionFee(*analysis.fee);
+    }
+    if (can_send || can_broadcast) {
+        m_current_psbt = std::make_unique<PartiallySignedTransaction>(std::move(analysis_psbt));
```
jarolrod: I think this should store the original imported PSBT, not `analysis_psbt`.

`analysis_psbt` has already gone through `FinalizePSBT()` above, and that call mutates the PSBT. 

For a complete-but-not-finalized PSBT, finalization moves data into the final script fields and clears fields like `partial_sigs`, keypaths, and scripts. So if the user only imports/reviews the PSBT and then clicks “Save transaction”, they can get back a different PSBT than the one they imported.

I added a local regression test for this with a complete unfinalized 2-of-2 P2WSH PSBT. It imports the PSBT, saves it without signing/broadcasting, then reloads it. On this branch the saved PSBT has 0 partial sigs, while the original has 2.

see:

<details>

```diff
diff --git a/test/test_walletqmlmodel.cpp b/test/test_walletqmlmodel.cpp
index 274f20a5ea..0e0a67f9e8 100644
--- a/test/test_walletqmlmodel.cpp
+++ b/test/test_walletqmlmodel.cpp
@@ -467,6 +467,7 @@ private Q_SLOTS:
     void externalSignerApprovalKeepsIncompleteSignedPsbt();
     void importPsbtFromFile_opensForeignUnsignedPsbtForReviewOnly();
     void importPsbtFromFile_broadcastsCompleteForeignMultisigPsbt();
+    void saveCompleteBroadcastableImportedPsbt_preservesOriginalPsbt();
     void importPsbtFromFile_opensUnsignedMultisigPsbtForReviewOnly();
     void importPsbtFromFile_blocksBroadcastWhenFeeIsInvalid();
     void importPsbtFromFile_returnsTransactionAlreadyKnownWhenTxIsInWallet();
@@ -2330,6 +2331,81 @@ void WalletQmlModelTests::importPsbtFromFile_broadcastsCompleteForeignMultisigPs
     QVERIFY(model->transactionError().isEmpty());
 }
 
+void WalletQmlModelTests::saveCompleteBroadcastableImportedPsbt_preservesOriginalPsbt()
+{
+    NiceMock<MockNode> node;
+    FakePasswordWallet* wallet{nullptr};
+    auto model = MakeWalletModel(wallet, &node);
+
+    wallet->fill_psbt_fn = [wallet](std::optional<int>,
+                                    bool sign,
+                                    bool,
+                                    size_t* n_signed,
+                                    PartiallySignedTransaction&,
+                                    bool& complete) {
+        wallet->fill_psbt_sign_args.push_back(sign);
+        if (n_signed) {
+            *n_signed = 0;
+        }
+        complete = false;
+        return std::nullopt;
+    };
+
+    CKey key1;
+    CKey key2;
+    key1.MakeNewKey(true);
+    key2.MakeNewKey(true);
+    const CScript witness_script{GetScriptForMultisig(2, {key1.GetPubKey(), key2.GetPubKey()})};
+
+    CMutableTransaction mtx;
+    mtx.vin.emplace_back(COutPoint{Txid::FromUint256(uint256::ONE), 0});
+    mtx.vout.emplace_back(1'500, GetScriptForDestination(DecodeDestination(VALID_MAINNET_ADDRESS.toStdString())));
+
+    PartiallySignedTransaction psbt{mtx};
+    psbt.inputs[0].witness_utxo = CTxOut{2'000, GetScriptForDestination(WitnessV0ScriptHash(witness_script))};
+    psbt.inputs[0].witness_script = witness_script;
+
+    FillableSigningProvider provider;
+    QVERIFY(provider.AddCScript(witness_script));
+    QVERIFY(provider.AddKey(key1));
+    QVERIFY(provider.AddKey(key2));
+
+    const PrecomputedTransactionData txdata{PrecomputePSBTData(psbt)};
+    QCOMPARE(
+        SignPSBTInput(provider, psbt, 0, &txdata, std::nullopt, nullptr, /*finalize=*/false),
+        PSBTError::OK);
+
+    QVERIFY(!psbt.inputs[0].partial_sigs.empty());
+    QVERIFY(!psbt.inputs[0].witness_script.empty());
+    QVERIFY(psbt.inputs[0].final_script_witness.IsNull());
+
+    PartiallySignedTransaction finalized_copy{psbt};
+    QVERIFY(FinalizePSBT(finalized_copy));
+    QVERIFY(finalized_copy.inputs[0].partial_sigs.empty());
+    QVERIFY(finalized_copy.inputs[0].witness_script.empty());
+    QVERIFY(!finalized_copy.inputs[0].final_script_witness.IsNull());
+
+    QTemporaryDir temp_dir;
+    QVERIFY(temp_dir.isValid());
+    const QString source_path{WritePsbt(psbt, temp_dir, QStringLiteral("complete-unfinalized.psbt"))};
+    QVERIFY(!source_path.isEmpty());
+
+    QCOMPARE(model->importPsbtFromFile(source_path), WalletQmlModel::PsbtImportResult::WalletCannotSign);
+    QVERIFY(!model->currentTransactionCanSend());
+    QVERIFY(model->currentTransactionCanBroadcast());
+
+    const QString saved_path{temp_dir.filePath(QStringLiteral("saved.psbt"))};
+    QCOMPARE(model->saveCurrentTransactionAsPsbt(saved_path), QString{});
+
+    PartiallySignedTransaction saved;
+    QCOMPARE(PsbtQmlModel::LoadPsbtFromFile(saved_path, saved), QString{});
+
+    QCOMPARE(saved.inputs[0].partial_sigs.size(), psbt.inputs[0].partial_sigs.size());
+    QVERIFY(saved.inputs[0].witness_script == psbt.inputs[0].witness_script);
+    QVERIFY(saved.inputs[0].final_script_witness.IsNull());
+    QCOMPARE(PsbtQmlModel::SerializePsbtRaw(saved), PsbtQmlModel::SerializePsbtRaw(psbt));
+}
+
 void WalletQmlModelTests::importPsbtFromFile_opensUnsignedMultisigPsbtForReviewOnly()
 {
     FakePasswordWallet* wallet{nullptr};
```
</details>

A possible fix is keep using `analysis_psbt` as a temporary scratch copy for `FinalizePSBT()`, `AnalyzePSBT()`, fee checks, and `can_broadcast`, but keep/store the original imported `psbt` for passive save/export. 

Broadcasting can still finalize/extract from a temporary copy.

I think this issue is hairy, and is tangentially related to how, currently, the gui PSBT usage implementation differs from the internal core PSBT wallet possible operations/possible outputs.

jarolrod: Also, maybe we should add more tests for PSBT operations in this PR. Another one we can add is the oversized PSBT one we previously found.

jarolrod: I took some more time to think about what was going on here, and I think overall we need to revisit some invariants.

Let's think of the possible user roles in a PSBT's lifecycle:

```mermaid
flowchart LR
    A["Creator"] --> B["Updater"]
    B --> C["Signer"]
    C --> D["Combiner"]
    D --> E["Finalizer"]
    E --> F["Extractor"]
    F --> G["Broadcaster"]
```

The QML Send form is **one** possible user role. Imported PSBTs arrive at any later stage: unsigned, partially signed, fully signed but not finalized, finalized, foreign, watch-only, multisig, external-signer, or broadcast-only.

That means imported PSBT review **cannot** depend on the editable `SendRecipient` model. Recipient rows are draft construction state. Imported PSBT outputs are review session state.

For wallet-created transactions, the current Send flow makes sense:

```mermaid
flowchart TD
    A["User edits recipients and fee"] --> B["Wallet creates funded transaction"]
    B --> C["Review Send draft"]
    C --> D{"User action"}
    D --> E["Save unsigned PSBT"]
    D --> F["Sign and wallet commit"]
    D --> G["External signer approval"]
    G --> H{"Complete?"}
    H -- "yes" --> F
    H -- "no" --> I["Save partially signed PSBT"]
```

But imported PSBTs need a different lifecycle:

```mermaid
flowchart TD
    A["Import PSBT"] --> B["Decode and validate shape"]
    B --> C["Create imported PSBT session"]
    C --> D["Analyze copies only"]
    D --> E["Render all outputs read-only"]
    E --> F{"Capabilities"}
    F -- "wallet can sign" --> G["Sign transaction"]
    F -- "complete and fee known" --> H["Broadcast transaction"]
    F -- "incomplete or foreign" --> I["Save / Done"]
    G --> J{"Still incomplete?"}
    J -- "yes" --> K["Save partial signature"]
    J -- "no" --> H
    H --> L["Finalize/extract copy"]
    L --> M["Node broadcast"]
```

The key point is object ownership. Passive review should not mutate the PSBT the user later saves. 

Core finalization is intentionally mutating: it can move partial signatures and scripts into final script fields and clear the original partial data.

So the imported session needs separate concepts:

- `original_psbt`: the exact imported PSBT.
- `current_psbt`: the working PSBT after explicit user actions.
- `save_psbt`: what Save exports.
- Analysis, finalization, and broadcast copies: temporary objects only.

That gives us the right behavior:

- Save before signing preserves the original.
- Signing updates the saveable PSBT with new signatures.
- Partial signatures are useful output and remain saveable.
- Broadcast finalizes and extracts a temporary copy.
- Broadcast does not rewrite the saved PSBT as a side effect.

**The same split fixes the output issue we've been having.** 

A valid PSBT can contain `OP_RETURN`, unknown scripts, wallet owned outputs, many outputs, or outputs that do not convert to addresses. Those are not reasons to reject import or broadcast. They are review and display concerns.

```mermaid
flowchart TD
    A["PSBT output"] --> B{"ExtractDestination works?"}
    B -- "yes" --> C["Address row"]
    B -- "no" --> D{"IsUnspendable?"}
    D -- "yes" --> E["Data output row"]
    D -- "no" --> F["Unknown script row"]
    F --> G{"value > 0?"}
    G -- "yes" --> H["Show warning before sign/broadcast"]
    G -- "no" --> I["Render script detail"]
```

The UI does not need to become a full transaction editor in this PR. It just needs to faithfully represent the imported PSBT session.

The minimum target architecture I would recommend for this PR should be:

1. Wallet created Send draft session stays editable and uses wallet commit for Send.
2. Imported PSBT review is read-only and uses PSBT output rows.
3. Imported PSBT **Save** preserves the original until the user explicitly signs.
4. Imported PSBT **Sign** mutates the session PSBT and keeps partial results saveable.
5. Imported PSBT **Broadcast** uses Core finalization and extraction on a copy, then node broadcast.
6. Imported PSBT **Broadcast** never calls wallet commit.
7. Unknown and proprietary fields are preserved.
8. Unknown positive value scripts bring up a warning (like sparrow wallet does), but do not block anything. `OP_RETURN` is addressless and normally zero-value, so it should render as a data output without becoming a scary failure. 

> This transaction has one or more positive-value outputs that cannot be shown as a Bitcoin address. Review the script details before signing or broadcasting.

So I think the original PR branch should move toward this shape rather than adding more exceptions around the existing Send recipient path. The branch is trying to add PSBT import, export, inspect, sign, and broadcast -> and we should think about those in terms of creation and import PSBT session operations.

## How I would adapt the current branch

The target state should look like this:

```mermaid
flowchart TD
    A["WalletQmlModel::importPsbtFromFile"] --> B["Decode PSBT"]
    B --> C{"Known wallet tx?"}
    C -- "yes" --> D["TransactionAlreadyKnown"]
    C -- "no" --> E["Create ImportedPsbtSession"]
    E --> F["Analyze temporary PSBT copies"]
    F --> G["Build read-only PSBT output rows"]
    G --> H["Return PsbtImportedForReview"]
    H --> I["Send.qml opens inspection review"]
    I --> J{"Session capabilities"}
    J -- "can_sign" --> K["Show Sign transaction"]
    J -- "can_broadcast" --> L["Show Broadcast transaction"]
    J -- "neither" --> M["Show Save / Done with review message"]
```

The model changes should be organized around an imported session object rather than around populated Send recipients. I would add a small private session type in `WalletQmlModel` with these fields:

- `original_psbt`: the exact PSBT loaded from disk.
- `current_psbt`: the working PSBT after explicit user actions.
- `save_psbt`: the PSBT that `Save transaction` writes.
- `can_sign`: whether the selected wallet can add signatures now.
- `can_broadcast`: whether Core can extract a final transaction with a known nonnegative fee.
- `review_message`: fee, multisig, cannot-sign, or unknown-script warning text.

The important implementation rule related to my original comment here is that we use review copies:

```mermaid
flowchart LR
    A["original_psbt"] --> B["save_psbt"]
    A --> C["analysis copy"]
    C --> D["fillPSBT sign=false bip32derivs=false"]
    C --> E["AnalyzePSBT"]
    C --> F["FinalizeAndExtractPSBT"]
    B --> G["Save transaction"]
```

pseudoramdom: Thanks for the detailed review. The diagrams were helpful.
I agree that send-recipient is probably overloaded here and that imported PSBT operations should eventually move into a separate review/session model rather than being projected into send-draft state.

As a middle ground for this PR, I’m planning to keep the current review flow, but change the saveable PSBT state so it preserves the original imported PSBT.

`analysis_psbt` will stay as a scrattch copy for `FinalizePSBT()`, `fillPSBT(sign=false)`, fee analysis etc, but it won’t be stored in `m_current_psbt`.
For imported PSBTs, `saveCurrentTransactionAsPsbt()` will write the retained original directly, so passive review + save won’t finalize or otherwise rewrite the imported PSBT.

test/qml/tst_address_components.qml
pseudoramdom: Thanks! Should be fixed now

jarolrod:
ACK 1113aa9d31ae621ba5055c47c76710b6cca18a97

In a reduced-scope, this is okay to merge. Will open issues for what is left to address.

qml/pages/wallet/Send.qml
```diff
@@ -115,11 +120,151 @@ PageStack {
...
+                wallet: reviewOnlyPsbtPopup.reviewWallet
+                onBack: reviewOnlyPsbtPopup.close()
+                onTransactionSent: {
+                    reviewOnlyPsbtPopup.broadcastSucceeded = true
+                    reviewOnlyPsbtPopup.close()
+                }
+            }
+        }
+
+        AlertPopup {
```
epicleafies, line 201: Blocker (crash, headed on Qt 6.4.x): this `AlertPopup`, and the two siblings this PR adds on the Send page (`unsupportedPsbtPopup`, `knownPsbtTxPopup`), crash the app when navigating to the Send tab. `AlertPopup.qml` nests a `Repeater` directly inside a Quick `RowLayout` (contentItem `ColumnLayout` -> `RowLayout` -> `Repeater` over `visibleActions`), and `Component.onCompleted: refreshActions()` swaps the `visibleActions` model during construction. On Qt 6.4.x the Repeater rebuilds its delegates while the `RowLayout` does its first rearrange on Send-tab navigation, so `QQuickLayout::effectiveSizePolicy_helper` (`qquicklayout.cpp:1188`) runs `qobject_cast<QQuickLayout*>` on a freed delegate and SIGSEGVs. Confirmed introduced by this PR via a same-machine merge-base A/B, and confirmed fixed by changing that `RowLayout` to a `Row`. It reproduces under a real windowing backend (xcb/wayland) but not offscreen, which is why the offscreen functional harness on the Qt 6.4.2 CI runner stays green even on the Send tests (and so wiring the new `qml_test_psbt.py` would not catch it either), and it is masked on newer Qt. Fix the shared `AlertPopup` (avoid a Repeater directly in a Quick Layout, or stabilize `visibleActions` before first layout) and verify headed on Qt 6.4.

qml/models/psbtqmlmodel.cpp
```diff
@@ -0,0 +1,394 @@
...
+    refreshState(tr("PSBT saved."));
+    return QString{};
+}
+
+void PsbtQmlModel::refreshState(const QString& status_override)
+{
+    if (!m_psbt) {
+        return;
+    }
+    bool complete{FinalizePSBT(*m_psbt)};
```
epicleafies, line 327: Suggestion (latent correctness, root cause): `FinalizePSBT(*m_psbt)` mutates the stored PSBT. `FinalizePSBT` is destructive (it clears `partial_sigs`/`redeem_script`/`witness_script` into the final scripts for any finalizable input), so a later `saveToFile`/`copyToClipboard` emits a finalized PSBT rather than the imported one, dropping combiner-relevant fields for a complete-but-not-finalized or partially-signed PSBT. This is the same bug class fixed in the wallet path (finalize a copy, keep the original); apply it here too: compute state on a local copy and leave `m_psbt` untouched. Only reachable from unit tests today, but a landmine once this model is wired to the UI.

qml/pages/wallet/Send.qml
```diff
@@ -435,17 +589,35 @@ PageStack {
...
                         x: menuButton.x - width + menuButton.width
                         y: menuButton.y + menuButton.height
+                        coinControlEnabled: settings.coinControlEnabled
+                        multipleRecipientsEnabled: settings.multipleRecipientsEnabled
+                        onClearFormRequested: {
+                            sendOptionsPopup.close()
+                            root.wallet.recipients.clear()
+                        }
+                        onImportPsbtFromFileRequested: {
+                            if (psbtAutomationPathField.text.length > 0) {
```
epicleafies, line 599: Suggestion (security-hygiene, raised on #557): this production handler has a test-only fast path backed by a hidden `psbtImportPathField` (L226) that ships in the release binary, and the same shape exists for save in `SendReview.qml` (`savePsbtAutomationPath`). The C++ test bridge is already compiled behind `ENABLE_TEST_AUTOMATION`; these QML fields and their handler branches are not. Real-world severity is bounded (only the gated bridge can populate the field, so it is inert in a release build), so this is hygiene/defense-in-depth rather than an exploitable hole, but please gate these behind `ENABLE_TEST_AUTOMATION` so they match the bridge and do not exist in release builds.

qml/pages/wallet/SendReview.qml
```diff
@@ -16,31 +17,127 @@ Page {
...
+        }
+        if (root.wallet && root.wallet.broadcastCurrentTransaction()) {
+            root.sending = true
+            root.transactionSent()
+        }
+    }
+
+    function defaultSavePsbtFileUrl() {
+        const stamp = Qt.formatDateTime(new Date(), "yyyy-MM-dd-HHmm")
+        return "file://" + walletController.homePath() + "/transaction-" + stamp + ".psbt"
```
epicleafies, line 65: Suggestion (cross-platform): `"file://" + homePath()` is malformed on Windows. `homePath()` returns a plain path, so on Windows (`C:/Users/Foo`) this yields `file://C:/Users/Foo`, parsing `C:` as the URL authority, and the save dialog's default folder will be wrong. Same at L126 (`currentFolder`). Return a proper URL from C++ (`QUrl::fromLocalFile(QDir::homePath()).toString()`) instead of concatenating the scheme in QML, and verify on Windows.

qml/models/walletqmlmodel.cpp
```diff
@@ -1680,12 +1860,239 @@ bool WalletQmlModel::sendTransactionInternal()
...
+        const QString address{QString::fromStdString(EncodeDestination(destination))};
+        draft_recipients.push_back({address, getAddressLabel(address), output.nValue});
+        recipient_total += output.nValue;
+    }
+
+    if (draft_recipients.empty()) {
+        reason = tr("The PSBT does not have any recipient outputs to review.");
+        return false;
+    }
+    if (draft_recipients.size() > 25) {
```
epicleafies, line 2031: Question (raised on #557): `> 25` is a bare magic number, and this is a new GUI-only restriction on broadcasting an imported PSBT (more recipients forces the user to the RPC). If kept, name it (e.g. `MAX_REVIEWABLE_RECIPIENTS`) with a one-line rationale, or reference the send-side cap if this is meant to mirror it, so the two cannot drift.

qml/models/psbtqmlmodel.cpp
```diff
@@ -0,0 +1,394 @@
...
+}
+
+QString PsbtQmlModel::LoadPsbtFromFile(const QString& path, PartiallySignedTransaction& psbt)
+{
+    QFile file(LocalFilePath(path));
+    if (!file.open(QIODevice::ReadOnly)) {
+        return tr("Could not open PSBT file: %1").arg(file.errorString());
+    }
+    if (file.size() > MAX_FILE_SIZE_PSBT) {
+        return tr("PSBT file must be smaller than 100 MiB");
```
epicleafies, line 112: Nit: `MAX_FILE_SIZE_PSBT` is 100,000,000 bytes (100 MB, ~95.4 MiB), so the message should say 100 MB.

```suggestion
        return tr("PSBT file must be smaller than 100 MB");
```

qml/models/psbtqmlmodel.cpp
```diff
@@ -0,0 +1,394 @@
...
+bool IsMultisigScript(const CScript& script)
+{
+    return ExtractMultisigSigInfo(script).has_value();
+}
+
+bool DecodePsbtFromBytes(const QByteArray& bytes, PartiallySignedTransaction& psbt, std::string& error)
+{
+    std::vector<std::byte> raw;
+    raw.reserve(bytes.size());
+    for (const char ch : bytes) {
```
epicleafies, line 88: Nit (raised on #557): this copies the `QByteArray` byte-by-byte. You can span it directly (`std::as_bytes` over `bytes.constData()`/`bytes.size()`) and drop the loop.

qml/models/psbtqmlmodel.cpp
```diff
@@ -0,0 +1,394 @@
...
+#include <QClipboard>
+#include <QFile>
+#include <QGuiApplication>
+#include <QIODevice>
+#include <QSaveFile>
+#include <QUrl>
+
+namespace {
+
+QString FormatBtc(CAmount amount)
```
epicleafies, line 39: Nit (raised on #557): `FormatBtc` hardcodes the BTC unit, ignoring the user's display setting, and it is only reached by the unused `buildSummary`. Remove it together with that code, or, if the summary path gets wired up later, format via the active display unit instead of hardcoding BTC.

qml/controls/ContextMenuToggle.qml
```diff
@@ -0,0 +1,41 @@
...
+import QtQuick 2.15
+import QtQuick.Layouts 1.15
+
+ContextMenuButton {
+    id: root
+
+    autoClose: false
+    checkable: true
+
+    Accessible.checkable: true
```
epicleafies, line 14: Nit (a11y, raised on #557): this interactive row sets `Accessible.checkable`/`checked` but no `Accessible.name`/`Accessible.role`. `ContextMenuButton` sets `Accessible.role: Accessible.MenuItem` and `Accessible.name: text`; mirror that here.

qml/models/psbtqmlmodel.cpp
```diff
@@ -0,0 +1,394 @@
...
+    Q_EMIT changed();
+}
+
+void PsbtQmlModel::setMatchedTxid(const QString& txid)
+{
+    m_matched_txid = txid;
+    Q_EMIT changed();
+}
+
+QString PsbtQmlModel::loadFromFile(const QString& path)
```
epicleafies, line 223: Nit: `loadFromFile` does not reset `m_matched_txid` on the success path (only `clear()`/`setError()` do), so a stale matched-txid can survive a reload. Low impact since the wallet flow clears the model on success, but resetting it on load makes the model self-consistent regardless of caller order.

epicleafies (commented):
Concept ACK, found 2 blockers. Was able to run all automated tests and look over code but because of 1st blocker wasn't able to manually test behavior in the gui. The main issues are below, the smaller ones are inline.

Blocker: navigating to the Send tab crashes on Qt 6.4.2. Not sure if any later Qt versions are affected, my understanding was 6.4 is the minimum Qt version. Root cause: AlertPopup.qml nests a Repeater directly inside a Quick RowLayout (contentItem ColumnLayout -> RowLayout -> Repeater over visibleActions), and Component.onCompleted: refreshActions() swaps the visibleActions model during construction. This PR adds three AlertPopup instances to the Send page (broadcastSuccessPopup, unsupportedPsbtPopup, knownPsbtTxPopup), which lay out on Send-tab navigation; on Qt 6.4.x the Repeater rebuilds its delegates while the RowLayout does its first rearrange, so QQuickLayout::effectiveSizePolicy_helper (qquicklayout.cpp:1188) runs qobject_cast<QQuickLayout*> on a freed delegate and SIGSEGVs. It reproduces under a real windowing backend (xcb/wayland) but not offscreen, which is why the offscreen functional harness on the Qt 6.4.2 CI runner stays green even on the Send tests.

Blocker: the new functional test is not wired into CI. test/functional/qml_test_psbt.py wasn't added to .github/workflows/gui-functional-tests.yml

Architecture (meta): one addition to jarolrod's PSBT-lifecycle thread is the extracted PsbtQmlModel's stateful methods (sign/broadcast/refreshState/save/copy) are reachable only from unit tests, not the UI, so the refreshState finalize bug below sits in code the app never runs today, though it will once that follow-up wires it up.

PR #729, opened 2026-05-28
Update submodule to latest and include depends building

johnny9:
Closes #679

PR #752, opened 2026-06-11
Build bitcoin-core-app against depends 

johnny9:
This PR is currently based on #729

### DIFF
--- a/.github/workflows/depends-build.yml
+++ b/.github/workflows/depends-build.yml
@@ -1,0 +1,168 @@
+# Copyright (c) 2026 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+name: Depends Build
+
+on:
+  pull_request:
+  push:
+    branches:
+      - "**"
+    tags-ignore:
+      - "**"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: ${{ matrix.name }}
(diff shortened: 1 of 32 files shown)


.github/workflows/depends-build.yml
```diff
@@ -0,0 +1,168 @@
...
+          path: bitcoin/depends/built
+          key: ${{ steps.depends-built-cache.outputs.cache-primary-key }}
+
+      - name: Configure
+        run: |
+          test -f "bitcoin/depends/${HOST}/toolchain.cmake"
+          cmake -S . -B "${BUILD_DIR}" -G Ninja \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_TOOLCHAIN_FILE="${PWD}/bitcoin/depends/${HOST}/toolchain.cmake" \
+            -DBUILD_APP_TESTS=OFF \
```
jarolrod, line 146: Doing `BUILD_APP_TESTS=OFF` is right here because of the extra test only dependencies needed, and whether or not to add those is a separate conversation. 

But, a build only workflow still obviously misses the main depends failure mode: the statically linked app can compile and link but fail at startup because a Qt platform plugin or QML import is missing -> this is something that has happened before with the Qt Widgets gui.

Core does have some depends jobs where one task builds an artifact and another task runs tests (windows cross build). That split isn't possible here because it requires everything built in, which we don't produce the tests because of the dependencies which are not in the depends build system yet.

That said, it is probably worthwhile to figure out a minimal smoke test that can verify the depends build can run correctly.

Or document that we will revisit this when we start discussing real depends build QML testing.

johnny9: This will be fixed when we undo gtest/gmock

jarolrod: > This will be fixed when we undo gtest/gmock

ack

Issue #762, opened 2026-06-19
Introduce imported PSBT session state separate from send draft state


jarolrod:
Long term, QML should model two separate flows:

- wallet-created send draft: editable recipients, fee controls, wallet commit
- imported PSBT session: read-only review, save, sign, broadcast

This issue tracks the architecture cleanup needed to prevent send-draft assumptions from leaking into imported PSBT behavior.

Acceptance criteria:

- Imported PSBT state stores:
  - original PSBT
  - current working PSBT
  - save/export PSBT
  - can sign
  - can broadcast
  - review warnings
  - output rows
- Passive review never mutates the save/export PSBT.
- Signing explicitly updates the working/save PSBT.
- Broadcast finalizes/extracts from a temporary copy.
- Wallet commit is only used for wallet-originated send drafts.

Follow-up to https://github.com/bitcoin-core/gui-qml/pull/750

Issue #760, opened 2026-06-19
Warn when imported PSBTs contain positive-value unknown output scripts


jarolrod:
If an imported PSBT contains a positive-value output that cannot be shown as a Bitcoin address and is not unspendable data, the GUI should make that visible before signing or broadcasting.

This should be a non-blocking warning, not a rejection.

Acceptance criteria:

- Positive-value addressless spendable outputs render as `Unknown output script`.
- The review UI shows a warning before signing or broadcasting.
- The warning does not disable Save, Sign, or Broadcast.
- Zero-value `OP_RETURN` outputs do not trigger this warning.
- Add tests for positive-value unknown scripts and zero-value data outputs.

Follow-up to https://github.com/bitcoin-core/gui-qml/pull/750

Issue #759, opened 2026-06-19
Support Imported PSBTs Whose Displayable Outputs Are All Wallet-Owned


jarolrod:
The current send-recipient projection can reject imported PSBTs when all outputs are skipped as wallet-owned/change-like outputs. That is a send-draft assumption leaking into imported PSBT review.

Acceptance criteria:

- Imported PSBTs remain reviewable/saveable even if all outputs are wallet-owned.
- Broadcast eligibility remains based on Core PSBT analysis/finalization and fee checks.
- The review UI clearly shows wallet-owned outputs instead of hiding them all.
- Add a regression test for an imported PSBT whose outputs are all wallet-owned.

Follow-up to https://github.com/bitcoin-core/gui-qml/pull/750

Issue #757, opened 2026-06-19
Add A Read-Only Imported PSBT Output Model


jarolrod:
Imported PSBTs should not depend on editable send-recipient state. Send recipients are draft-construction state; imported PSBT outputs are review state.

Add a PSBT-specific read-only output model for imported PSBT review.

Acceptance criteria:

- Imported PSBT outputs are exposed through a dedicated review model, not `SendRecipient`.
- Output rows support:
  - address outputs
  - wallet-owned/change-like outputs
  - data outputs
  - unknown/addressless scripts
- The model includes amount, display text, script details, output type, and wallet-owned status.
- Imported PSBT review no longer needs every output to be address-extractable.
- Existing wallet-created send draft flow continues using editable recipients.

Follow-up to https://github.com/bitcoin-core/gui-qml/pull/750

PR #769, opened 2026-07-06
qml: Render OP_RETURN outputs in imported PSBT review

### LINKED ISSUE #756: Render OP_RETURN Outputs In Imported PSBT Review
Imported PSBT review currently projects transaction outputs into editable `SendRecipient` rows. That means `OP_RETURN` / data outputs either get skipped or can make otherwise valid PSBTs hard to represent.

For imported PSBTs, the GUI should show data outputs as read-only transaction outputs instead of trying to treat them as recipients.

Acceptance criteria:

- Imported PSBT review shows `OP_RETURN` outputs as `Data output`.
- The row includes amount and script/data details where useful.
- Zero-value `OP_RETURN` outputs do not block import, save, sign, or broadcast.
- Complete broadcastable PSBTs with normal outputs plus `OP_RETURN` outputs remain broadcastable.
- Add model/UI tests for imported PSBTs containing `OP_RETURN`.

Follow-up to https://github.com/bitcoin-core/gui-qml/pull/750

Chigala:
Fixes #756. Follow-up to #750.

While learning the imported PSBT flow I reproduced the OP_RETURN skip on regtest: a PSBT with a recipient plus a `data` output imports fine, signs, broadcasts, and the OP_RETURN lands on-chain,  but the review screen never shows it. This implements what #756 (and the #750 review discussion) asked for: render it as a read-only "Data output" row with the amount and output script hex, so it stops being invisible "without becoming a scary failure."

## screenshot

Screenshot 2026-07-05 at 15 31 48

## Approach

I read through the #750 review thread before writing this, so to be upfront about the design tension: @jarolrod's comments there argue imported PSBT review should eventually stop depending on the editable `SendRecipient` model entirely (#757, #762). I agree,  but since #750 merged with the reduced-scope recipient projection, this PR stays inside that mechanism and just fixes the OP_RETURN rendering for now. 

- `SendRecipient` gains `isDataOutput` / `dataHex`. Data outputs always report valid, so they never block the flow.
- `SendRecipientsListModel` exposes both as roles.
- `tryImportPsbtToReview()` projects zero-value unspendable outputs into data rows instead of moving past them. Since the guard only admits zero-value scripts, the positive-amount check (`recipient_total`) behaves exactly as before.
- `MultipleRecipientsSummary` renders data rows read-only ("Data output" + script hex). `SingleRecipientSummary` is untouched, a lone data output implies a data-only PSBT, which is still rejected (that's #759's territory).

The state is intentionally a thin flag, so when the read-only output model (#757) / imported session state (#762) land, this rendering maps onto them one-to-one. If you'd rather I do those first and have me rebase this on top, or I build the output-row model here instead adding it to this PR, that still works. 

## Testing

- Rewrote `importPsbtFromFile_skipsZeroValueOpReturnOutputs` as `importPsbtFromFile_showsOpReturnAsDataOutput`: asserts the data row's flag, script hex, zero amount and validity, and keeps the existing checks that the PSBT stays broadcastable and the saved PSBT is byte-identical to the imported one.

## Things I hit but left alone (tracked elsewhere)

- Data-only PSBTs still rejected → #759
- Positive-value unknown scripts still rejected (rather than warn) → #760
- Back-navigation from review exposes imported rows as editable drafts, including the data output as a blank entry → #762. it's the draft-vs-session confusion your review described.
- Two small ones that seem unfiled — the review header counts data outputs as recipients ("There are 2 recipients."), and Activity renders OP_RETURN as a blank zero-amount row.

### DIFF
--- a/qml/models/sendrecipient.cpp
+++ b/qml/models/sendrecipient.cpp
@@ -109,6 +109,31 @@ CAmount SendRecipient::cAmount() const
     return m_amount->satoshi();
 }
 
+bool SendRecipient::isDataOutput() const
+{
+    return m_isDataOutput;
+}
+
+QString SendRecipient::dataHex() const
+{
+    return m_dataHex;
+}
+
+void SendRecipient::setDataOutput(const QString& hex)
+{
+    if (!m_isDataOutput) {
+        m_isDataOutput = true;
+        Q_EMIT isDataOutputChanged();
+    }
+    if (m_dataHex != hex) {
+        m_dataHex = hex;
+        Q_EMIT dataHexChanged();
+    }
+    setAddressError("");
+    setAmountError("");
+    Q_EMIT isValidChanged();
+}
+
 void SendRecipient::clear()
 {
     m_label = "";
@@ -116,6 +141,14 @@ void SendRecipient::clear()
     setSubtractFeeFromAmount(false);
     m_address->setAddress("", 0);
     m_amount->clear();
+    if (m_isDataOutput) {
+        m_isDataOutput = false;
+        Q_EMIT isDataOutputChanged();
+    }
+    if (!m_dataHex.isEmpty()) {
+        m_dataHex = "";
+        Q_EMIT dataHexChanged();
+    }
     Q_EMIT addressChanged();
     Q_EMIT labelChanged();
     Q_EMIT messageChanged();
@@ -140,6 +173,11 @@ void SendRecipient::validateAddress()
 
 void SendRecipient::validateAmount()
 {
+    if (m_isDataOutput) {
+        setAmountError("");
+        Q_EMIT isValidChanged();
+        return;
+    }
     if (m_amount->isSet()) {
         if (m_amount->satoshi() <= 0) {
             setAmountError(tr("Amount must be greater than zero"));
@@ -163,5 +201,6 @@ void SendRecipient::validateAmount()
 
 bool SendRecipient::isValid() const
 {
+    if (m_isDataOutput) return true;
     return m_addressError.isEmpty() && m_amountError.isEmpty() && m_amount->satoshi() > 0 && !m_address->isEmpty();
 }
--- a/qml/models/walletqmlmodel.cpp
+++ b/qml/models/walletqmlmodel.cpp
@@ -39,6 +39,7 @@
 #include <script/solver.h>
 #include <support/allocators/secure.h>
 #include <util/result.h>
+#include <util/strencodings.h>
 #include <util/threadnames.h>
 #include <util/translation.h>
 #include <wallet/coincontrol.h>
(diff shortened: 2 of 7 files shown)


hebasto:
Please rebase.

qml/models/sendrecipient.cpp
```diff
@@ -163,5 +201,6 @@ void SendRecipient::validateAmount()
 
 bool SendRecipient::isValid() const
 {
+    if (m_isDataOutput) return true;
```
151henry151, line 204: This clears the `isValid()` gate, then `allValid()` / `validationError()` still `DecodeDestination()` the empty address. Two data outputs collide as `CNoDestination` and fail with `Recipient addresses must be unique`. Skip data outputs in those checks.

qml/models/walletqmlmodel.cpp
```diff
@@ -2004,13 +2005,18 @@ bool WalletQmlModel::tryImportPsbtToReview(const PartiallySignedTransaction& psb
...
     };
     std::vector<DraftRecipient> draft_recipients;
     CAmount recipient_total{0};
     for (const CTxOut& output : analysis_psbt.tx->vout) {
         CTxDestination destination;
         if (!ExtractDestination(output.scriptPubKey, destination)) {
             if (output.nValue == 0 && output.scriptPubKey.IsUnspendable()) {
+                draft_recipients.push_back({QString{}, QString{}, output.nValue,
+                                            /*is_data_output=*/true,
+                                            QString::fromStdString(HexStr(output.scriptPubKey))});
```
151henry151, line 2019: This is the full script, so the row shows `6a020102` (opcode and push length) rather than the payload `0102`. The new test asserts `6a020102`, which pins that. Is showing the whole script deliberate, or should the row show the payload?

qml/models/walletqmlmodel.cpp
```diff
@@ -2004,13 +2005,18 @@ bool WalletQmlModel::tryImportPsbtToReview(const PartiallySignedTransaction& psb
...
+        bool is_data_output{false};
+        QString data_hex{};
     };
     std::vector<DraftRecipient> draft_recipients;
     CAmount recipient_total{0};
     for (const CTxOut& output : analysis_psbt.tx->vout) {
         CTxDestination destination;
         if (!ExtractDestination(output.scriptPubKey, destination)) {
             if (output.nValue == 0 && output.scriptPubKey.IsUnspendable()) {
+                draft_recipients.push_back({QString{}, QString{}, output.nValue,
```
151henry151, line 2017: These used to continue without being added, so they did not count toward the 25-recipient cap below. Should they count now, or stay excluded?

151henry151 (commented):
Concept ACK.

Rebased 2baa7f9cf onto bc4f688. The QML commit, 3c7493c5f, conflicts in MultipleRecipientsSummary.qml with 1a7c480 and 3678c40 from #860. Those dropped primaryText/secondaryText and the secondary TextArea, so there is nothing left to attach the data-output row to. A naive rebase imports the OP_RETURN and renders an empty addressText at amount 0.

Issue #576, opened 2026-05-06
Epic: Run and Monitor Your Node


johnny9:
## Goal
Track v1.0 work for node status, sync progress, connectivity, peers, network traffic, debug log, RPC console, warnings, and monitoring tools.

## Release Target Policy
- **Preview**: visible node status and peer/network controls are truthful and safe to use. User-visible warning states and peer actions must not fail silently. No documentation requirement.
- **Beta**: sync monitoring, peer monitoring, network traffic, debug log, mempool information, and console flows selected for release have coherent UI states and baseline documentation.
- **v1.0**: final node diagnostic parity decisions are made, final screenshots are captured, and any lower-priority Qt widget diagnostics are explicitly deferred.
- **Post-v1.0**: lower-priority Qt debug-window table/detail polish that is useful but not release-blocking.

## User Flows
- Monitor chain sync, block status, connection state, and warnings.
- Pause and resume network activity.
- Review storage, data directory, blocks directory, prune status, and related node configuration where selected for the release target.
- View peers, peer details, disconnect peers, ban peers, and unban peers.
- View network traffic, byte totals, and graph history.
- View mempool transaction count and memory usage.
- Use node diagnostics, debug log, and RPC console.

## Definition of Done
- Each child issue identifies a concrete implementation, documentation, or parity gap in the current QML app.
- Each child issue has a clear target: Preview, Beta, v1.0, or Post-v1.0.
- Preview-visible node monitoring and peer control flows do not misrepresent node state or silently lose failures.
- The Beta user guide has a top-level `Run and Monitor Your Node` page.
- The v1.0 guide includes steps, screenshots from the QML app, expected results, and important failure/recovery notes for the final flow set.
- Remaining Qt widget parity gaps are tracked as child issues or intentionally deferred.

PR #803, opened 2026-07-19
Improve block clock rendering and sync state

### LINKED ISSUE #227: Refactor ChainModel
- extract block sync verfication progress property
- expose block information as ChainModel property/methods and move time calculations to blockclockdial as the context there is more appropriate for those methods.

### LINKED ISSUE #743: Block Clock Header Pre-sync is not showing progress dial
Image

### LINKED ISSUE #785: Miniblock clock gradient is backwords
Image

johnny9:
## Summary

This PR fixes how the block clock represents sync/catch-up state and how confirmation history renders in both the full and mini clocks.

- Keep the UI in sync mode until Core has left IBD and both headers and the active chain have caught up. Near-tip catch-up remains at 99.9% instead of prematurely showing the normal clock.
- Render rounded confirmation segments without losing confirmation depth when closely spaced block boundaries coalesce, and use one continuous gradient across combined blocks.
- Correct the mini block clock gradient so clockwise time moves from older, higher-confirmation yellow toward newer, zero-confirmation red.
- Move timeline ownership from `ChainModel` to `BlockClockModel` and stop presentation work while the clock is not visible. Hidden pages or tabs—and hidden or minimized windows—do not schedule dial repaints or run dial, peer-indicator, or estimating animations. The backing model stays current with minute-aligned clock updates and an immediate refresh when a block arrives, so the clock renders the latest state when it becomes visible again.

### Catch-up sync-state bug

The unfixed build briefly shows the normal red clock while the active chain is still rapidly catching up from a stale signet tip. The new sync-state latch keeps the UI in syncing state at 99.9% until both header and block catch-up are complete.

https://github.com/user-attachments/assets/537fae60-f8ff-49b7-874c-8f497da66b03

### Confirmation segment rendering

| Before | After |
| --- | --- |
| Before: separated confirmation segments | After: rounded coalesced segments with a continuous gradient |
| Every block was drawn individually. When several boundaries landed only a few pixels apart, the segments crowded together and their confirmation-color changes became difficult to see. | Rounded segments coalesce when necessary while preserving confirmation depth and a continuous gradient. |

### Mini block clock gradient

The mini clock now uses the same direction as the full clock: older, higher-confirmation time is yellow, transitioning clockwise toward the newest, zero-confirmation time in red.

Fixed mini block clock gradient

## Motivation

`verificationProgress &gt; 0.999` can report the normal clock too early. With a saved signet tip at height 313768, Core had already left IBD at 0.999830; once a peer announced height 313904, the UI showed a normal clock while 136 blocks rapidly caught up.

The existing timeline also mixed model data with paint geometry, kept repaint requests, timers, and animations active while views were hidden, and lost confirmation transitions when rounded segments were too close to render separately.

## Related issues

Fixes #227.
Fixes #743.
Fixes #785.

Related to #576.
Follow-up to #609.

## Testing

- Built `bitcoin-core-app`, `bitcoinqml_unit_tests`, and `bitcoinqml_qmltests`.
- Ran `QT_QPA_PLATFORM=offscreen QT_QUICK_BACKEND=software LIBGL_ALWAYS_SOFTWARE=1 ctest --test-dir build --output-on-failure` (2/2 passed).
- Built each commit in the branch independently.
- Ran Bitcoin lint checks for commit messages, tabs/whitespace, and trailing newlines.
- Ran `git diff --check`.
- Verified visibility gating in the C++ and QML tests: inactive clocks stop dial timers, estimating animation, peer-indicator animation, and repaint scheduling while model state remains current for reactivation.
- Added `BlockClockModel` coverage for minute alignment and block-triggered time refresh (`BlockClockModelTests`: 10/10 passed), and reran the QML suite (389/389 passed).
- Manually reproduced stale-signet catch-up with the unfixed and fixed builds.

### DIFF
--- a/qml/bitcoin.cpp
+++ b/qml/bitcoin.cpp
@@ -40,6 +40,7 @@
 #include <qml/models/banlistmodel.h>
 #include <qml/models/bitcoinaddress.h>
 #include <qml/models/bitcoinurimodel.h>
+#include <qml/models/blockclockmodel.h>
 #include <qml/models/bumptransactionmodel.h>
 #include <qml/models/chainmodel.h>
 #include <qml/models/debuglogmodel.h>
@@ -602,9 +603,11 @@ int QmlGuiMain(int argc, char* argv[])
     AndroidNotifier android_notifier{node_model};
 #endif
 
-    ChainModel chain_model{*chain};
-    chain_model.setCurrentNetworkName(QString::fromStdString(gArgs.GetChainTypeString()));
-    setupChainQSettings(&app, chain_model.currentNetworkName());
+    ChainModel chain_model{QString::fromStdString(gArgs.GetChainTypeString())};
+    BlockClockModel block_clock_model{[chain = chain.get()](qint64 period_start, qint64 period_end) {
+        return LoadBlockClockHistory(*chain, period_start, period_end);
+    }};
+    setupChainQSettings(&app, chain_model.networkName());
     // Settings reset must happen before model instantiation so the models
     // read clean defaults from QSettings.
     if (gArgs.IsArgSet("-resetguisettings")) {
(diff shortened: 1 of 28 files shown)


hebasto:
Please rebase.

Issue #784, opened 2026-07-16
Debug log is blocking the main thread


johnny9:
This is causing performance issues throughout the application after the settings layout changes

johnny9:
Was not due to blocking but doing too much rendering while not visible. will add an issue to explain the optimizations to make.

Issue #768, opened 2026-06-27
Debug log relative-time labels stop refreshing when only the age changes


johnny9:
## Problem

PR #746 adds the debug log viewer's relative-time column and wires a timer to refresh labels such as `just now`, `1 min ago`, and `2 min ago`.

The intended refresh path exists, but role-only updates are currently dropped:

- `SettingsDebugLog.qml` has a 60s `Timer` that calls `debugLogModel.updateRelativeTimes()`.
- `DebugLogModel::updateRelativeTimes()` recomputes `LogLine::relativeTime` in `m_all_lines`, then calls `buildDisplayLines()`.
- `buildDisplayLines()` returns early when `filtered == m_display_lines`.
- `LogLine::operator==` deliberately excludes `relativeTime`, so rows whose only changed value is `RelativeTimeRole` compare equal.
- Because of that early return, `m_display_lines` is not updated and no `dataChanged`/`modelReset` signal is emitted.

Result: the visible right-column age labels can stay stale until some unrelated model change happens, such as a real log append, filter change, or manual refresh.

This was noticed while reviewing #746. It is also mentioned in the debug-log follow-up discussion here: https://github.com/bitcoin-core/gui-qml/pull/746#discussion_r3453082577

## Manual test case

With PR #746 running on regtest and the Debug log page visible:

```bash
LOG="$HOME/.bitcoin/regtest/debug.log"
MARK="relative-time-test-$(date +%s)"
printf '%s %s\n' "$(date -u -d '30 seconds ago' '+%Y-%m-%dT%H:%M:%SZ')" "$MARK" >> "$LOG"
echo "$MARK"
```

Then:

1. Paste the printed marker into the Debug log search field.
2. Confirm the row appears with the right column showing `just now`.
3. Wait 90 seconds without clicking Refresh and without changing the search text.
4. Expected: the right column changes to `1 min ago`.
5. Actual: the right column remains `just now`.
6. Clicking Refresh, or changing the search text and changing it back, makes the label jump to the current age. That shows the age calculation works, but the timer-driven model notification path does not update the visible row.

This is easiest to reproduce when the node is otherwise idle. New log writes can mask the problem by causing a real model refresh.

## Possible regression test

A focused model test can exercise the role-only update path directly. For example, add a test that writes a timestamped row close to the minute boundary, loads it, waits until the label should change, calls `updateRelativeTimes()`, and checks that `RelativeTimeRole` changes without requiring a full content reset.

Sketch:

[...]

PR #804, opened 2026-07-19
Improve debug log performance

johnny9:
## Summary

- Load and watch the debug log only while its settings page is selected.
- Read the bounded log tail on a worker thread and process ordinary appends incrementally instead of resetting every row.
- Limit individual physical log lines in the viewer to 1 MiB; oversized lines are skipped until their terminating newline without modifying `debug.log`.
- Initially load the newest 1,000 entries and expose older entries in 1,000-row increments through **Load more**.
- Replace the fully instantiated output with a virtualized `ListView`.
- Preserve the visible row and pixel offset across live prepends, tail pruning, and load-more appends.

## Motivation

The previous debug-log view instantiated every displayed row and rebuilt the full output whenever the model reset. A new log write could therefore trigger another full file read, model reset, and delegate rebuild, including while the debug-log page was not selected.

This change treats the model and view as one performance path: the model reads only the new bytes and publishes changed rows during normal appends, while the view instantiates only visible delegates and keeps its scroll anchor stable.

Follow-up to the performance problem described in #784 and the original viewer in #532. Contributes to #576.

Related: #768. Its targeted relative-time role update behavior is already present on `qt6` and remains intact after this refactor; this PR does not close that issue.

## Testing

- Built every commit in the branch independently.
- Built the final application with the matching non-OpenGL depends toolchain.
- Ran Bitcoin commit-message and whitespace checks on every commit.
- Ran Ruff 0.15 and `qmllint` on the changed Python and QML files.
- Ran the complete C++ unit suite: 540 passed, including `DebugLogModelTests`: 19 passed.
- Ran the complete QML suite: 391 passed; the remaining `Send::test_send_uri_import_schedules_fee_estimate` failure (`BitcoinUri is not defined`) reproduces on the PR base.
- Ran `test/functional/qml_test_debug_log.py` against the real application.
- Verified watcher-driven updates remain capped, offscreen rows are virtualized, and **Load more** adds older rows without moving the viewport.

### DIFF
--- a/qml/models/debuglogmodel.cpp
+++ b/qml/models/debuglogmodel.cpp
@@ -97,17 +97,43 @@ void DebugLogModel::setLoadLimit(int limit)
     Q_EMIT loadLimitChanged();
 }
 
+void DebugLogModel::setActive(bool active)
+{
+    if (m_active == active || m_stopping) return;
+
+    m_active = active;
+    ++m_activation_generation;
+    Q_EMIT activeChanged();
+
+    const QString path_str = QString::fromStdString(m_log_path.utf8string());
+    if (m_active) {
+        if (!path_str.isEmpty() && !m_watcher.files().contains(path_str)) {
+            m_watcher.addPath(path_str);
+        }
+        refresh(/*full_load=*/true);
+        return;
+    }
+
+    m_debounce.stop();
+    const auto watched_files = m_watcher.files();
+    if (!watched_files.isEmpty()) {
+        m_watcher.removePaths(watched_files);
+    }
+    m_refresh_pending = false;
+    m_pending_full_load = false;
+}
+
 void DebugLogModel::setFilter(const QString& filter)
 {
     if (m_filter == filter) return;
     m_filter = filter;
     Q_EMIT filterChanged();
-    buildDisplayLines();
+    if (m_active) buildDisplayLines();
 }
 
 void DebugLogModel::refresh(bool full_load)
 {
-    if (m_stopping) return;
+    if (!m_active || m_stopping) return;
 
     // Single-read-in-flight guard. If a read is already running, fold this
     // request into a trailing re-run rather than piling another job onto the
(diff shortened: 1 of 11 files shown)


pseudoramdom:
tested ACK e68a1b55476560bc29be808b12333e9148f2e612 that the performance is smoother now

qml/models/debuglogmodel.cpp
```diff
@@ -252,88 +344,218 @@ void DebugLogModel::stop()
...
+    qint64 cursor = result.file_size;
+    QByteArray suffix;
+    bool found_newline{false};
+    while (cursor > 0) {
+        if (cancelled.load(std::memory_order_relaxed)) return {};
+        const qint64 start = std::max(qint64{0}, cursor - TAIL_READ_BLOCK_SIZE);
+        if (!file.seek(start)) return {};
+        const QByteArray chunk = file.read(cursor - start);
+        if (chunk.size() != cursor - start) return {};
+        const QByteArray data = chunk + suffix;
```
pseudoramdom: Should we limit how much of an unfinished line we keep? 
For a very long unfinished log line, each block is added to and copies the entire accumulated suffix, which could use a lot of memory (and CPU?)

johnny9: I think this issue should be resolved now.

pseudoramdom:
Mostly good to go.
What are your thoughts on using monospace font for the logs
<details>
<summary>Before</summary>
Screenshot 2026-07-29 at 11 46 07 AM

</details>

<details>
<summary>After</summary>

Screenshot 2026-07-29 at 11 49 22 AM

</details>

pseudoramdom:
ACK fb87a8bf293e99e507fa8d3830f44bd3934e7d13

johnny9:
> Mostly good to go. What are your thoughts on using monospace font for the logs
> Before
> Screenshot 2026-07-29 at 11 46 07 AM
> After
> Screenshot 2026-07-29 at 11 49 22 AM

I think monospace can be scary so I think we're supposed to avoid it as much as possible and in the debug log the kerning that you get from a regular font lets us save space so it helps keep it dense and readable. 

I think the debug log could potentially use a layout/size adjustment, though.

PR #805, opened 2026-07-20
Avoid hidden network traffic graph updates

### LINKED ISSUE #611: Network traffic: harden graph updates and preserve range state
## Scope split

The graph reset behavior originally tracked here was split into #806 so it can be implemented and reviewed independently. This issue no longer includes reset behavior.

This issue is limited to the network traffic sampling lifecycle, suppression of hidden-page work, range preservation, and safe shutdown covered by PR #805.

## Code Gap

The QML network traffic page shows received/sent totals and graph history, but its updater has release risks.

`NetworkTrafficTower` creates a timer/thread without clear ownership or shutdown cleanup. It can continue deriving and publishing graph updates when the page is hidden, and the existing range handling needs to preserve totals and retained history consistently.

## Work Needed

- Make network traffic sampling thread/timer ownership explicit and safe during shutdown.
- Ensure graph updates do not leak timers/threads or update QML from the wrong thread.
- Avoid deriving or publishing graph snapshots while the page is inactive, while retaining the raw samples needed when it becomes active again.
- Preserve totals and graph history consistently when changing ranges.
- Add coverage for opening/closing the page, activation changes, changing ranges, and shutdown while sampling is active.

## Related

- Reset behavior was split into #806.
- Implementation: PR #805.

### LINKED ISSUE #801: `LineGraph::fillColor` shadows a member of the base class
```console
$ ./build/bin/bitcoin-core-app -signet
...
qt.qml.propertyCache.append: Member fillColor of the object LineGraph overrides a member of the base object. Consider renaming it or adding final or override specifier
...
```

johnny9:
## What changed

- move network traffic sampling, smoothing, and history retention onto an owned worker thread
- publish immutable traffic snapshots back to the GUI thread only while the Network Traffic page is active
- retain raw samples while hidden and rebuild the requested range when the page becomes active again
- lazy-load the Network Traffic settings page so its graphs do not exist while another settings section is selected
- preserve byte totals as 64-bit values
- rename the native graph gradient property so it no longer shadows `QQuickPaintedItem::fillColor`
- add C++ and QML coverage for thread affinity, hidden-page suppression, retained history, range changes, page lifecycle, and destruction while active

## Why

The previous timer was moved to a background thread, but its timeout updated a GUI-thread model every second regardless of whether the Network Traffic page was visible. That kept smoothing, full-history copies, signal delivery, and graph bindings active in the background. Timer/thread ownership was also not explicit during shutdown.

This keeps the inexpensive raw sample collection running off-thread while avoiding hidden GUI work and making worker shutdown deterministic.

`LineGraph` also declared its own `fillColor` property over the property inherited from `QQuickPaintedItem`, producing a QML property-cache warning. The native property is now named `gradientColor`; the public `NetworkTrafficGraph.fillColor` API and rendering behavior are unchanged.

## User impact

Opening Network Traffic still shows the samples collected while the page was hidden, and changing the graph range rebuilds the visible history without losing totals. Other settings pages no longer pay for traffic graph updates or retain the graph objects.

## Issue

Fixes #611.

Fixes #801.

Graph reset behavior was split from #611 into #806 and is intentionally not included here.

## Validation

- full RelWithDebInfo build with app tests enabled
- `qmllint -I qml -I test/mocks/qml qml/components/NetworkTrafficGraph.qml qml/pages/node/NetworkTraffic.qml qml/pages/node/NodeSettings.qml test/qml/tst_nodesettings.qml`
- `bitcoinqml_qmltests`: 380 passed, 0 failed
- `NetworkTrafficTowerTests`: 6 passed, 0 failed
- `git diff --check origin/qt6...HEAD`

The aggregate unit executable was also run locally. Its Network Traffic suite passed; unrelated existing OptionsModel tests could not write the default Bitcoin data directory inside the managed filesystem sandbox.

### DIFF
--- a/qml/bitcoin.cpp
+++ b/qml/bitcoin.cpp
@@ -592,7 +592,7 @@ int QmlGuiMain(int argc, char* argv[])
     }, Qt::QueuedConnection);
     QObject::connect(&init_executor, &QmlInitExecutor::runawayException, &node_model, &NodeModel::handleRunawayException);
 
-    NetworkTrafficTower network_traffic_tower{node_model};
+    NetworkTrafficTower network_traffic_tower{*node};
     NetworkStatusModel network_status_model;
 #ifdef __ANDROID__
     AndroidNotifier android_notifier{node_model};
--- a/qml/components/NetworkTrafficGraph.qml
+++ b/qml/components/NetworkTrafficGraph.qml
@@ -14,7 +14,7 @@ Item {
     id: root
     property alias backgroundColor: trafficGraph.backgroundColor
     property alias borderColor: trafficGraph.borderColor
-    property alias fillColor: trafficGraph.fillColor
+    property alias fillColor: trafficGraph.gradientColor
     property alias lineColor: trafficGraph.lineColor
     property alias markerLineColor: trafficGraph.markerLineColor
     property alias maxSamples: trafficGraph.maxSamples
@@ -31,7 +31,7 @@ Item {
         width: root.width
         backgroundColor: root.backgroundColor
         borderColor: root.borderColor
-        fillColor: root.fillColor
+        gradientColor: root.fillColor
         lineColor: root.lineColor
         markerLineColor: root.markerLineColor
         maxSamples: root.maxSamples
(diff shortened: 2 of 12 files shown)


jarolrod (approved):
ACK 35fecc9

Sampling stays on the worker, GUI publish is gated on page activity

Tested manually

PR #809, opened 2026-07-21
Close console autocomplete when leaving Console tab

### LINKED ISSUE #782: Close RPC console autocomplete when changing desktop tabs
## Problem

In the desktop wallet view, enter a partial RPC command such as `getblock` to open the console autocomplete popup, then select another top-level tab. The popup remains visible above the newly selected view.

`DesktopWallets` keeps `CommandConsole` alive as a `StackLayout` child. The autocomplete is a `Popup`, whose current dismissal paths are Escape, focus loss, and input changes; switching the top-level tab does not notify the console that it was left.

## Steps to reproduce

1. Start the desktop wallet view and select the Console tab.
2. Type a command prefix with available matches, for example `getblock`.
3. Confirm the autocomplete popup is open.
4. Select Activity, Settings, or another top-level tab.

**Actual:** The autocomplete popup remains visible over the selected tab.

**Expected:** The popup closes when leaving the Console tab.

## Proposed fix

Expose tab changes through the common navigation bar API. The signal should carry the previous and newly selected tab controls, e.g. `tabChanged(previousTab, currentTab)`, so a child view can react to leaving its own tab without depending on a numeric index.

`CommandConsole` can receive its associated navigation tab and call `autocompletePopup.close()` when `previousTab` is that tab. This keeps transient-UI cleanup local to the owning view while providing a reusable tab-transition hook for other persistent desktop views.

## Test coverage

Add a QML regression test that opens `consoleAutocompletePopup`, switches from Console to another desktop tab, and verifies the popup is no longer visible.

A proposed implementation is available on `johnny9:fix/console-autocomplete-tab-switch`.

uqlidi:
The RPC console lives as a persistent StackLayout child and its
autocomplete Popup renders in the window overlay layer, so hiding
the console on a tab switch does not hide the popup. It lingers over
the newly selected tab.

Add a `tabActive` property to CommandConsole, bound by DesktopWallets
to `consoleTabButton.checked`, and close the popup when the tab is
left. Keyed off the tab ButtonGroup rather than StackLayout visibility,
which never reports the transition.

fix #782

### DIFF
--- a/qml/pages/node/CommandConsole.qml
+++ b/qml/pages/node/CommandConsole.qml
@@ -30,6 +30,14 @@ Page {
     property string commandDraft: ""
     property string searchDraft: ""
 
+    // True while this view's tab is the selected one. As a persistent StackLayout
+    // child, the console is never destroyed on tab changes, and its autocomplete
+    // Popup renders in the window overlay layer, so it is not hidden along with
+    // the view. The embedder binds this to its tab's checked state; when the tab
+    // is left, dismiss the popup so it cannot linger over the newly selected tab.
+    property bool tabActive: true
+    onTabActiveChanged: if (!tabActive) autocompletePopup.close()
+
     function _pushPalette() {
         rpcConsoleModel.requestColor = consoleRequestColor
         rpcConsoleModel.replyColor   = consoleReplyColor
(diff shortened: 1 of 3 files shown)


johnny9 (approved):
ACK 50218387c6aa5600cf5b6e209016f0ed64071983

jarolrod (approved):
ACK https://github.com/bitcoin-core/gui-qml/commit/50218387c6aa5600cf5b6e209016f0ed64071983

hebasto:
This PR is merged.

Issue #775, opened 2026-07-08
Unreadable custom datadir can cause and exception


johnny9:
Still finishing the rest of my review but I did find a blocker so posting it now.

On Linux, if the saved GUI-selected custom datadir still exists but is no longer readable, the app aborts at launch with an unhandled exception instead of re-showing the datadir chooser:

terminate called after throwing an instance of 'std::filesystem::__cxx11::filesystem_error'
  what():  filesystem error: status: Permission denied [<custom-datadir>/bitcoin.conf]
Aborted (core dumped)

Repro:
1. Fresh config dir, onboard choosing a custom data directory.
2. chmod 000 <custom-datadir> (or any state where the directory exists but cannot be traversed, for example an external drive that remounts root-owned).
3. Relaunch. Expected: the storage location chooser reappears. Actual: abort before any window, on every launch, until the user fixes permissions by hand or passes -choosedatadir.

Root cause: ResolveOnboardingStartupStatus correctly computes force_show_onboarding (ShouldShowDataDirChooser flags the directory as not writable), but then still reads the profile from it, because the gate is custom_datadir_exists = QFileInfo::exists(...) (qml/onboarding_settings.cpp, around line 406), and a mode-000 directory exists. It applies the datadir to the preview args and calls ReadConfigFiles, where fs::is_directory(conf_path) (src/common/config.cpp:134) stats <datadir>/bitcoin.conf, gets EACCES, and throws. Nothing on the QML startup path catches std::filesystem_error, so the process terminates.

Suggested fix, two parts:
1. Gate the profile read on the datadir being usable, not merely existing: the ValidateCustomDataDir result is already computed for force_show_onboarding two lines above, so can_read_profile should exclude a custom datadir that failed validation.
2. Wrap the preview's config/settings read in a try/catch that turns fs::filesystem_error into status.error, so any remaining gap of this kind degrades to the onboarding error path instead of std::terminate.

The existing unit test covers the unwritable case (guiDataDirChooserShowsForUnwritableConfiguredDir) but not the unreadable one, which is why this survives a green suite; a sibling test with a mode-000 directory would pin the fix.

_Originally posted by @epicleafies in https://github.com/bitcoin-core/gui-qml/issues/742#issuecomment-4905653972_

epicleafies:
Flagging that this is a hard crash, not the soft failure the description implies: a GUI-saved custom datadir that exists but cannot be traversed aborts startup before the chooser is shown.

[...]

PR #812, opened 2026-07-28
Stabilize pre-init datadir and settings resolution

### LINKED ISSUE #773: Pre-init onboarding can reread config after mutating startup args
## Summary

Pre-init onboarding can mutate the live startup args before normal startup calls `InitConfig()`. In the default config-file case, that can make the app reread a different `bitcoin.conf` after onboarding completes and exit before the main window opens.

## Why this matters

The onboarding preview/apply path already reads `bitcoin.conf` and applies the active datadir to `gArgs`. `QmlGuiMain()` then continues and calls `InitConfig(gArgs)` again. If the first config read came from the default config location and contained `datadir=/tmp/B`, the second config lookup is based on the mutated datadir and can look under `/tmp/B` instead of the original default config location.

This can turn a successful onboarding flow into a startup failure after the pre-init window closes.

## Reproduction

Use a sandboxed home so the real profile is not touched:

```bash
APP=/tmp/gui-qml-pr742-updates-build/bin/bitcoin-core-app
REPRO=/tmp/gui-qml-config-reread-repro
B=/tmp/B

rm -rf "$REPRO" "$B"
mkdir -p "$REPRO/.bitcoin" "$B"

cat > "$REPRO/.bitcoin/bitcoin.conf" <<EOF
regtest=1
datadir=$B
EOF

HOME="$REPRO" "$APP" -regtest
```

Steps:

1. Complete pre-init onboarding.
2. Observe that the app closes/exits instead of loading the main window.

Expected behavior:

After onboarding completes, normal startup should continue using the same resolved startup/config state and load the main window.

Actual behavior:

Onboarding completes, but normal startup rereads config through mutated `gArgs`, can resolve the config path differently, and exits before `MainWindow.qml` is loaded.

## Suggested fix direction

Keep pre-init onboarding from mutating the live startup `gArgs`, or reparse clean startup args before the real `InitConfig()` call. Add a regression for the no-`-conf` path where the default `bitcoin.conf` contains `datadir=/tmp/B`.

### LINKED ISSUE #774: -resetguisettings can reset the wrong profile before datadir resolution
## Summary

`-resetguisettings` is applied before `bitcoin.conf` has been read and before the active datadir/profile is known. As a result, opening onboarding can reset the wrong profile.

## Why this matters

If the default config redirects to another datadir, the app still clears GUI settings for the default/main profile first. Finishing onboarding may later reset the redirected profile too, but canceling onboarding leaves the default/main profile modified even though it was not the active profile.

## Reproduction

Use a sandboxed home so the real profile is not touched:

```bash
APP=/tmp/gui-qml-pr742-updates-build/bin/bitcoin-core-app
REPRO=/tmp/gui-qml-resetguisettings-repro
B=/tmp/B

rm -rf "$REPRO" "$B"
mkdir -p "$REPRO/.bitcoin/regtest" "$B/regtest"

cat > "$REPRO/.bitcoin/bitcoin.conf" <<EOF
regtest=1
datadir=$B
EOF

cat > "$REPRO/.bitcoin/regtest/settings.json" <<EOF
{"server":true,"proxy":"10.0.0.1:9050","qml_onboarded":true}
EOF

cat > "$B/regtest/settings.json" <<EOF
{"server":true,"proxy":"10.0.0.2:9050","qml_onboarded":true}
EOF

HOME="$REPRO" "$APP" -regtest -resetguisettings -disablewallet
```

Cancel path:

1. Wait for onboarding to appear.
2. Close/cancel onboarding.
3. Inspect:

```bash
cat "$REPRO/.bitcoin/regtest/settings.json"
cat "$B/regtest/settings.json"
ls "$REPRO/.bitcoin/regtest/"*.bak "$B/regtest/"*.bak 2>/dev/null
```

Expected behavior:

Canceling onboarding should not reset the default/main profile before the active redirected profile has been resolved.

Actual behavior:

The default/main profile is already reset/backed up just by reaching onboarding.

Finish path:

1. Restore both `settings.json` files from the setup above.
2. Launch the same command.
3. Complete onboarding.
4. Inspect both profiles.

Expected behavior:

Only the resolved active profile should be reset.

Actual behavior:

The default/main profile is reset first, and the redirected profile can be reset later when onboarding finishes.

## Suggested fix direction

Defer the GUI-settings reset until after the active profile is resolved. Add regression coverage for both canceling and completing onboarding when the default config redirects to another datadir.

jarolrod:
This is a follow-up to #742.

Pre-init onboarding was doing too much work against the live `gArgs`. It read config and settings to build the preview, then normal startup called `InitConfig()` on the same `ArgsManager`. If `bitcoin.conf` redirected the datadir, the second config read could start from a different place and fail before the main window opened.

`-resetguisettings` had a similar ordering problem. We could reset the bootstrap profile before Core had resolved the datadir and network it was actually going to use.

This switches the pre-init reads to a scratch `ArgsManager` built from the original argv. The datadir selected in onboarding is kept separate from the one resolved through config, and settings changes are finalized after `InitConfig()` and the chain-specific `QSettings` setup.

A few related pieces are handled in the same path:

- `-resetguisettings=0` is treated as false.
- Reset uses the final Core settings precedence.
- `strDataDir` is preserved when GUI settings are cleared.
- Core and GUI settings are backed up and rolled back if a later write fails.
- Settings stores are only opened and synced when they are actually needed.
- An onboarding apply is rejected if the datadir, chain, settings path, or reset value changed while the window was open.
- Reset can recover an unreadable `settings.json` through the existing reset-or-abort flow.

Cancelling onboarding remains non-mutating. No settings are cleared and no backups are created.

Fixes #773.

Fixes #774.

This is related to #600, but does not close it. The Settings-page actions and `bitcoin.conf` opening flow are still separate work.

#775 also remains open. An existing saved datadir that cannot be traversed can still fail before the chooser is shown.

### DIFF
--- a/qml/bitcoin.cpp
+++ b/qml/bitcoin.cpp
@@ -69,6 +69,7 @@
 #endif
 #include <util/fs.h>
 #include <util/fs_helpers.h>
+#include <util/string.h>
 #include <util/threadnames.h>
 #include <util/translation.h>
 #ifdef ENABLE_WALLET
@@ -77,6 +78,7 @@
 
 #include <cassert>
 #include <memory>
+#include <optional>
 #include <tuple>
 #include <vector>
 
@@ -89,6 +91,7 @@
 #include <QPixmap>
 #include <QGuiApplication>
 #include <QJSEngine>
+#include <QMessageBox>
 #include <QPointer>
 #include <QQmlApplicationEngine>
 #include <QQmlContext>
@@ -296,8 +299,35 @@ enum class PreInitOnboardingStatus {
     FAILED,
 };
 
+bool ErrorSettingsRead(const bilingual_str& error, const std::vector<std::string>& details)
+{
+    if (gArgs.GetBoolArg("-resetguisettings", false)) {
+        return false;
+    }
+
+    QMessageBox message_box{
+        QMessageBox::Critical,
+        CLIENT_NAME,
+        QString::fromStdString(strprintf("%s.", error.translated)),
+        QMessageBox::Reset | QMessageBox::Abort,
+    };
+    message_box.setInformativeText(QObject::tr("Do you want to reset settings to default values, or to abort without making changes?"));
+    message_box.setDetailedText(QString::fromStdString(util::MakeUnorderedList(details)));
+    message_box.setTextFormat(Qt::PlainText);
+    message_box.setDefaultButton(QMessageBox::Reset);
+    switch (message_box.exec()) {
+    case QMessageBox::Reset:
+        return false;
+    case QMessageBox::Abort:
+        return true;
+    default:
+        assert(false);
+    }
+}
+
 struct PreInitOnboardingContext {
     std::unique_ptr<OnboardingOptionsModel> onboarding_options_model;
+    std::optional<QmlOnboardingSettings::PendingApply> pending_apply;
     QScopedPointer<const NetworkStyle> network_style;
     std::unique_ptr<QQmlApplicationEngine> engine;
 #ifdef ENABLE_TEST_AUTOMATION
@@ -324,6 +354,9 @@ bool ShouldShowPreInitOnboarding(const std::vector<std::string>& argv, bool can_
     const QmlOnboardingSettings::OnboardingStartupStatus status{
         QmlOnboardingSettings::ResolveOnboardingStartupStatus(argv, can_listen_ipc)
     };
+    if (status.settings_file_unreadable) {
+        return false;
+    }
     return !status.ok || status.should_show_onboarding;
 }
 
@@ -380,11 +413,13 @@ PreInitOnboardingStatus RunPreInitOnboarding(PreInitOnboardingContext& context, 
     }
 
     QString error;
-    if (!context.onboarding_options_model->applyToArgs(gArgs, &error)) {
+    QmlOnboardingSettings::PendingApply pending_apply;
+    if (!context.onboarding_options_model->prepareApplyToArgs(gArgs, pending_apply, &error)) {
         InitError(Untranslated(error.toStdString()));
         context.close();
         return PreInitOnboardingStatus::FAILED;
     }
+    context.pending_apply = std::move(pending_apply);
     return PreInitOnboardingStatus::COMPLETED;
 }
 } // namespace
@@ -465,13 +500,9 @@ int QmlGuiMain(int argc, char* argv[])
 
     app.setQuitOnLastWindowClosed(false);
     setupChainQSettings(&app, QString::fromStdString(gArgs.GetChainTypeString()).toUpper());
-    if (gArgs.GetBoolArg("-resetguisettings", false)) {
-        QString reset_error;
-        if (!QmlDataDir::ResetGuiSettings(gArgs, &reset_error)) {
-            InitError(Untranslated(reset_error.toStdString()));
-            return EXIT_FAILURE;
-        }
-    }
+    const QmlOnboardingSettings::GuiSettingsStore bootstrap_gui_settings{
+        QmlOnboardingSettings::CurrentGuiSettingsStore()
+    };
 
     LoadFontResource(":/fonts/bitcoincoresans/regular");
     LoadFontResource(":/fonts/bitcoincoresans/semibold");
@@ -512,25 +543,20 @@ int QmlGuiMain(int argc, char* argv[])
 
     if (auto error = common::InitConfig(
             gArgs,
-            [](const bilingual_str& msg, const std::vector<std::string>& details) {
-                return InitError(msg, details);
-            })) {
+            ErrorSettingsRead)) {
         return EXIT_FAILURE;
     }
 
-    const QmlLegacySettings::MigrationResult legacy_migration{
-        QmlLegacySettings::MigrateCoreSettings(gArgs, QmlLegacySettings::MigrationMode::Persist)
-    };
-    if (!legacy_migration.error.isEmpty()) {
-        InitError(Untranslated(legacy_migration.error.toStdString()));
+    setupChainQSettings(&app, QString::fromStdString(gArgs.GetChainTypeString()).toUpper());
+    QString finalize_settings_error;
+    if (!QmlOnboardingSettings::FinalizeStartupSettings(
+            gArgs,
+            bootstrap_gui_settings,
+            pre_init_onboarding_context.pending_apply ? &*pre_init_onboarding_context.pending_apply : nullptr,
+            /*result=*/nullptr,
+            &finalize_settings_error)) {
+        InitError(Untranslated(finalize_settings_error.toStdString()));
         return EXIT_FAILURE;
-    }
-    if (legacy_migration.settings_changed) {
-        std::vector<std::string> settings_errors;
-        if (!gArgs.WriteSettingsFile(&settings_errors)) {
-            InitError(_("Settings file could not be written"), settings_errors);
-            return EXIT_FAILURE;
-        }
     }
 
     // legacy GUI: parameterSetup()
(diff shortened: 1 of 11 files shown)


qml/onboarding_settings.cpp
```diff
@@ -105,16 +189,37 @@ bool ReadConfigAndSelectNetwork(ArgsManager& args, QString* error)
...
+struct ReadOnlyProfileResult {
+    bool settings_enabled{true};
+    bool config_file_path_available{false};
+    bool settings_file_unreadable{false};
+};
+
+bool ReadResolvedProfile(ArgsManager& args, ReadOnlyProfileResult& result, QString* error)
 {
+    result = {};
+    if (!ReadConfigAndSelectNetwork(args, error)) return false;
```
epicleafies, line 201: This is #775, flagged at the fix point. Worth noting it is a hard crash (uncaught `filesystem_error` -> SIGABRT on a mode-000 saved datadir), not the soft failure the issue text implies. Since this diff already reworks this path, a `try/catch(const std::exception&)` around the `ReadResolvedProfile` call routed to `status.error`, mirroring the `SelectParams` guard, would close it here. Not blocking if you prefer to keep it in the follow-up.

qml/onboarding_settings.cpp
```diff
@@ -491,85 +912,184 @@ bool ApplyToArgs(ArgsManager& args, const DataDirSelection& data_dir_selection,
...
-            if (!args.ReadSettingsFile(&settings_errors)) {
-                if (error) *error = QString::fromStdString(settings_errors.empty() ? std::string{"Settings file could not be read."} : settings_errors.front());
-                return false;
-            }
-            if (!args.WriteSettingsFile(&settings_errors, /*backup=*/true)) {
-                if (error) *error = QString::fromStdString(settings_errors.empty() ? std::string{"Settings file backup could not be written."} : settings_errors.front());
-                return false;
-            }
-        }
+        if (!BackupSettingsFile(args, error)) return false;
```
epicleafies, line 968: Suggestion: when `settings.json` is unreadable and a reset is requested, this backup captures an already-emptied file. `InitConfig` runs first; on an unparseable `settings.json`, `ReadSettingsFile` clears `rw_settings` and fails, the read-error handler returns false to proceed, and `InitConfig` immediately writes an empty document. Only afterward does `BackupSettingsFile` run, so `settings.json.bak` holds the emptied file, not the original bytes. Reproduced: seed a corrupt `settings.json`, launch with `-resetguisettings`, finish onboarding; the `.bak` contains only the auto-generated header. Low severity since the data was unreadable and the user asked to reset, but the backup gives false reassurance in the one case it exists for.

Test gaps around this path: `run_malformed_settings_reset_recovers` does not assert `.bak` contents, and `resetLegacyCleanupRollsBackOnWriteFailure` (the only rollback-failure test for the legacy-cleanup path) is `QSKIP`'d off POSIX, so it is unverified on Windows CI. Consider backing up before `InitConfig`, or documenting the limitation, plus asserting the backup content.

qml/onboarding_settings.cpp
```diff
@@ -491,85 +912,184 @@ bool ApplyToArgs(ArgsManager& args, const DataDirSelection& data_dir_selection,
...
+    const bool bootstrap_gui_settings_changed{
+        bootstrap_settings_needed && bootstrap_gui_settings_values != original_bootstrap_gui_settings
+    };
+    if (active_gui_settings_changed &&
+        !WriteGuiSettings(
+            *active_gui_settings,
+            active_gui_settings_values,
+            QStringLiteral("GUI settings update"),
+            error)) {
+        RestoreGuiSettings(active_gui_settings_store, original_active_gui_settings, error);
```
epicleafies, line 1028: Suggestion: this rollback opens a fresh `QSettings` on the same file while the original `active_gui_settings` owner is still alive, so two `QSettings` back the same file with independent caches. The explicit restore `sync()` should win, but `RestoreStores` in `legacy_settings_migration.cpp` handles the same shape by resetting the owner first. The two rollback paths are inconsistent; resetting or rewriting through the existing object would be less fragile.

qml/onboarding_settings.cpp
```diff
@@ -251,6 +356,305 @@ bool WriteSettingsFile(ArgsManager& args, QString* error)
...
+
+bool ValidatePendingApply(
+    ArgsManager& args,
+    const QmlOnboardingSettings::PendingApply& pending,
+    QString* error)
+{
+    if (pending.effective_reset != args.GetBoolArg("-resetguisettings", false)) {
+        if (error) {
+            *error = QStringLiteral(
+                "The selected Bitcoin profile changed while onboarding was open. "
```
epicleafies, line 603: Nit: these new startup-error strings reach the user via `InitError` but are untranslated `QStringLiteral`s. Wrap the user-visible ones in `tr()` (with `%1` placeholders rather than concatenation).

qml/bitcoin.cpp
```diff
@@ -296,8 +299,35 @@ enum class PreInitOnboardingStatus {
...
+        return false;
+    }
+
+    QMessageBox message_box{
+        QMessageBox::Critical,
+        CLIENT_NAME,
+        QString::fromStdString(strprintf("%s.", error.translated)),
+        QMessageBox::Reset | QMessageBox::Abort,
+    };
+    message_box.setInformativeText(QObject::tr("Do you want to reset settings to default values, or to abort without making changes?"));
```
epicleafies, line 314: Nit: this string lost the translator comment that the Widgets `ErrorSettingsRead` carries. In `bitcoin/src/qt/bitcoin.cpp` the same `setInformativeText` has a `/*: ... */` comment above it explaining the reset-or-abort choice for translators. Since this is the one translated string in the new code, restoring the comment keeps parity and satisfies the translator-comment convention.

epicleafies (commented):
tested ACK 49c2f253

Built on Qt 6.4.2, unit and functional suites green, plus a headed pass under xcb. Confirmed the reset targets the resolved datadir's settings.json (custom datadir kept, default untouched), config-redirected datadirs reach the main window, and -resetguisettings=0 is a no-op.

Two items the green suites do not catch (both headed-only) are inline: #775 (an unreadable saved datadir still aborts at startup, a hard crash rather than a soft failure; details on the issue) and an empty reset backup when settings.json is unparseable. A few smaller issues inline too.

johnny9 (commented):
Two settings concerns:

qml/onboarding_settings.cpp
```diff
@@ -491,85 +912,184 @@ bool ApplyToArgs(ArgsManager& args, const DataDirSelection& data_dir_selection,
...
+{
+    if (error) error->clear();
+    if (result) *result = {};
+
+    if (pending && !ValidatePendingApply(args, *pending, error)) return false;
+
+    const bool reset_gui_settings{args.GetBoolArg("-resetguisettings", false)};
+    const std::map<std::string, common::SettingsValue> original_rw_settings{SnapshotRwSettings(args)};
+
+    const GuiSettingsStore active_gui_settings_store{CurrentGuiSettingsStore()};
```
johnny9, line 937: This opens and checks the active GUI settings file even when there is no onboarding change and no reset. I reproduced an unreadable file causing startup to stop before the main window, even though nothing needs to be changed. Can this store be opened only when it is actually needed? A test with an unreadable active store and no pending apply or reset would catch this.

qml/onboarding_settings.cpp
johnny9, line 968: I think this should be fixed here. The reset succeeds, but `settings.json.bak` should contain the original malformed file, not the empty file written by `InitConfig()`. Since this PR adds the backup sequence, can it stage the original bytes before `InitConfig()` changes the file, keep them as `.bak` only after the reset succeeds, and test the backup contents?

johnny9:
Superseded by 883

PR #814, opened 2026-07-29
Receive page improvements

### LINKED ISSUE #807: Receive usability refinements
There are some tweaks to the receive screen (more specifically when you view a generated request) that I think could improve the usability. Here we go:

1. The "New request" button is primary. However, when you created a new payment request, the primary action should be to share it with someone. It is a bit of a trap to make the button for clearing the form so prominent. I suggest using the outline button style, and moving the button below the buttons that relate to the payment request directly
2. When a payment request was generated, the input fields still show placeholders like "Enter name...", although they are disabled. Replacing those with an em-dash to indicate no content communicates more clearly
3. Should we move the "Edit" functionality into the dropdown? It certainly seems like a secondary action to the act of sharing the payment request
4. In the original designs, we had the QR code displayed, small, next to the form. Clicking it was meant to show it larger. I am curious why it has become a separate button. Definitely an option if we don't want the overhead of rendering a QR code for every render, or if it's an option that is not used that often.
5. I was hoping we could tap into the share sheets that operating systems provide for the "Share" button (wherever available). How someone looked into that on the technical feasibility side?

The image shows a screenshot of the current preview app on the left, and a mock-up with ideas for changes on the right.

Image

pseudoramdom:
Fixes #807 

---

- Add a reusable `AddressLabel` component that:
    - Displays addresses in four-character groups with alternating colors.
    - Highlights on hover. Copies the full address when clicked.
- Use `AddressLabel` in address details and generated payment requests.
- Remove the redundant address Copy control and copied toast.
- Move `Edit payment request` into the overflow menu.
- Move `New Request` below the request actions as an outline button.
- Display an em dash for empty fields after a request is generated.

Screenshot 2026-07-29 at 3 33 08 PM

### DIFF
--- a/test/qml/tst_activity.qml
+++ b/test/qml/tst_activity.qml
@@ -169,7 +169,6 @@ TestCase {
         compare(testWalletModel.lastLoadedPaymentRequestDetailId, "req-1")
         compare(testPaymentRequest.id, "req-1")
         compare(testPaymentRequest.isEditing, false)
-
         const editButton = findChild(page.currentItem, "paymentRequestDetailEdit")
         verify(editButton !== null)
         editButton.clicked()
--- a/qml/bitcoin_qml.qrc
+++ b/qml/bitcoin_qml.qrc
@@ -2,6 +2,7 @@
     <qresource prefix="/qml">
         <file>components/AboutOptions.qml</file>
         <file>components/AddressDetailRow.qml</file>
+        <file>components/AddressLabel.qml</file>
         <file>components/AlertAction.qml</file>
         <file>components/AlertPopup.qml</file>
         <file>components/BitcoinAddressDisplayField.qml</file>
--- a/qml/components/AddressDetailRow.qml
+++ b/qml/components/AddressDetailRow.qml
@@ -4,7 +4,6 @@
 
 import QtQuick 2.15
 import QtQuick.Layouts 1.15
-import org.bitcoincore.qt 1.0
 
 import "../controls"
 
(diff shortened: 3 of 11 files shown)


qml/pages/wallet/PaymentRequestDetail.qml
```diff
@@ -223,48 +222,6 @@ Page {
                     onClicked: root.editRequest()
                 }
 
-                Button {
```
johnny9, line 226: There's no way to copy the payment request details now that this button is gone from the request details history. We should keep this here

johnny9 (commented):
commit 696c347 needs components/AddressLabel.qml registered in bitcoin_qml.qrc and move the AddressLabel test from tst_address_components.qml:87 into that same commit.

qml/pages/wallet/PaymentRequestDetail.qml
johnny9: Screenshot from 2026-08-03 09-07-31

pseudoramdom: Ah I thought this button was just copying the address (for some reason). I'll restore it.

pseudoramdom:
Rebased to 0c58619ba110216ca4b0505d298e747bfb0e6af5 to address the johnny's feedback (https://github.com/bitcoin-core/gui-qml/pull/814#pullrequestreview-4844335147)

test/qml/tst_activity.qml
```diff
@@ -169,6 +169,7 @@ TestCase {
         compare(testWalletModel.lastLoadedPaymentRequestDetailId, "req-1")
         compare(testPaymentRequest.id, "req-1")
         compare(testPaymentRequest.isEditing, false)
+        compare(findChild(page.currentItem, "paymentRequestDetailCopy"), null)
```
johnny9: This assertion contradicts the restored `paymentRequestDetailCopy` button in `PaymentRequestDetail.qml` and makes `bitcoinqml_qmltests` fail deterministically (392 passed, 1 failed) in both Qt CI jobs. Since keeping this Copy action was the resolution of the earlier review thread, please remove this assertion or change it to verify that the button is present.

pseudoramdom: Removed it.

johnny9 (approved):
ACK 4c198d5684432c45a542c867c1c939341347454a

Tested on Fedora. Previous comments addressed. Looks good and tests pass

PR #817, opened 2026-08-01
Activity: advanced filters, edited request labels, and used-address requests

### LINKED ISSUE #707: Activity filters: add advanced date range, amount, and Other filters

### LINKED ISSUE #725: Activity: pending receive request row keeps stale label after editing payment request
## Problem

Editing a saved payment request label does not update the pending request row shown in **Activity**.

A receive request can be created with one label, edited to a new label, and then still appear in Activity with the old label.

## Reproduction

1. Open **Receive**.
2. Set the Name field to `Old Label` and enter an amount.
3. Click **Generate payment request**.
4. Click **Edit**.
5. Change the Name field to `New label`.
6. Click **Update payment request**.
7. Open **Activity**.

## Observed behavior

The pending request row in Activity still shows `Old Label`.

## Expected behavior

The Activity row should show the updated request label (`New label`).

## Likely cause

`ActivityListModel::updateReceiveRequest()` updates the pending request row, but the Activity label path appears to overwrite that value by re-reading the wallet address-book label for the address. That means an edited payment-request label can get replaced by the older address label when Activity renders.

## Code references

- `qml/models/activitylistmodel.cpp`
  - `updateReceiveRequest()`
  - `updateTransactionLabel()`
  - `data(..., LabelRole)`

## Original review comment

- https://github.com/bitcoin-core/gui-qml/pull/708#discussion_r3295381807

## Parent epic

- #581

### LINKED ISSUE #726: Used-address payment requests are not shown in Activity under the Payment request filter
## Summary

On current `origin/qt6` (fetched locally to `f9e22ad41c`), payment requests for **already used addresses** are visible in **Activity details**, but they are **not surfaced as Activity rows** when the **Payment request** filter is selected.

This leaves the implementation only halfway to the behavior described in the PR review discussion.

Surfaced by review comment:
https://github.com/bitcoin-core/gui-qml/pull/708#discussion_r3295457321

## Current behavior

- Unused-address payment requests are added to the Activity list as pending request rows.
- Used-address payment requests are not added as Activity rows at all.
- For a real transaction row, Activity details already show matching payment requests for that address.

## Expected behavior

If the intended policy is that payment requests for already used addresses should only appear in the Activity list when the **Payment request** filter is selected, then those requests still need to be materialized in the Activity model in a way that the filter can expose them.

## Likely gap in the current code

`ActivityListModel::addPendingReceiveRequests()` skips any receive request whose address already exists in activity:

```cpp
if (address.isEmpty() || existing_addresses.contains(address)) continue;
```

So used-address requests never become rows in `ActivityListModel`.

`ActivityFilterProxyModel` only classifies/filter rows that already exist. In particular, it maps rows with `IsPendingRequestRole` to `PaymentRequest`, but it does not synthesize rows for used-address requests.

## Relevant code paths

- `qml/models/activitylistmodel.cpp`
  - `addPendingReceiveRequests()`
  - `transactionDetails()`
- `qml/models/activityfilterproxymodel.cpp`
  - `filterTypeForIndex()`
- `qml/pages/wallet/ActivityDetails.qml`

## Notes

The details-side support already exists via `matchingEntriesForAddress(...)`, so this seems specifically like a model/list-visibility gap rather than a missing details implementation.

epicleafies:
This PR closes three related gaps in the Activity page. The list filters now reach parity with the Qt Widgets transaction view and payment request labels and used-address requests behave correctly. Fixes #707. Fixes #725. Fixes #726.

 - Added an Other transaction type, a minimum amount and a custom date range as Activity filter options
 - Added a Grid-based calendar component for the custom date range
 - Updated the activity page search and filter options layout and kept filter popups within the list area 
 - Moved the Payment request type filter above Mined and Other
 - Made payment request labels stay in sync both ways. A request edit updates its Activity row and the address book, and an address book edit updates the matching requests and any views already open on them
 - Fixed used-address payment requests. They appear only under the Payment request filter and keep the request presentation (purple icon and creation date, since address use alone does not prove payment) and a payment marks every request for the paid address used so the live list matches a reload
  - Reset the Activity detail stack when opening a transaction from the send flow, so View Transaction shows the new transaction and Back returns to the list instead of a page left open earlier
 - Applied wallet transaction updates to the Activity model on the model's thread. Updating from the notification thread could transiently show an incoming payment twice and hide the matching request row
 - Aligned the Activity title and top padding with the Send and Receive pages

Added unit, QML and functional tests.

image

image

image

### DIFF
--- a/qml/controls/ActivityCalendar.qml
+++ b/qml/controls/ActivityCalendar.qml
@@ -1,0 +1,551 @@
+// Copyright (c) 2026 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import QtQuick 2.15
+import QtQuick.Controls 2.15
+import QtQuick.Layouts 1.15
+import org.bitcoincore.qt 1.0
+
+// Date-range calendar for the Activity custom date range filter: From/To
+// display, month/year navigation, a weekday header, and a day grid with range
+// selection, hover preview, and keyboard support. It owns the selection state;
+// the caller reads startDate/endDate (or startIso/endIso) and valid/hasSelection
+// and drives it with seed()/reset()/refreshToday(). It does not touch the model:
+// applying or clearing the filter is left to the caller.
+ColumnLayout {
+    id: root
+    spacing: 8
+
+    // Public interface ------------------------------------------------------
+
+    // The picked range as local Date values (null until picked).
+    readonly property var startDate: root.customFrom
+    readonly property var endDate: root.customTo
+    // The picked range as yyyy-MM-dd strings ("" until picked). Pass these to
+    // the model rather than the Dates: a JS Date converted straight to QDate can
+    // shift a day across time zones.
+    readonly property string startIso: root.customFrom ? root.isoDate(root.customFrom) : ""
+    readonly property string endIso: root.customTo ? root.isoDate(root.customTo) : ""
+    // Both ends picked (a complete range), and any end picked.
+    readonly property bool valid: root.customFrom !== null && root.customTo !== null
+    readonly property bool hasSelection: root.customFrom !== null || root.customTo !== null
+
+    // Clear the picked range so a new one can be chosen.
+    function reset() {
+        root.customFrom = null
+        root.customTo = null
+        root.pickingEnd = false
+        root.hoverDate = null
+    }
+
+    // Refresh "today" (call when the calendar is shown) so a page left up past
+    // midnight does not keep marking yesterday as today, and drop the focus ring.
+    function refreshToday() {
+        root.todayDate = new Date()
+        root.calendarFocusDate = null
+    }
+
+    // Seed the calendar with an existing range (or nulls to start clean) and
+    // show the month of its start.
+    function seed(from, to) {
+        root.customFrom = from
+        root.customTo = to
+        root.pickingEnd = false
+        root.hoverDate = null
+        root.showCalendarMonth(from)
+    }
+
+    // Internal state --------------------------------------------------------
+
+    // Dates are shown in the same long shape as the rest of the app (see
+    // Transaction::dateTimeString), but rendered with toLocaleDateString so
+    // month names localize (a Qt.formatDate format string renders C-locale
+    // English in Qt 6), matching the localized month header and weekday row.
+    readonly property string activityDateFormat: "MMMM d, yyyy"
+
+    // customFrom and customTo are local Date values (or null). While pickingEnd
+    // is true the start is chosen and the next click sets the end; hoverDate
+    // previews the range on desktop. The calendar shows calMonth/calYear.
+    property var customFrom: null
+    property var customTo: null
+    property bool pickingEnd: false
+    property var hoverDate: null
+    property int calMonth: (new Date()).getMonth()
+    property int calYear: (new Date()).getFullYear()
+    // First day of the week for the locale, as Qt::DayOfWeek
+    // (1 = Monday .. 7 = Sunday). The mod-7 arithmetic in calendarCells() and
+    // the weekday header maps it onto JS getDay()'s 0 = Sunday .. 6 = Saturday
+    // numbering (7 % 7 == 0).
+    readonly property int weekStart: Qt.locale().firstDayOfWeek
+    property var todayDate: new Date()
+    // Day the keyboard focus ring is on while the day grid has active focus;
+    // null until the grid is first focused after opening.
+    property var calendarFocusDate: null
+
+    // The chips are too narrow for the long Activity date format (a month
+    // like September wraps), so they use the locale's short date form.
+    function formatRangeDate(date) {
+        if (!date || isNaN(date.getTime())) {
+            return ""
+        }
+        return date.toLocaleDateString(Qt.locale(), Locale.ShortFormat)
+    }
+
+    function isoDate(date) {
+        return Qt.formatDate(date, "yyyy-MM-dd")
+    }
+
+    function sameDay(a, b) {
+        return a && b && root.isoDate(a) === root.isoDate(b)
+    }
+
+    // The two ends of the range currently in play: the committed range, or the
+    // tentative one while hovering after the first click.
+    function rangeBounds() {
+        if (root.customFrom && root.customTo) {
+            return [root.customFrom, root.customTo]
+        }
+        if (root.pickingEnd && root.customFrom && root.hoverDate) {
+            return root.hoverDate < root.customFrom
+                ? [root.hoverDate, root.customFrom]
+                : [root.customFrom, root.hoverDate]
+        }
+        return null
+    }
+
+    function isRangeEndpoint(date) {
+        // Check the picked ends directly so the start lights up on the click,
+        // not only once the mouse moves and a hover range exists.
+        return root.sameDay(date, root.customFrom)
+            || root.sameDay(date, root.customTo)
+            || (root.pickingEnd && root.customFrom && root.hoverDate
+                && root.sameDay(date, root.hoverDate))
+    }
+
+    function isInRange(date) {
+        const b = root.rangeBounds()
+        return b !== null && date >= b[0] && date <= b[1]
+    }
+
+    // Cells for the displayed month: leading blanks (null) so the first day
+    // lands under its weekday for the locale, then one Date per day.
+    function calendarCells() {
+        var cells = []
+        var firstDow = new Date(root.calYear, root.calMonth, 1).getDay()
+        var lead = (firstDow - root.weekStart + 7) % 7
+        for (var i = 0; i < lead; ++i) {
+            cells.push(null)
+        }
+        var days = new Date(root.calYear, root.calMonth + 1, 0).getDate()
+        for (var d = 1; d <= days; ++d) {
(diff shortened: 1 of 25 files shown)


johnny9:
Can you go ahead and make the title and top padding align with Send/Receive?

qml/models/activitylistmodel.cpp
```diff
@@ -329,25 +347,18 @@ void ActivityListModel::fulfillPendingRequest(int index, const QSharedPointer<Tr
...
-    if (!real_tx->label.isEmpty()) {
-        pending->label = real_tx->label;
-    }
-
+    // The request's address now has a real transaction. Rather than consuming
+    // the request in place, keep it as a used-address request (surfaced only
+    // under the Payment request filter) and insert the transaction as its own
+    // row, so the live list matches the state rebuilt on reload.
+    pending->isUsedAddressRequest = true;
     m_pending_request_addresses.remove(pending->address);
```
johnny9, line 355: Could this mark every pending request for the address as used before removing it from `m_pending_request_addresses`?

Bitcoin Core explicitly supports multiple receive requests for one address. With two pending rows, `findPendingRequestIndex()` returns one, this function sets `isUsedAddressRequest` only on that row, and then removes the address from pending tracking. The other row therefore stays purple/pending and remains visible in the default Activity list for the rest of the session. After a reload, `addPendingReceiveRequests()` sees the real transaction and marks both rows as used, so the live and reloaded states disagree.

I reproduced this by extending `paidPendingRequestBecomesUsedAddressRequestLive()` with two requests for the same address; the second request failed the `IsUsedAddressRequestRole` assertion.

Suggested fix: capture the address, iterate `m_transactions`, set `isUsedAddressRequest` on every pending request with that address, emit `dataChanged` for each affected row (including `IsUsedAddressRequestRole` and `DateTimeRole` if roles are specified), then remove the address from `m_pending_request_addresses` once. Please add a two-request regression test that asserts both requests are used and hidden by the default proxy after the payment arrives. This also addresses the duplicate-request case raised in the [original PR #708 review](https://github.com/bitcoin-core/gui-qml/pull/708#discussion_r3295457321).

qml/pages/wallet/Activity.qml
```diff
@@ -108,6 +108,8 @@ PageStack {
                     return qsTr("This week")
                 case ActivityFilterProxyModel.ThisMonth:
                     return qsTr("This month")
+                case ActivityFilterProxyModel.LastMonth:
+                    return qsTr("Last month")
```
pseudoramdom: In `qml: add Other type and Last month filters to Activity` -

---
I'm unsure if we really need `Last month` and not clear how it differs from `This month` when I see it in the menu.
IIUC, these are adjacent, non-overlapping ranges. So on August 26:
  - This month: August 1–31
  - Last month: July 1–31

Searching for something in July when you're in Aug doesn't deserve a top-level spot IMO. We can drop this since we have the custom range picker.

pseudoramdom: Also it's easier to read when all the options have the `This` prefix. Throwing a `Last X` in between is odd.

qml/controls/ActivityCalendar.qml
```diff
@@ -0,0 +1,532 @@
...
+            root.pickingEnd = false
+        }
+        root.hoverDate = null
+    }
+
+    // Accessible label for a calendar day: the full localized date plus its
+    // selection state, so the state is announced to a screen reader and not
+    // conveyed by color alone.
+    function dayCellAccessibleName(cell) {
+        var base = Qt.formatDate(cell.modelData, root.activityDateFormat)
```
pseudoramdom: In qml: add minimum amount and custom date range Activity filters

Codex notes that the range-chips & day cells use a different locale than what we use in Month & days heading.

qml/models/activityfilterproxymodel.h
```diff
@@ -18,6 +21,9 @@ class ActivityFilterProxyModel : public QSortFilterProxyModel
     Q_PROPERTY(DateFilter dateFilter READ dateFilter WRITE setDateFilter NOTIFY dateFilterChanged)
     Q_PROPERTY(TypeFilter typeFilter READ typeFilter WRITE setTypeFilter NOTIFY typeFilterChanged)
     Q_PROPERTY(int displayUnit READ displayUnit WRITE setDisplayUnit NOTIFY displayUnitChanged)
+    Q_PROPERTY(qint64 minAmount READ minAmount WRITE setMinAmount NOTIFY minAmountChanged)
+    Q_PROPERTY(QDate rangeStart READ rangeStart WRITE setRangeStart NOTIFY rangeChanged)
```
pseudoramdom: These could be made read-only

qml/models/activityfilterproxymodel.cpp
```diff
@@ -155,6 +155,80 @@ void ActivityFilterProxyModel::setDisplayUnit(int display_unit)
...
+    invalidateFilter();
+#endif
+    Q_EMIT rangeChanged();
+    Q_EMIT countChanged();
+}
+
+void ActivityFilterProxyModel::setCustomRange(const QString& start_iso, const QString& end_iso)
+{
+    setRangeStart(QDate::fromString(start_iso, Qt::ISODate));
+    setRangeEnd(QDate::fromString(end_iso, Qt::ISODate));
```
pseudoramdom: Each of the two calls will call
```qml
Q_EMIT rangeChanged();
Q_EMIT countChanged();
```
Could we do it one go? Should we also do some validations like
```qml
bool ActivityFilterProxyModel::applyCustomRange(
      const QString& start_iso,
      const QString& end_iso)
{
  const QDate start = QDate::fromString(start_iso, Qt::ISODate);
  const QDate end = QDate::fromString(end_iso, Qt::ISODate);

  if (!start.isValid() || !end.isValid() || start > end) {
    return false;
  }

  const bool range_changed = start != m_range_start || end != m_range_end;
  const bool filter_changed = m_date_filter != CustomRange;

  if (!range_changed && !filter_changed) return true;

//.. beginFilterChange
// set range
// endFilterChange
 

  if (range_changed) Q_EMIT rangeChanged();
  if (filter_changed) Q_EMIT dateFilterChanged();
      
  Q_EMIT countChanged();
  return true;
}
```

pseudoramdom (commented):
Partial review till `qml: add minimum amount and custom date range Activity filters` 7daae41acce2093cbb22037acd894e6fdd48f290
Will complete the rest tomorrow.

johnny9 (approved):
ACK 5e89ae25c7dc8ac8a2f7f960d8816894530f0d3e

Some minor issues for follow up:

- Label updates: Editing only a request’s amount or message should preserve the label. Update the label only when the request label changes.
- Keyboard access: Make the Presets and calendar navigation buttons reachable with Tab and usable with Enter or Space. We'll need to do accessibility on all of our views at some point so can defer until then.
- Date boundaries: Use startOfDay() so custom date filters work when daylight saving skips midnight on Qt 6.2.

hebasto:
Please rebase.

johnny9 (approved):
ACK e473174ea8a116740b1ce8a7b8be8897777c7837

Reviewed the rebase and both fixes. Keyboard accessibility can follow separately.

hebasto:
https://github.com/bitcoin-core/gui-qml/actions/runs/34518133156/job/103008632780?pr=817:
```
 FAIL!  : bitcoinqml_qmltests::SettingsNavigation::test_walletStackResetsWhenSelectedWalletChangesWhileHidden() Compared values are not the same
   Actual   (): false
   Expected (): true
   Loc: [qrc:/test/qml/tst_settingsnavigation.qml(378)]
```

pseudoramdom:
> https://github.com/bitcoin-core/gui-qml/actions/runs/34518133156/job/103008632780?pr=817:
> 
> ```
>  FAIL!  : bitcoinqml_qmltests::SettingsNavigation::test_walletStackResetsWhenSelectedWalletChangesWhileHidden() Compared values are not the same
>    Actual   (): false
>    Expected (): true
>    Loc: [qrc:/test/qml/tst_settingsnavigation.qml(378)]
> ```

I think #872 broke the `qt6` branch. https://github.com/bitcoin-core/gui-qml/pull/876 has the fix for it (and it's ready to merge if ACK'ed).

epicleafies:
Will rebase again once fix is in.

PR #860, opened 2026-08-14
Improve Send and Send Review pages

### LINKED ISSUE #830: “View transaction” button after sending takes you to Send tab
I sent a transaction and clicked the “View transaction” button:
Image

However, it took me to the “Send bitcoin” tab rather than a transaction details page:

Image

### LINKED ISSUE #831: "View Transaction" closes the window
"View Transaction’ closes the window instead of displaying the transaction.

Image

### LINKED ISSUE #836: bug: Text entry issues in sending address input
In the send page in the address field

when entering an address if you type in a symbolic character e.g. `.[]'"\|()**&%^!@#$` etc. the cursor stays fixed and the characters go to the right, see imager: 

Image

and you can keep typing in symbols and they get appended to the right

Image

and then once you type in an alphanumeric, even if it's not in the character set, e.g. the letter `o` it deletes all those characters:

Image

Orthogonally, maybe characters that are not permitted in the data section of bech32 should be filtered?

### LINKED ISSUE #839: Pasting bitcoin:<address> into recipient field doesn’t drop prefix
I tried pasting `bitcoin:tb1q9255gm5sug4jp9etgfg9l08qepex9xvtqm0ql4` into the recipient field, and it dropped the colon, but tried to make an address from the remaining characters:

Image

It would be neat if it automatically dropped the `bitcoin:` prefix.

### LINKED ISSUE #844: "Review Transaction"  shows different units
Image

The amount for the first recipient is shown in Bitcoin while the second in sats.

### LINKED ISSUE #873: Right clicking on a textfield pulls up the context menu in dark mode colors
This seems to be part of Qt controls default offering. We should override the colors. 

Image

pseudoramdom:
This is an omnibus PR that fixes a several issues related to Send and Send review pages.

- Allow payment requests to be pasted directly into the address field.
  - Fixes #839 
- Reject invalid characters in the address field.
  - Fixes #836 
- Navigate to the activity details page when selecting “View Transaction” after sending.
  - Fixes #830. 
  - Fixes #831.
- Keep the bitcoin/sats unit preference synchronized on the Send Review page.
  - Fixes #844
 - Fixes #873  
- Widen and animate the wallet passphrase popup.
- Allow addresses to toggle between truncated and full forms on the Send Review page.
- Improve the multiple-recipient Send Review layout.

**Review note:** The changes are organized into focused commits, so this PR is best reviewed commit by commit.

### DIFF
--- a/qml/components/BitcoinAddressInputField.qml
+++ b/qml/components/BitcoinAddressInputField.qml
@@ -17,9 +17,35 @@ ColumnLayout {
     property string labelText: qsTr("Send to")
     property bool enabled: true
     property alias inputObjectName: addressInput.objectName
+    property bool interceptPaste: false
 
     signal textChanged()
     signal editingFinished()
+    signal pasteRequested()
+
+    function paymentUri(text) {
+        const trimmedText = String(text).trim()
+        return root.interceptPaste && trimmedText.toLowerCase().startsWith("bitcoin:")
+            ? trimmedText
+            : ""
+    }
+
+    function pastedPaymentUri(previousText, currentText) {
+        const clipboardText = Clipboard.text()
+        const uri = paymentUri(clipboardText)
+        if (uri.length === 0 || previousText === currentText) return ""
+        return String(currentText).indexOf(clipboardText) === -1 ? "" : uri
+    }
+
+    function paste() {
+        addressInput.paste()
+    }
+
+    function syncFromAddress() {
+        addressInput.syncFromAddress()
+    }
+
+    onAddressChanged: syncFromAddress()
 
     Layout.fillWidth: true
     spacing: 4
@@ -43,13 +69,34 @@ ColumnLayout {
 
         TextArea {
             id: addressInput
+            property bool syncingFromAddress: false
+
+            function syncFromAddress() {
+                const formattedAddress = root.address ? root.address.formattedAddress : ""
+                if (text === formattedAddress) return
+                syncingFromAddress = true
+                text = formattedAddress
+                syncingFromAddress = false
+            }
+
+            function handlePaymentUriInput() {
+                const previousText = root.address ? root.address.formattedAddress : ""
+                const uri = root.pastedPaymentUri(previousText, text)
+                if (uri.length === 0) return false
+                syncFromAddress()
+                root.pasteRequested()
+                syncFromAddress()
+                return true
+            }
+
             objectName: root.inputObjectName
             anchors.left: label.right
             anchors.right: parent.right
             anchors.verticalCenter: parent.verticalCenter
             enabled: root.enabled
             placeholderText: qsTr("Enter address...")
-            text: root.address ? root.address.formattedAddress : ""
+            text: ""
+            Component.onCompleted: syncFromAddress()
             wrapMode: Text.WrapAnywhere
             leftPadding: 0
             topPadding: 0
@@ -62,7 +109,19 @@ ColumnLayout {
             background: Item {}
             selectByMouse: true
 
+            Keys.onPressed: (event) => {
+                if (root.interceptPaste && event.matches(StandardKey.Paste)) {
+                    root.pasteRequested()
+                    event.accepted = true
+                }
+            }
+
             onTextChanged: {
+                if (syncingFromAddress) {
+                    root.textChanged()
+                    return
+                }
+                if (handlePaymentUriInput()) return
                 if (root.address) {
                     cursorPosition = root.address.setAddress(text, cursorPosition)
                 }
@@ -79,6 +138,13 @@ ColumnLayout {
             onActiveFocusChanged: {
                 if (!activeFocus && root.address) {
                     root.address.setAddress(text)
+                }
+            }
+
+            Connections {
+                target: root.address ? root.address : null
+                function onFormattedAddressChanged() {
+                    addressInput.syncFromAddress()
                 }
             }
         }
--- a/qml/components/BitcoinAmountInputField.qml
+++ b/qml/components/BitcoinAmountInputField.qml
@@ -22,13 +22,45 @@ ColumnLayout {
     property alias errorTextObjectName: errorTextLabel.objectName
     property bool enabled: true
     property string placeholderText: root.amount ? root.amountInputPlaceholder(root.amount.unit) : "0.00000000"
+    property bool interceptPaste: false
 
     signal inputTextChanged
     signal textEdited
     signal editingFinished(string value)
+    signal pasteRequested()
 
     function syncFromAmount(force) {
         amountInput.syncFromAmount(force)
+    }
+
+    function paymentUri(text) {
+        const trimmedText = String(text).trim()
+        return root.interceptPaste && trimmedText.toLowerCase().startsWith("bitcoin:")
+            ? trimmedText
+            : ""
+    }
+
+    function pastedPaymentUri(previousText, currentText) {
+        const clipboardText = Clipboard.text()
+        const uri = paymentUri(clipboardText)
+        if (uri.length === 0 || previousText === currentText) return ""
+
+        let offset = String(currentText).indexOf(clipboardText)
+        while (offset !== -1) {
+            const prefix = String(currentText).slice(0, offset)
+            const suffix = String(currentText).slice(offset + clipboardText.length)
+            if (String(previousText).startsWith(prefix)
+                    && String(previousText).endsWith(suffix)
+                    && prefix.length + suffix.length <= String(previousText).length) {
+                return uri
+            }
+            offset = String(currentText).indexOf(clipboardText, offset + 1)
+        }
+        return ""
+    }
+
+    function paste() {
+        amountInput.paste()
     }
 
     function amountInputPlaceholder(unit) {
@@ -79,6 +111,7 @@ ColumnLayout {
         TextField {
             id: amountInput
             property bool syncingFromAmount: false
+            property string lastAcceptedText: ""
 
             function amountDisplay() {
                 return root.amount ? root.amount.display : ""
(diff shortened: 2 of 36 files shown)


test/test_bitcoinaddress.cpp
```diff
@@ -11,10 +11,49 @@ class BitcoinAddressTests : public QObject
...
+
+    QCOMPARE(address.address(), QStringLiteral("bc1qexampleaddress"));
+    QCOMPARE(address.formattedAddress(), QStringLiteral("bc1q exam plea ddre ss"));
+    QCOMPARE(cursor_position, static_cast<int>(address.formattedAddress().size()));
+    QCOMPARE(formatted_address_changed.count(), 1);
+}
+
+void BitcoinAddressTests::setAddress_stripsBitcoinUriPrefix()
+{
+    const QString uri{QStringLiteral("bitcoin:bc1qexampleaddress")};
```
johnny9: Should we go ahead and add Uri parsing for all input fields? 

The uri can contain data for the amount and label as well. Would probably need to be handled at the form level instead of just the input fields so that the rest of the fields could be modified when pasting. We'd also probably need to confirm overwrite if there is already information in the unfocused fields.

johnny9 (commented):
Everything looks good in this PR. Just want to check if we should go ahead and handle full bitcoin uri's with this PR or not.

test/test_bitcoinaddress.cpp
pseudoramdom: This is a good suggestion. The commit `qml: Parse payment URIs pasted into send fields` (8cf4f065d7f8e9a6917b3bc42087fa73d5290250) does this now.

pseudoramdom: > We'd also probably need to confirm overwrite if there is already information in the unfocused fields

Screenshot 2026-08-17 at 9 16 58 PM

qml/components/BitcoinAddressInputField.qml
```diff
@@ -62,6 +68,14 @@ ColumnLayout {
             background: Item {}
             selectByMouse: true
 
+            Keys.onPressed: (event) => {
+                const pasteModifier = event.modifiers & (Qt.ControlModifier | Qt.MetaModifier)
+                if (root.interceptPaste && event.key === Qt.Key_V && pasteModifier) {
```
johnny9: I think this might be making too big of an assumption about how pasting form the clipboard is configured. I don't know that we can assume ctrl+V is always the paste combo

pseudoramdom: I removed the shortcut-specific paste interception. Bitcoin URIs are now detected when the field content changes and handled at the Send form level, so this works with platform-specific keyboard shortcuts, context-menu paste, and other paste mechanisms.

qml/components/BitcoinAddressInputField.qml
```diff
@@ -63,12 +98,19 @@ ColumnLayout {
...
+                    return
+                }
+                if (handlePaymentUriInput()) return
                 if (root.address) {
                     cursorPosition = root.address.setAddress(text, cursorPosition)
                 }
                 root.textChanged()
             }
 
+            onTextEdited: handlePaymentUriInput()
```
johnny9: `TextArea` has no `textEdited` signal in supported Qt 6.4 and 6.8. CI fails here and the Send page cannot load. `onTextChanged` already calls `handlePaymentUriInput()`, so this line can be removed.

qml/controls/LabeledTextInput.qml
```diff
@@ -49,7 +69,8 @@ Item {
         placeholderTextColor: enabled ? Theme.color.neutral7 : Theme.color.neutral4
         background: Item {}
         selectByMouse: true
-        onTextEdited: root.textEdited()
+        onTextChanged: handlePaymentUriInput()
```
johnny9: `onTextChanged` also runs for model updates. If a label contains a valid Bitcoin URI, it gets parsed again and can set an amount that was not in the original request. I reproduced an encoded label setting the amount to 1 BTC. Please only check user edits or add a sync guard.

qml/components/BitcoinAmountInputField.qml
```diff
@@ -127,12 +145,14 @@ ColumnLayout {
             Component.onCompleted: root.syncFromAmount(true)
 
             onTextChanged: {
+                if (handlePaymentUriInput()) return
```
johnny9, line 198: This only works when the URI replaces the whole field. The amount field starts with `0.00000000`, so a normal paste at the cursor is rejected by the validator. The test hides this by calling `selectAll()` first.

Please compare the old and new text to find what was inserted, verify it matches the clipboard, and parse that value. `StandardKey.Paste` can handle platform keyboard shortcuts, with the same handler used by the context menu. This should also reduce the need for `syncFromAddress()`.

qml/components/BitcoinAddressInputField.qml
```diff
@@ -17,9 +17,24 @@ ColumnLayout {
     property string labelText: qsTr("Send to")
     property bool enabled: true
     property alias inputObjectName: addressInput.objectName
+    property bool interceptPaymentUri: false
 
     signal textChanged()
     signal editingFinished()
+    signal paymentUriEntered(string uri)
+
+    function paymentUri(text) {
```
johnny9, line 26: Would it be simpler to keep the previous explicit paste handler and use `event.matches(StandardKey.Paste)` instead of checking Ctrl/Meta+V? Qt would handle the platform shortcut mappings, the handler would run before validation, and model updates would not be treated as pastes. A context-menu Paste action could call the same shared handler. This may remove the need for the new text synchronization code.

johnny9 (changes requested):
Paste handling still breaks older supported Qt versions and mixes user paste with model updates.

qml/components/BitcoinAddressInputField.qml
pseudoramdom: Fixed. I removed TextArea.onTextEdited.  Now handle native context-menu paste through the guarded onTextChanged path, which should be compatible with the Qt versions

qml/components/BitcoinAmountInputField.qml
pseudoramdom: Fixed in Native context-menu paste now compares the previous and current text, verifies that the inserted substring matches the clipboard.
I also removed the `selectAll()` calls from the functional test so it now covers pasting directly at the cursor

qml/controls/LabeledTextInput.qml
pseudoramdom: This is crazy and a great catch. Should be fixed now.

qml/components/BitcoinAddressInputField.qml
pseudoramdom: Took the suggestion. Keyboard paste now uses `event.matches(StandardKey.Paste)`, allowing Qt to handle the platform shortcut mapping. Keyboard and native context-menu URI pastes are both routed through the shared Send form handler

test/test_walletqmlmodel.cpp
```diff
@@ -1560,6 +1588,42 @@ void WalletQmlModelTests::removeReceiveRequestRemovesPendingActivityRow()
...
 
+void WalletQmlModelTests::activityDetailsSelectLowestOutputIndex()
+{
+    NiceMock<MockWallet>* wallet{nullptr};
+    auto model = MakeWalletModel(wallet);
+    const interfaces::WalletTx wallet_tx = MakeOutgoingActivityWalletTx(
+        {10'000, 20'000, 30'000},
+        {true, false, false},
+        {true, false, false});
+    ON_CALL(*wallet, getWalletTxs()).WillByDefault(Return(std::set<interfaces::WalletTx>{wallet_tx}));
```
johnny9: These need to be updated now to remove gTest

test/qml/tst_send.qml
```diff
@@ -536,4 +541,126 @@ TestCase {
...
+        compare(sendPage.paymentRequestMessage, "Pay me")
+        compare(sendPage.paymentRequestStatus, "Payment request imported from clipboard")
+        compare(sendPage.paymentRequestIsError, false)
+    }
+
+    function test_send_payment_uri_paste_confirms_overwriting_other_fields() {
+        testSendRecipient.amount.satoshi = 100000000
+        testSendRecipient.label = "existing label"
+
+        const page = createTemporaryObject(sendComponent, testWindow.contentItem)
```
johnny9:
```suggestion
        const page = createTemporaryObject(sendComponent, testCase.Window.window.contentItem)
```

johnny9: this is to use the window that qt quick test already provides

test/functional/qml_test_send_receive.py
```diff
@@ -376,7 +377,12 @@ def run_test(*, save_screenshots=False, screenshot_root=None):
...
 
         gui.wait_for_page("sendResultPopup", timeout_ms=10000)
-        gui.click("sendResultDoneButton")
+        gui.click("sendResultViewTransactionButton")
+        gui.wait_for_page("activityDetailsPage", timeout_ms=10000)
+        assert gui.get_property("activityDetailsPage", "txid") == txid
+        assert gui.get_property("activityDetailsPage", "outputIndex") == receiver_output_index
+        checkpoints.checkpoint("view transaction opens the sent activity", gui)
+        gui.click("activityDetailsBackButton")
         gui.wait_for_property("activityTabButton", "visible", True, timeout_ms=10000)
```
johnny9:
```suggestion
        gui.wait_for_property("activityStack", "depth", 1, timeout_ms=10000)
```

this should be less flakey

johnny9 (commented):
Rebase and make the changes to the testcases to match the current pattern

johnny9 (approved):
ACK f121078eff786c7757a87d58f07bf14a23540ae8

PR #864, opened 2026-08-17
qml: Fix onboarding data directory failures

### LINKED ISSUE #775: Unreadable custom datadir can cause and exception

### LINKED ISSUE #783: The app fails to start on Ubuntu 22.04
```console
hebasto@jammy:~/dev/bitcoin-qml$ QT_QPA_PLATFORM=wayland ./build/bin/bitcoin-core-app
QQmlApplicationEngine failed to load component
qrc:/qml/pages/preinit.qml:25:5: Type OnboardingWizard unavailable
qrc:/qml/pages/onboarding/OnboardingWizard.qml:42:9: Type OnboardingStorageLocation unavailable
qrc:/qml/pages/onboarding/OnboardingStorageLocation.qml:36:17: Type StorageLocations unavailable
qrc:/qml/components/StorageLocations.qml:82:5: FolderDialog is not a type
```

### LINKED ISSUE #846: bitcoin-core-app: Assert hit when using a relative datadir path
to reproduce, run bitcoin-core-app with a datadir that has a relative path or a path that includes `~`, e.g.:

```console
$ ./build/bin/bitcoin-core-app -datadir=~/altdatadir
bitcoin-core-app: ./util/fs.cpp:34: fs::path fsbridge::AbsPathJoin(const fs::path&, const fs::path&): Assertion `base.is_absolute()' failed.
Aborted                    (core dumped) ./build/bin/bitcoin-core-app -datadir=~/altdatadir
```

johnny9:
Depends on #812.

This branch is based on the current head of #812. Until #812 merges, this pull request also includes that commit.

This fixes three onboarding failures:

- Show the data directory chooser when a saved data directory cannot be accessed. Config read errors now return a startup error instead of aborting.
- Use a Qt 6.2-compatible folder dialog while keeping the Qt 6.5 and newer implementation.
- Validate explicit data directories before onboarding. Resolve valid relative paths from the current directory and report invalid paths through normal startup error handling.

Tests:

- Added unit tests for unreadable saved directories, config read errors, invalid explicit directories, and valid relative directories.
- Added a QML contract test for both folder dialog implementations.
- Focused OptionsModel tests pass: 6 passed.
- Full QML test target passes.
- Application and test targets build.
- Memory-safety policy scan reports no new findings in changed lines.

Fixes #775.
Fixes #783.
Fixes #846.

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -44,7 +44,7 @@ jobs:
             build-essential ccache cmake pkgconf \
             libboost-dev libsqlite3-dev libgl-dev libqrencode-dev \
             qt6-base-dev qt6-tools-dev qt6-l10n-tools qt6-tools-dev-tools \
-            qt6-declarative-dev qml6-module-qt-labs-settings qml6-module-qtquick qml6-module-qtquick-controls qml6-module-qtquick-dialogs qml6-module-qtquick-layouts qml6-module-qtquick-templates qml6-module-qtqml
+            qt6-declarative-dev qml6-module-qt-labs-platform qml6-module-qt-labs-settings qml6-module-qtquick qml6-module-qtquick-controls qml6-module-qtquick-dialogs qml6-module-qtquick-layouts qml6-module-qtquick-templates qml6-module-qtqml
           echo "CCACHE_DIR=${{ runner.temp }}/ccache" >> "$GITHUB_ENV"
 
       - name: Restore Ccache cache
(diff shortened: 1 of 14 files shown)


johnny9:
Superseded by #883

Issue #858, opened 2026-08-12
FileDialog can crash the app on Qt 6.4 when no platform theme is available


epicleafies:
When Qt 6 runs without a platform theme (for example QT_QPA_PLATFORMTHEME=qt5ct, a Qt 5 only plugin that leaves Qt 6 themeless), QtQuick.Dialogs FileDialog falls back to its non-native QML implementation. On Qt 6.4 that implementation null-derefs in QQuickFileDialogImpl::selectedFile() and crashes the app the moment the dialog is used. Reproduced from the Activity CSV export click, but the defect is app-wide: any FileDialog call site is affected in such an environment.

The application code is not at fault (with a native dialog available everything works), but the failure mode is a hard crash of a routine action with no visible cause, on a plausible minimal-Linux setup. The same themeless fallback also degrades save flows: the non-native dialog does not honour fileMode: FileDialog.SaveFile, so its accept button behaves as "Open" and a new file name cannot be committed.

Suggested fix: prefer the native dialog where one exists, and guard or avoid the non-native Qt 6.4 path so a themeless environment degrades gracefully instead of crashing. CI cannot catch this because the functional tests write the export path through an automation field and never open the dialog.

PR #875, opened 2026-08-25
Activity list and payment request correctness fixes

### LINKED ISSUE #847: Users can edit the expected payment amount after creating it
You can edit the expected payment amount after creating it.

This field isn’t enforced anywhere - not by the sender, the receiver, or the blockchain - so there’s no harm in editing it.
However, it does seem brittle that a user can send one value and then, for example, accidentally edit the wrong payment request.

Image

Image

### LINKED ISSUE #848: Transactions are not sorted in Activity list
Image

During a testing session, I received these transactions to the same address. I navigated to the Activity page after an hour or so after receiving, it showed them out of order and recent-ish time.

Here's the same list that I saw after closing and reopening the app

Image

### LINKED ISSUE #849: design: distinguish zero confirmation transactions from transactions with confirmations
### Steps to reproduce

1. Receive some signet coins e.g. from a faucet, and when they appear in your mempool have a look at transaction activity:

Image

Even though the transaction has 0 confirmations, it is not distinguished from a transaction that has 1.

I don't think it is sufficiently clear why these funds are not reflected in their balance, and that considering these funds as received is extremely dangerous.

Contrast to the qt widgets which shows this:

Image

The pending transaction is on top, and the pending balance is reflected separately.

Also in the transactions view of qt widgets:

There is a question mark for unconfirmed and then a circle that fills up once a transaction has 6 confirmations, not perfect but it does try to communicate to users the danger of treating funds as received after 1 or 2 confirmations

Image

In contrast to this branch where all of these tx'es look the same even though one of them is unconfirmed!!

Image

epicleafies:
This PR fixes a set of correctness problems in the Activity list and payment requests.
Fixes #847. Fixes #848. Fixes #849.

 - Gave the list one ordering rule, shared by the refresh, live inserts and fulfillment, so live and reloaded order agree
 - Made data() a pure read: statuses refresh per block, labels on address book changes, dates on a timer
 - Retried a status read lost to lock contention or taken before the wallet processed the announced block
 - Marked zero-conf transactions with the dashed pending icon and a Pending cue in place of the date
 - Greyed amounts that do not count toward the balance yet, including a request that has received nothing
 - Locked a saved request's amount, and made the address pages edit an existing request instead of creating a second one
 - Tracked pending requests per request rather than per address, so one payment fulfills one request
 - Neutralized a leading =, +, -, @, tab or CR in the CSV export's text columns, leaving amount cells alone. Note: Currently there is a crash (#858), will push fix for it before merge. 
 - Fixed rows being inserted and moved from the node's notification thread, which is unsafe for a Qt model and showed up as a pending row landing below an older confirmed one
 - Localized activity dates through QLocale, matching the Widgets GUI

Added unit, QML and functional tests.

Verified headed on Wayland including first-run onboarding.


qml/models/activitylistmodel.cpp
```diff
@@ -168,6 +203,77 @@ void ActivityListModel::setDisplayUnit(int unit)
...
+    // count until the next block. The Widgets table avoids both by comparing
+    // each row against the wallet's last processed block on every paint.
+    const bool wallet_behind{chain_height >= 0 && wallet_height >= 0 && wallet_height < chain_height};
+    if (!all_read || wallet_behind) {
+        if (!m_status_retry_scheduled && m_status_retry_attempts < MAX_STATUS_RETRIES) {
+            m_status_retry_scheduled = true;
+            const int delay{std::min(STATUS_RETRY_INTERVAL_MS << m_status_retry_attempts,
+                                     MAX_STATUS_RETRY_INTERVAL_MS)};
+            ++m_status_retry_attempts;
+            QTimer::singleShot(delay, this, [this, chain_height] {
```
johnny9: `chain_height` is captured here. If tip H+1 arrives while the retry for H is already scheduled, H+1 cannot schedule another retry. The old callback can finish at H and leave every row one confirmation behind until another block arrives. Please store the latest target height in a member and make the pending timer read that member instead of capturing an old height. A test should cover H+1 arriving while the H retry is pending.

qml/models/activitylistmodel.cpp
```diff
@@ -345,21 +462,77 @@ void ActivityListModel::fulfillPendingRequest(int index, const QSharedPointer<Tr
...
+    endMoveRows();
+}
+
+void ActivityListModel::applyTransactionChanged(const uint256& hash)
+{
+    interfaces::WalletTxStatus wtx;
+    int num_blocks;
+    int64_t block_time;
+    if (m_wallet_model->tryGetTxStatus(hash, wtx, num_blocks, block_time)) {
+        updateTransaction(hash, wtx, num_blocks, block_time);
```
johnny9, line 570: `updateTransaction()` updates only the first row returned by `findTransactionIndex()`. One wallet transaction can produce several activity rows, such as a multi-recipient or self-payment transaction. Now that `data()` is a pure read, the other rows no longer refresh themselves after a non-block `CT_UPDATED`, such as abandoning a transaction. Please find all rows with this hash and update each one. A single O(n) scan is fine here. A regression test should verify that both rows of one transaction become Abandoned.

qml/models/activitylistmodel.cpp
```diff
@@ -345,21 +462,77 @@ void ActivityListModel::fulfillPendingRequest(int index, const QSharedPointer<Tr
...
+        return;
+    }
+    // The wallet often fires the change notification while it still holds
+    // cs_wallet (block processing), so the status read can lose the lock
+    // race. Dropping the update would leave a stale or missing row now
+    // that data() no longer re-reads the wallet, so retry shortly; this is
+    // the event-driven counterpart of the Widgets lazy status re-read on
+    // paint. Notifications only carry wallet transactions, so the read
+    // eventually succeeds.
+    QTimer::singleShot(250, this, [this, hash] { applyTransactionChanged(hash); });
```
johnny9: `ChangeType` is discarded before this retry path. For `CT_DELETED`, `tryGetTxStatus()` can never succeed because the wallet record is gone, so this schedules another read every 250 ms for the lifetime of the model. Please pass `ChangeType` into this function, remove all matching rows from highest to lowest for `CT_DELETED` without reading status, and keep only bounded or deduplicated retries for transient `CT_NEW` and `CT_UPDATED` reads. The stale deleted row was pre-existing; the endless polling is introduced here.

qml/models/activityfilterproxymodel.cpp
```diff
@@ -23,6 +23,19 @@ void WriteCsvValue(QTextStream& stream, QString value)
     stream << '"' << value << '"';
 }
 
+// Sender- and address-book-controlled text cells must not execute as
+// formulas when the exported file is opened in a spreadsheet: neutralize
+// a leading =, +, -, @, tab or CR with a single-quote prefix. Only text
+// columns go through this, so numeric cells keep their minus sign.
+QString GuardedCsvText(QString value)
+{
+    static const QString dangerous_leads{QStringLiteral("=+-@\t\r")};
```
johnny9: Line feed (`\n`) is also a dangerous CSV formula prefix. A label starting with it is not neutralized. Please add `\n` to `dangerous_leads` and cover it with a test.

qml/models/activitylistmodel.cpp
```diff
@@ -168,6 +203,77 @@ void ActivityListModel::setDisplayUnit(int unit)
...
+
+    // Two ways a refresh leaves a row stale. A read lost to wallet lock
+    // contention keeps its cached status above, and a read taken before the
+    // wallet processed the block the node just announced reports a depth one
+    // short: tryGetTxStatus answers with the wallet's own height, which trails
+    // the node's tip until the wallet catches up. Neither re-reads itself,
+    // since data() is a pure read, so a row would keep the wrong confirmation
+    // count until the next block. The Widgets table avoids both by comparing
+    // each row against the wallet's last processed block on every paint.
+    const bool wallet_behind{chain_height >= 0 && wallet_height >= 0 && wallet_height < chain_height};
```
johnny9: This only detects a wallet below the node tip. After a tip disconnect, the wallet can be above the new tip, so stale confirmations are accepted. Please treat any known height mismatch as out of sync and retry.

qml/models/activitylistmodel.cpp
```diff
@@ -195,17 +299,28 @@ void ActivityListModel::addPendingReceiveRequests()
     ReceiveRequestHistoryModel* history = m_wallet_model->receiveRequests();
     if (!history) return;
 
-    QSet<QString> existing_addresses;
+    // Payments the address already received consume its oldest requests
+    // first, the order fulfillPendingRequest uses live, so a reload
+    // reproduces the same set of still-pending rows.
+    QHash<QString, int> received;
```
johnny9: This matches payments and requests only by count. After restart, an unused old payment can consume a request created later. Please replay payments and requests by timestamp and only match requests that existed when the payment arrived.

qml/models/activitylistmodel.cpp
```diff
@@ -75,12 +107,11 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
...
         return tx->label;
     case StatusRole:
-        return tx->status;
+        return static_cast<int>(tx->status);
     case TypeRole:
-        return tx->type;
+        return static_cast<int>(tx->type);
     case TxidRole:
         return tx->isPendingRequest ? QString{} : tx->txid;
     case CanBumpRole:
```
johnny9: `data()` runs while QML creates and repaints delegates. `canBumpTransaction()` can wait for the wallet lock and block the UI. Qt Widgets checks this only when the menu opens. Please query it when transaction details open instead of exposing it as a row role.

johnny9 (commented):
Four additional findings based on the Qt Widgets transaction model and view behavior.

hebasto:
https://github.com/bitcoin-core/gui-qml/actions/runs/34508274197/job/102975724330?pr=875:
```
[373](https://github.com/bitcoin-core/gui-qml/actions/runs/34508274197/job/102975724330?pr=875#step:6:2374)
FAIL!  : bitcoinqml_qmltests::SettingsNavigation::test_walletStackResetsWhenSelectedWalletChangesWhileHidden() Compared values are not the same
   Actual   (): false
   Expected (): true
   Loc: [qrc:/test/qml/tst_settingsnavigation.qml(378)]
```

johnny9 (approved):
ACK a3804929a43cf27d757e107dbb185f5901be0495

This meaningfully improves the Activity page and addresses the main problems in #847, #848, and #849.

Follow-ups:

- Preserve the selected incoming/outgoing record when opening self-payment details.
- Keep open transaction details updated.
- Check every row’s wallet height during status refresh.
- Handle conflicted transactions explicitly instead of displaying a literal `%n`.

jon052885 (approved):

PR #877, opened 2026-08-26
Redesign RPC console and debug log pages

pseudoramdom:
Redesign the RPC console and debug log using the shared Settings page layout and theme

### Key Changes
  - Rebuild both pages with `SettingsPage` and standard page headings.
  - Add a reusable search bar with clear and result-navigation controls.
  - Redesign the debug log as a structured table showing type, time, and message.
  - Add debug-log filtering for all messages or warnings and errors.
  - Add load-more, scroll-to-bottom, and open-debug-log actions.
  - In RPC Console, replace search filter with highlighting search matches and provide previous/next navigation. 

## Screenshots

Screenshot 2026-08-26 at 1 26 20 PM

Screenshot 2026-08-26 at 1 27 12 PM

Screenshot 2026-08-26 at 2 00 16 PM


qml/components/MonospaceOutputView.qml
```diff
@@ -110,6 +115,103 @@ Item {
...
+
+    function rebuildSearchMatches(resetCurrent) {
+        const matches = []
+        if (root.normalizedSearchText.length > 0) {
+            for (let row = 0; row < rowRepeater.count; ++row) {
+                const item = rowRepeater.itemAt(row)
+                if (!item || !item.contentEditor) continue
+                const editor = item.contentEditor
+                const plainText = editor.getText(0, editor.length)
+                const normalized = plainText.toLowerCase()
```
johnny9: When an RPC reply contains `"İstanbul savings"`, searching for `savings` highlights `avings"` instead. The search lowercases the reply before finding matches. Lowercasing `İ` produces two Unicode characters, which shifts the match position by one. That position is then used to select text in the original reply, where `İ` still occupies one character. Use an escaped, case-insensitive regex on the original `plainText` so the match positions refer to the displayed text. Add a regression test using this example.

test/functional/qml_test_debug_log.py
```diff
@@ -103,373 +105,191 @@ def stop(self):
...
-        f"got {restored_count}"
-    )
-    print(
-        f"  PASSED: appended entry loaded and count remained capped at "
-        f"{restored_count}"
-    )
-    return restored_count
+    gui.set_text("debugLogSearchField", "rpc error marker")
+    gui.wait_for_property("debugLogListView", "count", 1, timeout_ms=5000)
+    gui.click("debugLogOptionsButton")
```
johnny9: The debug-log functional test still opens the options menu, but the filters have moved to a separate menu. On Qt 6.8+, the test’s synthetic click activates the hidden filter item and leaves the options menu open. The next options-button click closes it, so the test times out waiting for it to open. Update both filtering sequences to click `debugLogFilterButton` and wait for `debugLogFilterMenu`, then rerun the functional test with Qt 6.8+.

johnny9 (commented):
Two fixes for search highlighting and debug-log test navigation.

qml/components/MonospaceOutputView.qml
pseudoramdom: Addressed in 19f26683d0dd7ec689cf3c7185119427b7707adf

test/functional/qml_test_debug_log.py
pseudoramdom: Addressed in 19f26683d0dd7ec689cf3c7185119427b7707adf

johnny9 (approved):
ACK 4a17a7ef5bd420eabec6219fc118beb9354fbaf0

Both issues are fixed. No new issues found.

PR #878, opened 2026-08-27
fix: use root instead of the nullable parent

### LINKED ISSUE #868: ImportWalletOptions reads parent.width while parent can be null
## Summary

`ImportWalletOptions.qml` defines a top-level property that dereferences `parent.width` without checking whether the page has been parented yet:

```qml
readonly property real heroWidth: Math.min(parent.width - 40, 520)
```

A component can evaluate this binding during construction while `parent` is still null, producing:

```
qrc:/qml/pages/wallet/ImportWalletOptions.qml:22: TypeError: Cannot read property 'width' of null
```

## Validation

I reproduced the warning on both Qt 6.4.2 and Qt 6.11.1 by creating the page component without a parent and then parenting it, which is a supported component lifecycle.

The current normal `PageStack.push()` path happened to assign the parent early enough on Qt 6.11.1 and did not warn in my focused test. This makes the bug lifecycle/path-sensitive, but the binding itself is unambiguously unsafe and the reported Qt 6.4 failure is reproducible.

## Impact

The initial `heroWidth` evaluation fails. The hero block and import-error area can begin with an invalid width until the binding is reevaluated after parenting.

## Expected behavior

The page should be safe throughout component construction and should size itself without dereferencing a nullable parent.

## Suggested fix

Bind to the page itself:

```qml
readonly property real heroWidth: Math.min(root.width - 40, 520)
```

Alternatively, guard `parent`, although the page's own width expresses the intended dependency more directly. A regression test should instantiate the component without an initial parent and assert that no QML error is emitted.

uqlidi:
The parent may be null in the `ImportWalletOptions`, so it's better to use the root instead. Also adding a test for this case

fix #868

### DIFF
--- a/qml/pages/wallet/ImportWalletOptions.qml
+++ b/qml/pages/wallet/ImportWalletOptions.qml
@@ -19,7 +19,7 @@ Page {
     signal next
     background: null
     readonly property bool hasImportError: walletController.walletLoadError.length > 0
-    readonly property real heroWidth: Math.min(parent.width - 40, 520)
+    readonly property real heroWidth: Math.min(root.width - 40, 520)
     readonly property int heroTopMargin: 64
     readonly property int importHeroTopMargin: heroTopMargin + 10
     readonly property int heroIconSize: 60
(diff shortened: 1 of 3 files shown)


uqlidi:
moved to #882

PR #879, opened 2026-08-27
fix: detected binding loops on `implicitWidth`

### LINKED ISSUE #866: Node settings pages report implicitWidth binding loops on Qt 6.4
## Summary

Opening the Wallet, Storage, or Mempool information page from node settings produces an `implicitWidth` binding-loop warning on Qt 6.4.2.

## Reproduction

Open each affected page under Node settings with a Qt 6.4.2 build:

```
qrc:/qml/pages/node/NodeSettings.qml:271:21: QML SettingsWallet: Binding loop detected for property "implicitWidth"
qrc:/qml/pages/node/NodeSettings.qml:274:21: QML SettingsStorage: Binding loop detected for property "implicitWidth"
qrc:/qml/pages/node/NodeSettings.qml:282:21: QML MempoolInformationSettings: Binding loop detected for property "implicitWidth"
```

I reproduced all three warnings with a focused QML test on Qt 6.4.2. The same test did not warn on Qt 6.11.1, so this is Qt-version-sensitive rather than fixed in the application.

## Cause

The affected pages are children of a `StackLayout`, which sizes them using implicit dimensions. Their scroll content then derives its explicit width from its parent:

- `SettingsWallet.qml`: `ColumnLayout { width: Math.min(parent.width, 450) }`
- `InformationPage.qml`, used by Storage and Mempool: `ColumnLayout { width: Math.min(parent.width, 600) }`

That creates a cycle between the layout/page implicit width, the scroll view, and the content width.

## Expected behavior

Navigating among node settings pages should not produce binding loops, and the page width should not depend on Qt breaking the cycle at an implementation-dependent point.

## Suggested fix

Break the feedback path by sizing content from the scroll viewport (for example `scrollView.availableWidth`) or by using `Layout.fillWidth` plus `Layout.maximumWidth`. `SettingsWallet.qml` and the shared `InformationPage.qml` both need the corresponding adjustment.

A regression test should navigate to all three pages under the minimum supported Qt version.

uqlidi:
fix #866

### DIFF
--- a/qml/controls/InformationPage.qml
+++ b/qml/controls/InformationPage.qml
@@ -45,6 +45,7 @@ Page {
 
     background: null
     clip: true
+    contentWidth: root.maximumWidth
 
     header: NavigationBar {
       id: navbar
@@ -60,7 +61,7 @@ Page {
 
         ColumnLayout {
             id: information
-            width: Math.min(parent.width, 600)
+            width: Math.min(parent.width, root.maximumWidth)
             anchors.horizontalCenter: parent.horizontalCenter
             spacing: 0
             Loader {
--- a/qml/pages/settings/SettingsStorage.qml
+++ b/qml/pages/settings/SettingsStorage.qml
@@ -10,6 +10,7 @@ import "../../components"
 
 InformationPage {
     id: root
+    objectName: "settingsStoragePage"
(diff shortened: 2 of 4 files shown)


uqlidi:
moved to #882

PR #880, opened 2026-08-27
fix: undefined `banListModel` on click unban

### LINKED ISSUE #867: Synchronous ban-list reset invalidates the Unban click handler
## Summary

Clicking **Unban** can synchronously reset the ban-list model while the delegate's `onClicked` handler is still running. The reset destroys that delegate; when execution resumes at the next statement, its QML context is gone and `banListModel` is no longer resolvable.

## Reproduction

I added a focused mock in which `unbanAt()` performs a synchronous `beginResetModel()` / `endResetModel()`. On both Qt 6.4.2 and Qt 6.11.1, clicking Unban produced:

```
qrc:/qml/pages/node/BannedPeers.qml:101: ReferenceError: banListModel is not defined
```

The handler never reached its explicit `refresh()` call.

## Cause

[`BannedPeers.qml`](https://github.com/bitcoin-core/gui-qml/blob/3ce18e1e950b8952dca94b46bd713eb138f88526/qml/pages/node/BannedPeers.qml) currently does:

```qml
if (banListModel.unbanAt(index)) {
    banListModel.refresh()
} else {
    unbanActionError.message = qsTr("Could not unban peer. The ban list may have changed.")
    unbanActionError.open()
    banListModel.refresh()
}
```

`BanListModel::unbanAt()` calls the node synchronously. The resulting core notification is re-emitted through `QMetaObject::invokeMethod` with `Qt::AutoConnection`; when it arrives on the GUI thread, `bannedListChanged` is emitted directly and `BanListModel::refresh()` resets the model before `unbanAt()` returns.

The reset removes the delegate that owns the active handler.

## Impact

The handler aborts partway through. A successful unban may still take effect, hiding the problem; a failed unban can abort before the error popup opens.

## Suggested fix

- Remove the handler's redundant `refresh()` calls, because `bannedListChanged` already refreshes the model.
- Queue the core-to-QML `bannedListChanged` hop explicitly with `Qt::QueuedConnection` so a model reset cannot re-enter the QML handler that caused it.

A regression test can use a model that resets synchronously inside `unbanAt()` and assert that the handler completes and the error path remains usable.

uqlidi:
fix #867

### DIFF
--- a/qml/models/nodemodel.cpp
+++ b/qml/models/nodemodel.cpp
@@ -808,7 +808,7 @@ void NodeModel::ConnectToBannedListChangedSignal()
     m_handler_notify_banned_list_changed = m_node.handleBannedListChanged([this]() {
         QMetaObject::invokeMethod(this, [this] {
             Q_EMIT bannedListChanged();
-        });
+        }, Qt::QueuedConnection);
     });
 }
 
--- a/qml/pages/node/BannedPeers.qml
+++ b/qml/pages/node/BannedPeers.qml
@@ -15,6 +15,13 @@ Page {
     signal back()
     background: null
 
+    function unbanPeer(row) {
+        if (!banListModel.unbanAt(row)) {
+            unbanActionError.message = qsTr("Could not unban peer. The ban list may have changed.")
+            unbanActionError.open()
+        }
+    }
+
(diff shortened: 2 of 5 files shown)


uqlidi:
moved to #882

PR #881, opened 2026-08-27
fix: make `CoreCheckBox` import `org.bitcoincore.qt`

uqlidi:
fix #865

PR #882, opened 2026-08-28
fix: QML runtime errors

### LINKED ISSUE #865: CoreCheckBox uses AppMode without importing org.bitcoincore.qt
## Summary

`CoreCheckBox.qml` binds `hoverEnabled` to `AppMode.isDesktop` but does not import the module that provides `AppMode`. Every instance therefore logs a `ReferenceError`, and checkbox hover behavior does not follow the application mode.

## Reproduction

Instantiate a `CoreCheckBox` and observe:

```
qrc:/qml/controls/CoreCheckBox.qml:19: ReferenceError: AppMode is not defined
```

I reproduced this in focused QML tests with both Qt 6.4.2 and Qt 6.11.1.

## Cause

[`qml/controls/CoreCheckBox.qml`](https://github.com/bitcoin-core/gui-qml/blob/3ce18e1e950b8952dca94b46bd713eb138f88526/qml/controls/CoreCheckBox.qml) imports only `QtQuick` and `QtQuick.Controls`, while `AppMode` is registered in `org.bitcoincore.qt 1.0`.

A tree-wide check found this is the only QML file that references `AppMode` without importing that module.

## Expected behavior

`CoreCheckBox` should resolve `AppMode` and set `hoverEnabled` according to `AppMode.isDesktop`, without producing QML errors.

## Suggested fix

Add:

```qml
import org.bitcoincore.qt 1.0
```

A small component test would protect this import dependency.

### LINKED ISSUE #866: Node settings pages report implicitWidth binding loops on Qt 6.4

### LINKED ISSUE #867: Synchronous ban-list reset invalidates the Unban click handler

### LINKED ISSUE #868: ImportWalletOptions reads parent.width while parent can be null

uqlidi:
This bundles four small QML runtime fixes that were opened separately, since each one is a couple of lines plus a test and they are all fixing errors/warnings printed at runtime rather than behaviour changes. Each fix keeps its own commit and its own regression test.

Taken from:

- #878: `ImportWalletOptions` computed `heroWidth` from `parent.width`, but the page's parent can be null (it's not always in a `StackView`), so the binding evaluated on `null` and printed a TypeError. Use `root.width` instead. Fix #868
- #879: `InformationPage` and `SettingsWallet` sized their content `ColumnLayout` from `parent.width` inside a `Flickable` whose `contentWidth` derived from that same content, producing `implicitWidth` binding loops on Qt 6.4. Both pages now set an explicit `contentWidth` from a fixed maximum-width property and the layout reads that property. Fix #866
- #880: Clicking *Unban* invalidated the delegate mid-handler: `handleBannedListChanged` emitted `bannedListChanged` synchronously, which reset the model and destroyed the delegate before the handler finished, leaving `banListModel` undefined. The emit is now queued (`Qt::QueuedConnection`) and the click handler moved to a function on the page. Fix #867
- #881: `CoreCheckBox` uses `AppMode` without importing `org.bitcoincore.qt`, so the type was unresolved at runtime. Added the missing import. Fix #865

### DIFF
--- a/qml/controls/CoreCheckBox.qml
+++ b/qml/controls/CoreCheckBox.qml
@@ -4,6 +4,7 @@
 
 import QtQuick 2.15
 import QtQuick.Controls 2.15
+import org.bitcoincore.qt 1.0
 
 AbstractButton {
     id: root
--- a/qml/controls/InformationPage.qml
+++ b/qml/controls/InformationPage.qml
@@ -45,6 +45,7 @@ Page {
 
     background: null
     clip: true
+    contentWidth: root.maximumWidth
 
     header: NavigationBar {
       id: navbar
@@ -60,7 +61,7 @@ Page {
 
         ColumnLayout {
             id: information
-            width: Math.min(parent.width, 600)
+            width: Math.min(parent.width, root.maximumWidth)
             anchors.horizontalCenter: parent.horizontalCenter
             spacing: 0
             Loader {
--- a/qml/models/nodemodel.cpp
+++ b/qml/models/nodemodel.cpp
@@ -808,7 +808,7 @@ void NodeModel::ConnectToBannedListChangedSignal()
     m_handler_notify_banned_list_changed = m_node.handleBannedListChanged([this]() {
         QMetaObject::invokeMethod(this, [this] {
             Q_EMIT bannedListChanged();
-        });
+        }, Qt::QueuedConnection);
     });
 }
 
--- a/qml/pages/node/BannedPeers.qml
+++ b/qml/pages/node/BannedPeers.qml
@@ -15,6 +15,13 @@ Page {
     signal back()
     background: null
 
+    function unbanPeer(row) {
+        if (!banListModel.unbanAt(row)) {
+            unbanActionError.message = qsTr("Could not unban peer. The ban list may have changed.")
+            unbanActionError.open()
+        }
+    }
+
     header: NavigationBar2 {
         leftItem: NavButton {
             objectName: "bannedPeersBackButton"
@@ -96,15 +103,7 @@ Page {
                     bold: false
                     horizontalPadding: 24
                     text: qsTr("Unban")
-                    onClicked: {
-                        if (banListModel.unbanAt(index)) {
-                            banListModel.refresh()
-                        } else {
-                            unbanActionError.message = qsTr("Could not unban peer. The ban list may have changed.")
-                            unbanActionError.open()
-                            banListModel.refresh()
-                        }
-                    }
+                    onClicked: root.unbanPeer(index)
                 }
             }
         }
--- a/qml/pages/settings/SettingsStorage.qml
+++ b/qml/pages/settings/SettingsStorage.qml
@@ -10,6 +10,7 @@ import "../../components"
 
 InformationPage {
     id: root
+    objectName: "settingsStoragePage"
     property var settingsModel: optionsModel
     property bool customStorage: false
     property int customStorageAmount
--- a/qml/pages/settings/SettingsWallet.qml
+++ b/qml/pages/settings/SettingsWallet.qml
@@ -15,8 +15,10 @@ Page {
 
     signal back
     property bool showBackButton: true
+    readonly property int maximumContentWidth: 450
 
     background: null
+    contentWidth: root.maximumContentWidth
 
     header: SettingsHeader {
         title: qsTr("External signer")
@@ -31,7 +33,7 @@ Page {
         clip: true
 
         ColumnLayout {
-            width: Math.min(parent.width, 450)
+            width: Math.min(parent.width, root.maximumContentWidth)
             anchors.horizontalCenter: parent.horizontalCenter
             spacing: 0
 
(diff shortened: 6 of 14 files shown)


johnny9 (approved):
ACK bf1d64a32a

PR #883, opened 2026-09-01
qml: Stabilize onboarding settings and data directories

### LINKED ISSUE #773: Pre-init onboarding can reread config after mutating startup args

### LINKED ISSUE #774: -resetguisettings can reset the wrong profile before datadir resolution

### LINKED ISSUE #775: Unreadable custom datadir can cause and exception

### LINKED ISSUE #783: The app fails to start on Ubuntu 22.04

### LINKED ISSUE #846: bitcoin-core-app: Assert hit when using a relative datadir path

johnny9:
This combines #812 by @jarolrod with #864 and the review fixes for #812. Jarol's original commit keeps its authorship.

Pre-init onboarding was reading config and settings through the live `gArgs`, then normal startup read them again. A redirected data directory could make the second read use a different profile. GUI reset also ran before Core knew the active data directory and network.

This change builds the onboarding preview with a scratch `ArgsManager`. It keeps the selected data directory separate from the profile resolved by Core, then applies settings after `InitConfig()` and chain-specific `QSettings` setup. Canceling onboarding does not change settings.

## Fixes

- #773: Avoid changing the live startup arguments while building the preview, so config is not read again from a redirected data directory.
- #774: Delay GUI reset until the active profile is resolved. Reset only that profile and leave all profiles unchanged when onboarding is canceled.
- #775: Treat an unreadable saved data directory as unusable and show the chooser. Return config read failures as startup errors instead of aborting.
- #783: Use a Qt 6.2-compatible folder dialog, while keeping the Qt 6.5 and newer implementation.
- #846: Validate explicit data directories before onboarding. Resolve valid relative paths from the current directory and report invalid paths through normal startup error handling.

Reset now uses the final Core settings precedence, treats `-resetguisettings=0` as false, preserves `strDataDir`, and backs up both Core and GUI settings. If a later write fails, it restores the earlier state. For malformed `settings.json` files, the backup keeps the original bytes. Settings stores are opened only when needed, and an apply is rejected if its target profile changed while onboarding was open.

User-facing startup errors and translator context are included.

The Settings-page reset and `bitcoin.conf` actions tracked by #600 remain separate work.

## Tests

- C++ unit tests: 579 passed.
- QML tests: 419 passed.
- Reset GUI settings functional harness with the minimal platform plugin: passed.
- `bitcoin-core-app` build: passed.

Fixes #773
Fixes #774
Fixes #775
Fixes #783
Fixes #846


hebasto:
Please rebase.

PR #886, opened 2026-09-07
fix: ActivityListModel threading race and deprecated injected event params

### LINKED ISSUE #869: CommandConsole relies on deprecated injected event parameters
## Summary

`CommandConsole.qml` uses Qt's deprecated implicit signal-parameter injection for key events. The original runtime warning points to the Tab handler, and current `qmllint` also identifies the Return and Enter handlers.

## Reproduction

Use Tab while the command autocomplete popup is visible:

```
qrc:/qml/pages/node/CommandConsole.qml:365:17 Parameter "event" is not declared.
Injection of parameters into signal handlers is deprecated.
Use JavaScript functions with formal parameters instead.
```

Running Qt 6.11.1 `qmllint` on the file reports the same implicit `event` dependency in:

- `Keys.onReturnPressed`
- `Keys.onEnterPressed`
- `Keys.onTabPressed`

The Tab handler is the user-observed path; the lint result shows the defect is broader than that one warning.

## Cause

The handlers use block syntax while referring to an undeclared `event`, for example:

```qml
Keys.onTabPressed: {
    // ...
    event.accepted = true
}
```

Qt currently injects the signal parameter for compatibility, but this is deprecated.

## Impact

The handlers warn today. When implicit injection is removed, `event` will be undefined: autocomplete acceptance and command submission can throw or allow the key to fall through to another action.

## Suggested fix

Declare the key event explicitly for all three handlers, for example:

```qml
Keys.onTabPressed: (event) => {
    // ...
    event.accepted = true
}
```

A focused key-input test should cover Return, keypad Enter, and Tab autocomplete acceptance.

### LINKED ISSUE #870: ActivityListModel mutates from wallet notification threads and can crash its proxy
## Summary

`ActivityListModel` handles wallet transaction notifications on the notifying thread and directly mutates a model consumed by QML and `QSortFilterProxyModel` on the GUI thread. This violates the model/view threading contract and explains two rare Activity-page crashes observed on Qt 6.4.2.

## Crash evidence

One ASan crash observed `ActivityListModel::data()` reading a null transaction and dereferencing it in `updateTransactionStatus()`:

```
#0 ActivityListModel::updateTransactionStatus(QSharedPointer<Transaction>) const
   qml/models/activitylistmodel.cpp:38
#1 ActivityListModel::data(QModelIndex const&, int) const
   qml/models/activitylistmodel.cpp:66
#2 QSortFilterProxyModel::data(QModelIndex const&, int) const
```

A second report initially had only Qt offsets while posted events were being drained. I resolved them against the exact Debian Qt 6.4.2 debuginfo for build ID `6976b895aa51ea4619aaa4d7f19f2b055ad05f7d`:

```
libQt6Core.so.6+0x2b9b97  QModelIndex::parent()
libQt6Core.so.6+0x2f25ad  QSortFilterProxyModelPrivate::_q_sourceDataChanged(...)
libQt6Core.so.6+0x30a352  postEventSourceDispatch
```

That places the second crash in the Activity proxy processing a source-model update, corroborating the same race rather than an unrelated QObject-lifetime bug.

## Cause

[`ActivityListModel::subscribeToCoreSignals()`](https://github.com/bitcoin-core/gui-qml/blob/3ce18e1e950b8952dca94b46bd713eb138f88526/qml/models/activitylistmodel.cpp) calls `updateTransaction()` directly from the wallet transaction-changed callback. `WalletQmlModel::handleTransactionChanged()` forwards the wallet callback without a queued GUI-thread hop.

Wallet notifications may arrive from validation/network threads, so `ActivityListModel` can modify `m_transactions`, call `beginInsertRows()` / `endInsertRows()`, and emit `dataChanged()` while the GUI thread and proxy are reading the same model.

The established insertion paths construct non-null `QSharedPointer<Transaction>` values. That makes the observed null element more plausibly a symptom of concurrent container/model access than a normal value inserted by application logic.

Other wallet-signal handlers already use `QMetaObject::invokeMethod(..., Qt::QueuedConnection)`, providing an in-tree pattern for the fix.

## Impact

Rare hard crashes while scrolling, opening, or updating the Activity list. A null check in `data()` would avoid one dereference but would not make the model or proxy thread-safe.

## Suggested fix

Marshal every `ActivityListModel` mutation caused by wallet notifications onto the model/GUI thread before touching `m_transactions` or emitting model signals. Add thread-affinity assertions around mutation paths and retain defensive null checks where appropriate.

A stress regression should deliver transaction-change notifications from a worker thread while the proxy/delegates repeatedly read and filter the model.

uqlidi:
- `ActivityListModel` was updated straight from the wallet's transaction-changed callback, which arrives on validation and network threads, while the GUI thread and `ActivityFilterProxyModel` read the same model. The notification is now queued onto the model thread, like the other wallet-signal handlers in `WalletQmlModel` already do. Fix #870

- The `Return`, `Enter` and `Tab` handlers in `CommandConsole.qml` used an undeclared `event`, relying on Qt's deprecated signal-parameter injection. They now declare the parameter explicitly. Fix #869

### DIFF
--- a/qml/models/activitylistmodel.cpp
+++ b/qml/models/activitylistmodel.cpp
@@ -8,9 +8,19 @@
 #include <qml/models/walletqmlmodel.h>
 
 #include <QDateTime>
+#include <QMetaObject>
+#include <QThread>
 #include <QVariantList>
 
 #include <algorithm>
+#include <cassert>
+
+namespace {
+void assertModelThread(const QObject& model)
+{
+    assert(QThread::currentThread() == model.thread());
+}
+} // namespace
 
 ActivityListModel::ActivityListModel(WalletQmlModel *parent)
     : QAbstractListModel(parent)
@@ -35,7 +45,7 @@ int ActivityListModel::rowCount(const QModelIndex &parent) const
 
 void ActivityListModel::updateTransactionStatus(QSharedPointer<Transaction> tx) const
 {
-    if (m_wallet_model == nullptr || tx->isPendingRequest) {
+    if (m_wallet_model == nullptr || tx.isNull() || tx->isPendingRequest) {
         return;
     }
     interfaces::WalletTxStatus wtx;
@@ -50,7 +60,7 @@ void ActivityListModel::updateTransactionStatus(QSharedPointer<Transaction> tx) 
 
 void ActivityListModel::updateTransactionLabel(QSharedPointer<Transaction> tx) const
 {
-    if (m_wallet_model == nullptr) {
+    if (m_wallet_model == nullptr || tx.isNull()) {
         return;
     }
 
@@ -63,6 +73,9 @@ QVariant ActivityListModel::data(const QModelIndex &index, int role) const
         return QVariant();
 
     QSharedPointer<Transaction> tx = m_transactions.at(index.row());
+    if (tx.isNull())
+        return QVariant();
+
     updateTransactionStatus(tx);
 
     switch (role) {
@@ -128,6 +141,7 @@ QHash<int, QByteArray> ActivityListModel::roleNames() const
 
 void ActivityListModel::reload()
 {
+    assertModelThread(*this);
     beginResetModel();
     m_transactions.clear();
     refreshWallet();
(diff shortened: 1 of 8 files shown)


uqlidi:
Force push to rebase and remove merge commit to make it cleaner

PR #884, opened 2026-09-01
ci: Update GHA actions

hebasto:
Also fixes "Node.js 20 is deprecated" CI warnings.

PR #893, opened 2026-09-12
ci: Update GHA `actions/upload-artifact`

hebasto:
This was overlooked in https://github.com/bitcoin-core/gui-qml/pull/884.

Fixes the remaining "Node.js 20 is deprecated" CI warnings.


uqlidi (approved):
ACK a781588

PR #894, opened 2026-09-12
stage, ci: Exclude Qt autogenerated files from compilation database

hebasto:
Exclude Qt autogenerated files under the `qml` subdirectory from the compilation database used in "tidy" CI job.


johnny9 (approved):
ACK 78bf895069502208872a4dd3b320ba4dc503bf65

Follows the existing Qt Widgets autogen exclusion. Autosquash into `qml: Add application shell` on the next rebase.

PR #895, opened 2026-09-14
qml: tooltips on unbivous icon buttons

### LINKED ISSUE #851: design: file icon in activity view unclear (export csv)
In the transaction activity view there is a file icon:

Image

hovering over it gives no tooltip:

Image

And then when I click it the file picker opens, with the only hint about what's going on being the filetype:

Image

Only once I type in a file name does it become clear what I'm doing:

Image

### LINKED ISSUE #862: Icons on Console page have no detail
Unclear what the individuals icon do until you click them. Ideally when you hover over them there is some detail, similar to how there is detail when you hover over the settings toggle and other icons on the top right of the application. 

Image

uqlidi:
- Fix #862: Add hover tooltips to the Console input icons (search/command toggle, increase text size, decrease text size, clear). Each tooltip reuses the button's existing accessible name and opens above the input row.
- Fix #851: Add a tooltip to the Activity export icon ("Export activity to CSV"). The save dialog now has a title and suggests


pseudoramdom:
> * Fix [Icons on Console page have no detail #862](https://github.com/bitcoin-core/gui-qml/issues/862): Add hover tooltips to the Console input icons (search/command toggle, increase text size, decrease text size, clear). Each tooltip reuses the button's existing accessible name and opens above the input row.

FYI - I have a PR to redesign the RPC console page - https://github.com/bitcoin-core/gui-qml/pull/877

uqlidi:
> > * Fix [Icons on Console page have no detail #862](https://github.com/bitcoin-core/gui-qml/issues/862): Add hover tooltips to the Console input icons (search/command toggle, increase text size, decrease text size, clear). Each tooltip reuses the button's existing accessible name and opens above the input row.
> 
> FYI - I have a PR to redesign the RPC console page - #877

Thank you for noticing that. I rebased the branch and kept the tooltips

AndreaDiazCorreia:
Reviewed 5fff2236f775ab19e0b653db939884353c599a51. Built on Arch with Qt 6.11.2, `ctest` passes 2/2, and I tried the tooltips by hand on regtest.

Found one bug on the Activity page: when the filters are open, the header tooltips get painted behind the search row and you only see the arrow. Click the search icon, then hover export or search.

`filterRow` comes after `activityHeader` inside `pageHeader` and neither sets `z`, so the filter row wins. The Loader's `z: 100` only orders inside the `IconButton`, and `pageHeader`'s `z: 1` handles header-over-list, not this. `z: 1` on `activityHeader` fixes it:

https://github.com/bitcoin-core/gui-qml/blob/5fff2236f775ab19e0b653db939884353c599a51/qml/pages/wallet/Activity.qml#L334-L341

One question: I read in `doc/translator-comments.md` that translator comments are required for all user-facing strings, and the four new `qsTr` calls (L308, L365, L390-391) don't have a `//:`. Am I misreading it, or is that for a follow-up?

hebasto:
> One question: I read in `doc/translator-comments.md` that translator comments are required for all user-facing strings, and the four new `qsTr` calls (L308, L365, L390-391) don't have a `//:`. Am I misreading it, or is that for a follow-up?

I'd prefer we include the translator comments in this PR rather than leaving them for a follow-up.

AndreaDiazCorreia (commented):
Re-tested d5978f0c1f778b5bdc50c4a1f2f9151a7be228b4 on Arch with Qt 6.11.2. Both points from my previous review are addressed and `ctest` passes 2/2.

The new regression test is load-bearing: dropping the `z: 1` from `activityHeader` fails test_activity_header_stacks_above_the_filter_row at tst_activity.qml:652, while test_activity_header_stacks_above_the_transaction_list still passes.

Non-blocking: tooltip.shown = true at tst_activity.qml:660 and :673 overrides the hover binding, so these tests cover geometry and stacking without exercising hover activation on the actual button. That seems appropriate for these regression tests; a separate test using mouse movement could complement the existing hover-binding test in tst_hovertooltip.qml.

PR #896, opened 2026-09-15
stage, fixup: Import plugins only for static builds

hebasto:
According to the `qt_import_qml_plugins` [docs](https://doc.qt.io/qt-6/qt-import-qml-plugins.html):
> Note: This command only has any effect if Qt was built statically. If called using a non-static Qt, it will do nothing and return immediately.

However, on some systems, such as Ubuntu 26.04, when using system's Qt packages, the configure step logs noisy warnings as follows:
```
CMake Warning at /usr/lib/x86_64-linux-gnu/cmake/Qt6Qml/Qt6QmlMacros.cmake:4669 (message):
  The qml plugin 'qtquick2plugin' is a dependency of 'bitcoin-qt', but the
  link target it defines (Qt6::qtquick2plugin) does not exist in the current
  scope.  The plugin will not be linked.
Call Stack (most recent call first):
  src/qml/CMakeLists.txt:68 (qt6_import_qml_plugins)
```

This PR fixes the issue.


cmake/module/ImportQtPlugins.cmake
```diff
@@ -35,5 +35,6 @@ function(import_plugins target)
         networkinformation
         tls
     )
+    qt6_import_qml_plugins(${target})
```
johnny9: Static builds with `BUILD_GUI=OFF` and `BUILD_GUI_LEGACY=ON` now fail because the legacy GUI uses this helper without loading Qt’s QML component. Guard this call with `if(BUILD_GUI)` and test that configuration.

hebasto: Thanks! Should be fixed now.

johnny9 (approved):
ACK 65dba128f9b51129479756c89ed6f55e44119b22

The previous static legacy-GUI failure is fixed. No new findings.

Issue #603, opened 2026-05-06
Payment URI handling: add desktop startup, single-instance, and file-open routing


johnny9:
## Code Gap
The QML wallet has Bitcoin URI parsing support in the Send flow, but it does not yet have the Qt widgets `PaymentServer` style desktop integration.

Missing pieces include startup command-line URI capture, routing from a second application instance into the running app, macOS file/open event handling, and app-level delivery of valid `bitcoin:` URIs to the active Send flow.

## Work Needed
- Add desktop/app integration for `bitcoin:` URI open requests.
- Capture payment URIs passed at startup.
- Route URI requests from an already-running instance where supported.
- Handle platform file/open events where applicable.
- Validate and report malformed URI requests using QML app-shell UI.
- Deliver valid requests into the Send flow without losing wallet context.

PR #897, opened 2026-09-16
qml: Do not let Qt parse the process command line

### LINKED ISSUE #885: Add test for URI Argument Injection Vulnerability
See https://achow101.com/2021/02/0.18-uri-vuln.

AndreaDiazCorreia:
Closes #885.

Desktop URI handlers pass the URI to the application as a command line argument, so a crafted `bitcoin:` URI can append further arguments. Qt parses the command line when `QApplication` is constructed, before Bitcoin Core does, and consumes
options such as `-platformpluginpath`. See https://achow101.com/2021/02/0.18-uri-vuln.

The widgets GUI has constructed `QApplication` with a synthetic argument list since bitcoin/bitcoin#16578, but that was never carried over here, so the test this issue asks for cannot pass as things stand. This PR adds the mitigation and the test. Injected options now reach `ParseParameters()` and are rejected as unknown instead of being applied by Qt.

Startup errors are also logged to stderr. `InitErrorMessageBox()` only showed them in a window, and the test asserts on the parser error.

Argument sanitation in the style of `PaymentServer::ipcParseCommandLine()` is left for #603, which adds command line URI capture.


johnny9 (approved):
ACK 9f104156cc937b22f4bf953b253c822eaf6b9300

The argument-isolation change matches the Qt Widgets implementation.

Built and tested on Ubuntu 24.04 with Qt 6.4.2. Both CTest suites and the argument-injection, startup, onboarding, URI-import, and shutdown tests passed. Both intermediate commits also built and passed CTest and lifecycle checks.

Confirmed both injection probes fail on the base and pass on this head. 

Bitcoin submodule lint and CI clang-tidy 22.1.8 passed without source findings.

I think the functional python test is appropriate here as well and gives enough regression coverage to be confident. A c++ test would require a new target.

hebasto:
CI fails: https://github.com/bitcoin-core/gui-qml/actions/runs/35124376780/job/105211721961?pr=897

AndreaDiazCorreia:
@hebasto  I fixed the issue, and everything should be working now

hebasto (approved):
ACK ac0b32f1383244e8fdcfe084e8b9816cd2080041.

PR #900, opened 2026-09-17
test: Fix animation timing assumptions in QML tests

### LINKED ISSUE #898: `bitcoinqml_qmltests` fails
https://github.com/bitcoin-core/gui-qml/actions/runs/35222004840/job/105204193203:
```
FAIL!  : bitcoinqml_qmltests::SettingsNavigation::test_rpcConsoleUsesSettingsContainerAndTracksVisibility() Compared values are not the same
   Actual   (): true
   Expected (): false
   Loc: [qrc:/test/qml/tst_settingsnavigation.qml(606)]
```

johnny9:
Fixes #898.

Wait for focus borders, filter colors, and popups to finish changing before checking them. Keep the Copy button feedback visible until its appearance checks finish, then trigger and check the reset.

Only QML tests change. All 520 checks pass with normal and 4× animation durations. Lint and clang-tidy pass. Sanitizer runs report the same Qt hash and leak issues with and without the fix.


pseudoramdom:
ACK 538470db6119e19b38d20ddc88c0056332e03c6a

---

If it helps, we can add a repo wide motion-policy. A central setting that could conceptually look like:

```qml
// Theme or a Motion singleton
property bool animationsEnabled: true

// animation
Behavior on opacity {
    enabled: Theme.animationsEnabled
}
```

or
```qml
property int shortDuration: animationsEnabled ? 150 : 0

Behavior on opacity {
    duration: shortDuration
}
```

johnny9:
> If it helps, we can add a repo wide motion-policy. A central setting that could conceptually look like

Yes I think we should either disable animations or fake time

PR #901, opened 2026-09-18
qml: Fix use-after-free when clearing extra send recipients

### LINKED ISSUE #856: Heap use-after-free when clearing send recipients
SendRecipientsListModel::clear() (qml/models/sendrecipientslistmodel.cpp) deletes each SendRecipient in a loop while its pointer is still in m_recipients; the list is only cleared after the delete loop. During the surrounding beginResetModel()/endResetModel() signal emissions, QML re-enters WalletQmlModel (for example through a coin-control or fee binding) and calls scheduleFeeEstimates(), which runs BuildRecipients() over m_recipients and dereferences the freed recipients.

ASAN on an unmodified qt6 build (99618b0c18) confirms the use-after-free: freed in SendRecipient::~SendRecipient via clear(), used in BuildRecipients via scheduleFeeEstimates invoked from QML.

clearToFront() in the same file has the same delete-before-remove shape, and additionally removes rows without beginRemoveRows/endRemoveRows (it only emits countChanged), a model-contract violation that desyncs any attached view.

Impact: reproduces intermittently (roughly 3 to 7 percent of runs) when importing a multi-recipient PSBT, which is the recurring qml_test_psbt.py CI failure at the "multi-recipient PSBT submitted" step. It is also reachable from transaction discard and the normal Send page teardown, so it can crash the app during ordinary send flows. #825 (crash on send, with a cascade of null-property TypeErrors from Send.qml during post-send teardown) and #841 (intermittent crash sending a labeled multi-recipient transaction, not reproducible on demand) are consistent with this defect: multi-recipient, teardown-timed, and intermittent.

Suggested fix: detach the container before destroying elements (swap-then-clear, or deleteLater) in both clear() and clearToFront(), and add the missing row-removal notifications to clearToFront() (mirroring the existing remove()).

Related: #762 proposes separating imported-PSBT state from the send draft, which would remove the entanglement at the root; the memory-safety fix stands on its own and is needed regardless.

AndreaDiazCorreia:
Completes #856. The `clear()` half was fixed in 66ded734; `clearToFront()` kept the same shape.

It destroyed each extra recipient while its pointer was still in `m_recipients`, so anything re-entered from that destruction, such as the fee estimation a QML binding schedules, walked the list and dereferenced freed memory. The rows were also removed without `beginRemoveRows()`/`endRemoveRows()`, which leaves an attached view out of sync. The path runs whenever multiple recipients are turned off in the send options popup.

The rows are now removed through the model notifications and detached before the recipients are destroyed, with the deletion deferred as `clear()` already does.

Two regression tests, both failing before the fix: one asserts the removal is reported once, the other that no recipient is still listed when its `destroyed()` signal fires and that the deletion is deferred rather than skipped.

`ctest` passes on Qt 6.11 and on Qt 6.2 with GCC 12, plus `qml_test_send_review.py` and `qml_test_psbt.py` locally, the latter three times since #856 links it to an intermittent failure there.

Fixes #856


johnny9 (approved):
ACK 3b4761c1c800483e0715b2ed8132be47921d1172

Correctly addresses the recipient lifetime issue. Both new tests pass and fail against the old implementation as expected.

PR #915, opened 2026-09-18
qml: Redesign activity around grouped transactions and transaction flows

### LINKED ISSUE #845: design: Payment requests should probably not be shown as "pending receives"
To reproduce, create a payment request:

Image

Pending receive implies that these are funds that are en-route in some way and might eventually arrive.

### LINKED ISSUE #851: design: file icon in activity view unclear (export csv)

### LINKED ISSUE #853: Consider unifying activity rows that belong to the same underlying Bitcoin transaction
Today, sends and receives are presented as independent activity rows. This is straightforward for transactions containing a single recipient.
But this becomes misleading when one transaction contains multiple recipients.

Consider the scenario here - 
Bob makes a single transaction paying Alice 0.2BTC and Charlie 0.1 BTC

The activity list displays 2 separate send entries, even though they both belong to the same transaction and share the same TXID, confirmations, timestamp, fee, inputs, replacement history (RBF).

Image

Navigating to the detail view of one of the entires, the page is titled "Transaction" but the primary content describes only the selected destination output, which seems to be conflating a transaction & an output.

Image

### LINKED ISSUE #904: Send flow: the note field is never persisted, and it is actually the recipient label
The Send page collects a "Note to self" into the recipient's label field (qml/pages/wallet/Send.qml, sendNoteField writes recipient.label), but the send path never stores it: sendTransactionInternal commits the transaction without the setAddressBook call the Widgets GUI performs after a successful send (src/qt/walletmodel.cpp, SendCoins stores the label as a send-purpose address book entry). The text the user typed is silently dropped, and the Activity row for the sent transaction shows the bare address even though the user provided exactly the text that should title it.

The naming compounds it: the field says "Note to self" (private-sounding) but is functionally the recipient/address label, and is even prefilled from a scanned BIP21 URI's label, which is authored by the payee.

Suggested fix: store the label as a send-purpose address book entry after a successful send (Widgets parity; the existing address book refresh then titles the Activity row), and retitle the field for what it is, for example "Recipient" or "Their name", with its visibility/source marked the way the receive-side fields now mark theirs.

### LINKED ISSUE #907: The "Sent to yourself" Activity filter can never match anything
Nothing in gui-qml ever sets Transaction::SendToSelf. Transaction::fromWalletTx decomposes a self-payment into a Sent part and a Received part sharing one txid, matching current Core (src/qt/transactionrecord.cpp has the same decomposition and no SendToSelf type at all; the type was dropped upstream while gui-qml kept the enum member and built a filter option on it). Selecting "Sent to yourself" in the Activity type filter therefore always yields an empty list, and its CSV export label is unreachable.

Fixing it is a decision rather than a mechanical change: either drop the filter option (matches Core's current rendering, visible UI change) or reintroduce self-payment detection (a deliberate divergence from Core). A user who paid their own address will reasonably select this filter and conclude the transaction is missing.

### LINKED ISSUE #908: Design pass: how Activity rows and payment requests present status, titles, and time
A set of interdependent presentation questions collected while testing the Activity and request flows. They constrain each other, so proposing them together for a designer call:

- Status vocabulary. The color/symbol language (purple/blue/green, triangle/circle/filled circle) has grown ad hoc and could use one deliberate revisit.
- Confirming has no cue. A transaction at 1 to 5 confirmations looks identical to unconfirmed, so a mined transaction reads as unsettled for six blocks. Consider an intermediate cue or an n/6 hint.
- Titles are inconsistent. List rows and embedded request entries are titled by name, the transaction details header is the static "Transaction", the request detail header is "Payment request". No consistent rule.
- Dates mix absolute and relative. The transaction details request subtitle shows an absolute creation date; the request detail page and all list rows use relative time.
- The Payment request filter's date column mixes semantics. Pending rows show the status text "Pending receive" while used rows show a bare relative creation date, so a correctly sorted list reads shuffled, and "1 hr ago" on a used row reads like a payment time when it means created. Options: prefix request dates ("Requested 1 hr ago"), group pending above used, or a separate status line keeping the date column temporal.
- Titles are inconsistent. List rows and embedded request entries are titled by name, the transaction details header is the static "Transaction", the request detail header is "Payment request". No consistent rule.
- Dates mix absolute and relative. The transaction details request subtitle shows an absolute creation date; the request detail page and all list rows use relative time.
- The Payment request filter's date column mixes semantics. Pending rows show the status text "Pending receive" while used rows show a bare relative creation date, so a correctly sorted list reads shuffled, and "1 hr ago" on a used row reads like a payment time when it means created. Options: prefix request dates ("Requested 1 hr ago"), group pending above used, or a separate status line keeping the date column temporal.

### LINKED ISSUE #913: Small fixes around the Activity and Addresses pages
A batch of small independent fixes:

- Activity list has no scroll bar. activityListView attaches no ScrollBar.vertical while the request detail, wallet select, and coin selection views do; long histories are the normal case for this page.
- "Show used addresses" toggle has no objectName (qml/pages/wallet/AddressList.qml page menu), so functional tests cannot reach used addresses through the real UI path.
- "New request" no longer matches the QR-code button beside it. The OutlineButton default border was dimmed to neutral2 in the addresses redesign, so the generated-request view now pairs a thin grey non-bold button with a bright bold hand-rolled sibling; align both on the shared control with explicit props.
- The Spanish catalog is stale for the Activity page. Its Activity context holds only the string "Activity"; the search field, filter chips, and relative date strings are missing, so they render in English regardless of language. Regenerate the catalog from source.
- Amount filter comment overclaims (qml/pages/wallet/Activity.qml, above minAmountPlaceholder/minAmountPattern): the BTC pattern admits 8 integer digits, so input can exceed MAX_MONEY. Consequence is cosmetic (an impossible threshold filters everything, the C++ parse clamps), but the comment should state what holds, or the BTC digits be tightened.

pseudoramdom:
Transactions with multiple recipients or mixed wallet actions can appear as separate activity entries, making it difficult to understand which actions belong together and their overall effect on the wallet (see https://github.com/bitcoin-core/gui-qml/issues/853). This PR introduces transaction-level activity and rebuilds the activity list and transaction details around it.

- A dedicated **Pending** section for unconfirmed transactions and awaiting-payment requests. The pending balance reflects wallet impact, including outgoing fees, without counting unpaid requests.
- Grouped transaction actions, keeping multiple sends, receipts, and internal transfers under one transaction while preserving individual amounts, notes, and payment-request associations.
-  Clearer transaction types, distinguishing sends, receives, multiple actions, self-sends, consolidations, splits, and mining rewards. The “Sent to yourself” filter includes all internal-transfer categories.
- A transaction flow diagram showing inputs, outputs, ownership, change, and fees. Large transactions collapse non-wallet outputs while keeping wallet-owned outputs visible, and unavailable input amounts remain explicitly unknown.
- Rebuilt transaction details with confirmation status, wallet impact, fee information, transaction metadata, raw transaction copying, payment-request links, and fee-bump and replacement navigation.
- More flexible filtering and browsing, including multiple activity types, date presets and custom ranges, amount ranges, search, day/month grouping, and saved compact or comfortable layouts.

Fixes #845
Fixes #851  
Fixes #853 
Fixes #907 
Fixes #904 
Fixes #908 
Fixes #913

### Other fixes
- The changes also address responsiveness. Activity loading, input resolution, address-label lookups, and fee-bump eligibility checks run off the GUI thread. 
- Details can display cached wallet outputs immediately while remaining information loads. Collapsed flows avoid creating hidden output delegates, and batched model updates and cached totals reduce work during confirmation refreshes.

## Screenshots

<figure>
    Screenshot 2026-09-18 at 9 32 44 AM
    <figcaption>Activity list</figcaption>
</figure>

--- 

<figure>
    Screenshot 2026-09-18 at 9 34 21 AM
    <figcaption>Activity list in Compact display density</figcaption>
</figure>

---

<figure>
    Screenshot 2026-09-18 at 9 32 54 AM
    <figcaption>Sends</figcaption>
</figure>

---

<figure>
    Screenshot 2026-09-18 at 9 33 06 AM
    <figcaption>Sends</figcaption>
</figure>

---

<figure>
   Screenshot 2026-09-18 at 9 33 12 AM
    <figcaption>Receive</figcaption>
</figure>

---

<figure>
   Screenshot 2026-09-18 at 9 33 16 AM
    <figcaption>Replaced Tx</figcaption>
</figure>


.github/workflows/ci.yml
```diff
@@ -17,9 +17,11 @@ jobs:
           - job_name: 'Ubuntu 22.04, Qt 6.2'
             image: ubuntu-22.04
             # Jammy's default GCC 11 is below Bitcoin Core's GCC 12.1 minimum.
-            extra_packages: 'g++-12'
+            extra_packages: 'g++-12 xvfb xauth'
             svg_package: 'libqt6svg6-dev'
-            cmake_options: '-DCMAKE_CXX_COMPILER=g++-12'
+            # Qt 6.2's offscreen plugin cannot capture software-rendered images.
+            cmake_options: '-DCMAKE_CXX_COMPILER=g++-12 -DQML_TEST_RENDERING_PLATFORM=xcb'
+            test_wrapper: 'xvfb-run -a'
```
pseudoramdom, line 24: Codex suggested adding xvfb and xauth to provide a virtual display for the Ubuntu 22.04 rendering tests. 
We'll see if checks pass with this

PR #916, opened 2026-09-20
qml: Fix Qt 6.4 use-after-free in node warnings and information popups

### LINKED ISSUE #857: Repeater inside ColumnLayout can crash node info and warnings popups
NodeWarningsPopup.qml and NodeInformationPopup.qml both place a Repeater as a direct child of a ColumnLayout, with a model that can change after the first layout pass (the node warning list, and the caller-set rows property respectively).

This is the same shape that crashed the Send tab through AlertPopup: on Qt 6.4, a Repeater that is a direct child of a Quick Layout whose model changes during the first layout pass hits a use-after-free in QQuickLayout::effectiveSizePolicy_helper (a qobject_cast on a freed delegate) and SIGSEGVs. The crash only occurs under a real windowing backend (xcb or wayland), so the offscreen test suites stay green while the app can crash for users. See the fix and explanatory comment in AlertPopup.qml from #750; these two instances were not converted at the same time.

Suggested fix: wrap the Repeater delegates in a Column positioner inside the layout cell, as AlertPopup now does.

AndreaDiazCorreia:
Fixes #857.

`NodeWarningsPopup` and `NodeInformationPopup` place a `Repeater` directly inside a `ColumnLayout`, the shape that crashed `AlertPopup` and was fixed in 91f2b9b4 (#750); these two were not converted then. On Qt 6.4 the grid layout engine reads a freed delegate when the layout rearranges while the `Repeater` is rebuilding one, and both models change after the first layout pass: the warning list follows node notifications, and `rows` is replaced on every `onAboutToShow`.

The rows move into a `Column` positioner, which never hands its children to the layout engine, with the delegate widths bound to the scroll view width.

### Testing

The crash needs a real windowing backend, so no offscreen layer catches it and no test is added, as in 91f2b9b4. Reproduced under Qt 6.4.2 with `xcb` on Xvfb: a temporary test that reopens and resizes the popups segfaults on f7c3b42073 and completes 100 cycles on this branch, while the 15 existing `NodeFeedback` cases pass on both. On Qt 6.11.2 the QML suite and the unit tests pass, with the 4 pre-existing `PeerActions` focus failures unchanged.

The same shape remains in `PeerDetails.qml:363`, `SegmentedPicker.qml:41`, `ContextMenuPicker.qml:93` and `SettingsDesignSystem.qml:377`; happy to convert them here or in a follow-up.


johnny9 (approved):
ACK 0e5fc14452d8e3f7443821a05e8eecf792015f55

pseudoramdom:
> The same shape remains in PeerDetails.qml:363, SegmentedPicker.qml:41, ContextMenuPicker.qml:93 and SettingsDesignSystem.qml:377; happy to convert them here or in a follow-up.

I can take care of these files since in one of my PRs

pseudoramdom:
> > The same shape remains in PeerDetails.qml:363, SegmentedPicker.qml:41, ContextMenuPicker.qml:93 and SettingsDesignSystem.qml:377; happy to convert them here or in a follow-up.
> 
> I can take care of these files since in one of my PRs

Fixed in https://github.com/bitcoin-core/gui-qml/pull/915/commits/78a870836684a6570cd98bd4e6756c9c43c06708 as part of #915

pseudoramdom:
ACK https://github.com/bitcoin-core/gui-qml/commit/0e5fc14452d8e3f7443821a05e8eecf792015f55

